PR #7, opened 2016-02-25
Const-correctness for brain::Circuit

eile:


adevress:
Hi Stefan,

The reason I did not declare these method as "const" is that const is suppose to be  thread-safe methods and a lot of HDF5 ones are not by default.
They can also write when you "read" for some meta-data operations.

adevress:
I guess that's fine if we map https://www.hdfgroup.org/HDF5/doc/RM/RM_H5.html#Library-IsLibraryThreadsafe to the namespace and document this in the README.

It would be a pattern to follow for the other calls too.

eile:
1) const does not necessarily mean threadsafe
2) HDF5 is not thread-safe, HighFive does not need to be neither. In Brion we have an HDF5 lock to get thread-safety. We could decide to push this down to HighFive

adevress:
about const http://stackoverflow.com/questions/14127379/does-const-mean-thread-safe-in-c11 ;)

That's fine anyway, I will open an issue to propagate this change on the rest of the API

PR #13, opened 2016-06-16
MSVC fixes.

ptoharia:


ptoharia:
Still getting an error in H5Node_traits_misc.hpp:109 because non const array size inside the "do/while" loop. I can try to fix it but I don't exactly understand what that code is doing.
@tribal-tec FYI

adevress:
Hi @ptoharia 

Thank you for your pull request.

I just updated the /ReleaseDebugAutoFlags.cmake with our current new version that introduce CompilerFlagsHelpers.cmake and portability for different compiler.

Could you rebase your modification on this ? It should make easier and more straight forward for you to support MSVC.

Concerning H5Node_traits_misc.hpp:109, does MSVC throw your an error or a warning ?

 It is using a small dynamically-sized stack allocated buffer, it is not officially part of the C++ norm but part of C99. Most of the compiler ( gcc, clang, icc, xlc ) support this by default, I would be surprised the Microsoft one does not.

ptoharia:
@adevress thanks and sorry because I missed your message. Done.

Regarding H5Node_traits_misc.hpp:109 it is an error. I'm using Visual Studio 12 (2013). This is the compiler output:

H5Node_traits_misc.hpp(109): error C2057: expected constant expression 
H5Node_traits_misc.hpp(109): error C2466: cannot allocate an array of constant size 0

On the other hand, with your MSVC compiler flags I get this warning:

execution_monitor.ipp(948): warning C4535: calling _set_se_translator() requires /EHa

adevress:
@ptoharia  Ok, MSVC does not seem to manager variable-length array properly.

You can replace the buffer by a std::vector<char> pre-allocated at the max_read_size. 

It will have a little performance hint due to the memory allocation, but nothing significant compared to the I/O itself anyway.

What this section does is rather simple :

- Get the number elements in a collection ( e.g a HDF5 Group )
- Get the named list of each element ( e.g all the dataset names )
- If any element has a name with a length longer to the default name size:
  ->  do not get it
  -> call it again with a NULL buffer in order to get the real length size
  -> use the real length size as buffer size
  -> call it again with the proper length size

ptoharia:
@adevress Done ... now a new thing came up which don't see how to fix it:

H5Slice_traits_misc.hpp(77): error C2668: 'HighFive::details::vectors_to_single_buffer' : ambiguous call to overloaded function

Any ideas?

Issue #26, opened 2017-03-17
License


mrzv:
Are you open to re-licensing the library under something more permissive than LGPL? (Like a BSD or an MIT license?) My employer insists on releasing software under BSD and won't let us include any GPL or LGPL code.

(It's also not entirely clear to me what it means for a header-only library to be under LGPL. There is no possibility of dynamic linking, so isn't it exactly the same as GPL?)

adevress:
I dont have any strong feeling about License. 

My understanding of the LGPL is that it acts like a BSD for header-only source code, excepted that you have to maintain HighFive under LGPL if you fork it.

Eigen a very popular header-only C++ Library is also under LGPL and seems to understand the license the same way I do with confirmation from the FSF.

http://eigen.tuxfamily.org/index.php?title=Licensing_FAQ&oldid=1117#Forking_Eigen

You can release your software under BSD and still use HighFive under LGPL, there is not issue from my side.

Is it a major problem for you ?

mrzv:
Eigen has switched to MPL2, but that one unfortunately is also a no go. I understand your interpretation of LGPL (although I think it's debatable that that's what it says). For me, the real issue are our lawyers, who won't allow any software included in any of our own releases that expresses any opinion about patents. As a results, GPL, LGPL, Apache, Mozilla, etc. are non-starters.

Eigen has the advantage that it's preinstalled on most systems, so it can be treated as an external dependency. HighFive is not as widely adopted.

adevress:
I see. I have personnality no problem for re-licensing to something more permissive if the others contributors do not oppose to it  @AliCanDemiralp  @eile  @ptoharia.

For header-only library, I generally favorise the Boost Software License (http://www.boost.org/users/license.html) which has the advantage to be even more permissive than the BSD, and without all the problems of the BSD licenses ( compatibility with GPL, re-licensing, do not need redistribution of License file ).

if it's fine for you and the other contributors approve, I can re-license np

mrzv:
Boost license would be perfect. Thanks!

PR #27, opened 2017-03-20
Re-license project from LGPL2.1+ to Boost

adevress:
Due to the header only nature of the project.  Re-license project from LGPL2.1+ to Boost


adevress:
Ping issue #26

PR #52, opened 2017-07-05
Added target exporting logic for find_package to work from an install…

hernando:
… tree.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,6 @@ enable_testing()
 set(HIGHFIVE_VERSION_MAJOR 1)
 set(HIGHFIVE_VERSION_MINOR 2)
 
-
 add_subdirectory(deps/git-cmake-format)
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/CMake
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -28,27 +26,64 @@ if(HIGHFIVE_PARALLEL_HDF5)
...
-    find_package(MPI)
+  find_package(MPI)
 endif()
 
 install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include/highfive
 		DESTINATION ${INCLUDE_INSTALL_DIR})
 
-include_directories(${CMAKE_CURRENT_SOURCE_DIR}/include ${HDF5_INCLUDE_DIRS} ${Boost_INCLUDE_DIR} ${MPI_C_INCLUDE_PATH})
+# cmake target install
+if(CMAKE_MAJOR_VERSION GREATER 2) # No export(EXPORT) in CMake 2
```
adevress: Hi Juan,

Sorry to bother you on that but two little points please :
- Can you make that disable by default and only enable with a -D option ?
- And if possible in a separated cmake file.

I know you guys love cmake targets for your git subprojects but I do not share this love a lot.

I would prefer the targets files to not invade every installation per default please

hernando: Agreed for the separate file.

In any case, I think in the end it's not needed because I was doing something wrong initially (I should clean the CMakeCache.txt).
This is needed for facilitating FindPackage(HighFive) if you install it and not having to worry about much more (just adding HighFive to your target_link_libraries). 
In other words, this saves you from having to create a FindHighFive.cmake script, something you haven't faced because you are including HighFive as a submodule.

If you have such a strong opinion against it you can simply reject it, for me having to use an option to enable it it's almost the same thing as not having it at all.

And I don't think the cmake files it produces are polluting any more than pkgconfig does.

adevress: My main issue with the cmakeConfig files is that they are intrusive and directly export and modify variables in cmake configuration.
But if you think it's needed for your project, jsut split it as a separated file and I will accept it.

hernando: I've done the split.

I don't think the variables exported and from the cmakeConfig files are any more intrusive than a FindFoo.cmake borrowed from somewhere, with two added advantages:
- The Config files are always in sync with the version of the software that produced them.
- Since they use target properties to configure include and link requirements instead of variables, your client code is simpler and your namespace is less polluted than with FindFoo.cmake files.
Regarding modified variables, I don't get what you mean by that, I don't see parent scope variables being modified in a Config.cmake file.

In any case, the config,cmake file generation is not the most important part of this PR. It's the target_include_directories what's really useful for client code because then you get all the include dependencies automatically (including the transitive ones) by simple adding HighFive to your target_link_libraries.

PR #54, opened 2017-07-24
Fix problems with declaration order when including certain headers.

hernando:


adevress:
Hi Juan,

Could you attach a little file triggering the issue, I'm curious to see which header order triggers that

hernando:
#include <highfive/H5DataSet.hpp>

That's it.
From what I could understand, DataSet is used in H5Annotation_traits_misc.hpp. The soluation works by not including this file from H5Annoation_traits.hpp but doing it at the end of the files including it.
Another possible solution is to guard the inclusion of this header with an ifdef to not do it from the include in H5DataSet.hpp

adevress:
Yep, but I'm surprise because I don't see any reference to DataSet into H5Annotate_traits_misc.hpp

hernando:
The reference is through the template parameter in static_cast<Derivate*>(this)->getId(). The error is a bit surprising to me because I don't remember having this problem with CRTP before.

adevress:
Fixed in a separate pull request out of your comments

PR #56, opened 2017-07-28
Additions for Brion

hernando:


include/highfive/bits/H5Node_traits_misc.hpp
```diff
@@ -104,6 +104,28 @@ inline size_t NodeTraits<Derivate>::getNumberObjects() const {
...
+    ssize_t length =
+        H5Lget_name_by_idx(static_cast<const Derivate*>(this)->getId(), ".",
+                           H5_INDEX_NAME, H5_ITER_INC, index,
+                           buffer, maxLength, H5P_DEFAULT);
+    if (length < 0)
+        HDF5ErrMapper::ToException<GroupException>(
+            "Error accessing object name");
+    if (length <= maxLength)
+        return std::string(buffer);
+    char *bigBuffer = new char[length + 1];
```
adevress: Could you use a std::vector<char> here instead of a dynamic allo for exception safety ? 

Also force the truncation of the std::string to "length", I would not trust HDF5 to do the right thing for that.

include/highfive/bits/H5Node_traits_misc.hpp
```diff
@@ -104,6 +104,28 @@ inline size_t NodeTraits<Derivate>::getNumberObjects() const {
...
+    char buffer[maxLength + 1];
+    ssize_t length =
+        H5Lget_name_by_idx(static_cast<const Derivate*>(this)->getId(), ".",
+                           H5_INDEX_NAME, H5_ITER_INC, index,
+                           buffer, maxLength, H5P_DEFAULT);
+    if (length < 0)
+        HDF5ErrMapper::ToException<GroupException>(
+            "Error accessing object name");
+    if (length <= maxLength)
+        return std::string(buffer);
```
adevress: same here for the truncation at length

include/highfive/bits/H5Node_traits_misc.hpp
hernando: What do you mean by truncating the string to length? As I've read in the reference, HDF5 writes the null terminating character.

hernando: In fact, the best option would be to use std::unique_ptr<char[]> to not initialize the memory.

adevress: I mean replacing std::string(buffer) by something like std::string(buffer, length) which will enforce Null terminaison exactly at "length" if HDF5 does not do it properly.

adevress: > In fact, the best option would be to use std::unique_ptr<char[]> to not initialize the memory.

There is no C++11 for now unfortunatly :(

hernando:
Ready to merge?

PR #60, opened 2017-08-17
Do not throw from ~Object

hernando:
Triggers a warning in new compilers and it's not recommended because
it can cause double exceptions during stack unwinding.
In C++11 destructors are implicitly noexcept, so std::terminate will be
called unless noexcept(false) (C++11 only syntax) is added.


hernando:
@adevress, please review

PR #61, opened 2017-08-24
Added a utility class to silence HDF5 errors inside a scope.

hernando:


hernando:
@adevress  please review (You should also change the configuration of the project so I can assign reviewers or at least you get the PR notifications automatically).

PR #63, opened 2017-08-25
Prepare release 1.4

tristan0x:
I would like to bump HighFive version in BlueBrain/bbp-nixpkgs to take benefits of documentation generation. @adevress What do you think of a new release for the occasion?


adevress:
+1 for the release

adevress:
I just found a crash on a unit test. let's fix that before 1.4

PR #70, opened 2017-09-05
Split H5Slice_traits::read in two functions.

hernando:
The new one is for plain C arrays. This also solves the annoyance
of having to provide an l-value for T* buffers.


hernando:
@adevress please review. 
This works for my client code without changing it, please review that it also does for yours.

adevress:
Hi Juan,

Could you add the same logic for write ? 
to keep it consistent

hernando:
Sure, I didn't thought about it

PR #75, opened 2017-10-17
[H5DataType] added h5py compatible support for std::complex<double> data type

dluitz:
Adds numpy/h5py compatible support for complex as AtomicType.


adevress:
Thanks for the contribution, merged

PR #89, opened 2018-01-10
add getOffset method to DataSet

wolfv:
As promised. Was much easier than anticipated. 

I'm not sure wether the tests work as expected.

### DIFF
--- a/include/highfive/bits/H5DataSet_misc.hpp
+++ b/include/highfive/bits/H5DataSet_misc.hpp
@@ -56,9 +56,8 @@ inline DataSpace DataSet::getMemSpace() const { return getSpace(); }
 
 inline size_t DataSet::getOffset() const {
     haddr_t addr = H5Dget_offset(_hid);
-    if (addr == HADDR_UNDEF)
-    {
-        throw std::runtime_error("Cannot get offset of DataSet.");
+    if (addr == HADDR_UNDEF){
+        HDF5ErrMapper::ToException<DataSetException>("Cannot get offset of DataSet.");
     }
     return addr;
 }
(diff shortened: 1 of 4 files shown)


tests/unit/tests_high_five_base.cpp
```diff
@@ -693,6 +693,20 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(readWriteVector2D, T, numerical_test_types) {
...
+    std::string dsetname = "dset";
+    const size_t size_dataset = 20;
+
+    File file(filename, File::ReadWrite | File::Create | File::Truncate);
+    std::vector<int> data(size_dataset);
+    std::iota(data.begin(), data.end(), 0);
+    DataSet ds = file.createDataSet<int>(dsetname, DataSpace::From(data));
+    ds.write(data);
+    DataSet ds_read = file.getDataSet(dsetname);
+    BOOST_CHECK_EQUAL(ds_read.getOffset(), 2144);
```
adevress: The 2144 value here is guarantee by the spec somewhere or it is what work on your machine ? Is this portable ?

include/highfive/bits/H5DataSet_misc.hpp
```diff
@@ -53,6 +53,10 @@ inline DataSpace DataSet::getSpace() const {
 }
 
 inline DataSpace DataSet::getMemSpace() const { return getSpace(); }
+
+inline size_t DataSet::getOffset() const {
+    return H5Dget_offset(_hid);
```
adevress: Could you throw an exception in case H5Dget_offset return HADDR_UNDEF ?

tests/unit/tests_high_five_base.cpp
wolfv: You're right, I didn't consider that this might be different on different machines. 
Thanks for the comments. I'm going with your suggestions of testing `>0`.

PR #95, opened 2018-01-26
H5Object_misc.hpp uses std::cerr but doesn't import iostream

dbichko:


adevress:
Thank you for the fix

PR #117, opened 2018-06-19
Deflate and shuffle properties

joeyo:
This adds support for the deflate (compression) and shuffle (aids compression) filters, using the same interface as for chunking. I include a simple unit-test that writes compressed datasets to files, reads them back and compares. Manual inspection with h5dump shows that the filters are being applied, e.g.:

```
$ h5dump -H -p h5_rw_deflate_x_test.h5 
HDF5 "h5_rw_deflate_x_test.h5" {
GROUP "/" {
   DATASET "dset" {
      DATATYPE  H5T_STD_I64LE
      DATASPACE  SIMPLE { ( 128, 32 ) / ( 128, 32 ) }
      STORAGE_LAYOUT {
         CHUNKED ( 16, 16 )
         SIZE 2623 (12.493:1 COMPRESSION)
      }
      FILTERS {
         PREPROCESSING SHUFFLE
         COMPRESSION DEFLATE { LEVEL 9 }
      }
      FILLVALUE {
         FILL_TIME H5D_FILL_TIME_IFSET
         VALUE  H5D_FILL_VALUE_DEFAULT
      }
      ALLOCATION_TIME {
         H5D_ALLOC_TIME_INCR
      }
   }
}
}
```

### DIFF
--- a/tests/unit/tests_high_five_base.cpp
+++ b/tests/unit/tests_high_five_base.cpp
@@ -1113,37 +1113,44 @@ void readWriteShuffleDeflateTest() {
     const std::string DATASET_NAME("dset");
     const size_t x_size = 128;
     const size_t y_size = 32;
+    const size_t x_chunk = 16;
+    const size_t y_chunk = 16;
 
     const int deflate_level = 9;
 
-    File file(filename.str(), File::ReadWrite | File::Create | File::Truncate);
-
-    // Create the data space for the dataset.
-    std::vector<size_t> dims;
-    dims.push_back(x_size);
-    dims.push_back(y_size);
-
-    DataSpace dataspace(dims);
-
-    // Use chunking
-    DataSetCreateProps props;
-    props.add(Chunking(std::vector<hsize_t>{16, 16}));
-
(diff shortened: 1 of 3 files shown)


tests/unit/tests_high_five_base.cpp
```diff
@@ -1105,3 +1105,63 @@ BOOST_AUTO_TEST_CASE(HighFiveOutofDimension) {
...
+
+    // Create a dataset with arbitrary type
+    DataSet dataset = file.createDataSet<T>(DATASET_NAME, dataspace, props);
+
+    T array[x_size][y_size];
+
+    ContentGenerate<T> generator;
+    generate2D(array, x_size, y_size, generator);
+
+    dataset.write(array);
```
hernando: A flush should be added after writing.

ferdonline: @hernando How about writing within its own scope? (like it's done for reading?)

hernando: That's also OK. I suggested a flush because it's what other tests do.

PR #118, opened 2018-06-22
Move constructor and move assignment for Properties.

hernando:


hernando:
@ferdonline, please review. This fixes the compilation of syntool

PR #119, opened 2018-06-22
Use ifdef instead of if, to fix link errors in other projects.

hernando:


hernando:
@ferdonline, please review

ferdonline:
Thanks

PR #120, opened 2018-06-30
added HDF5_DIR cmake variable support

ct-clmsn:
detects where a non-system hdf5 installs are user provides the following command-line configuration to cmake.

```
cmake -DHDF5_DIR=<path to hdf5-cmake files>
```
script assumes the following placement for hdf5-config.cmake

```
<path to hdf5-cmake files>/share/cmake
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,7 @@ option(USE_BOOST "enable Boost Support" TRUE)
 option(HIGHFIVE_EXAMPLES "Compile examples" TRUE)
 option(HIGHFIVE_PARALLEL_HDF5 "Enable Parallel HDF5 support" FALSE)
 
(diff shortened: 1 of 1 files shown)


ferdonline:
Hi @ct-clmsn. I am wondering if your case can't be covered by simply setting HDF5_ROOT, instead of changing the Cmake in that way. Eventually provide some more info on your case.

ct-clmsn:
@ferdonline when I attempt to use -DHDF5_ROOT on the master branch, CMake declares:

```
CMake Error at .../FindHDF5.cmake:85 (include)" include could not find load file:
HDF5_ROOT_DIR-NOTFOUND/hdf5-config.cmake
```

When HDF5_ROOT_DIR is set in the cmake command as '-DHDF5_ROOT_DIR=<path>' my system creates linker errors (undefined references).

Both of these issues prompted the PR - hope this helps contextualize the PR a little better.

PR #125, opened 2018-08-16
Support chunk cache configuration properties for dataset create/open

hernando:


hernando:
@tristan0x, @adevress Please review

hernando:
Could you merge it? I can't do it

Issue #100, opened 2018-02-21
Adding support for std::array


wizmer:
Hi,

It would be great if HighFive could support std::array as well :)

Thanks !

joebarker87:
I've started coding this in a branch: https://github.com/joebarker87/HighFive/tree/feature/std_array

At the moment I've just covered basic writing of std::arrays and nested versions (i.e. std::array<std::array> and std::vector<std::array> etc.).

I didn't both with `#ifdef H5_USE_CXX11` guards because it looks like version 2.0 requires CXX11 anyway.

It's fairly straightforward to change the template specialisations and recursions. One potential problem will be that std::array cannot be resized. Upon reading the H5 file I think we will just have to throw an exception if the sizes don't match.

Any feedback or pointers from the main developers would be most welcome.

henryiii:
My PR solves the first level of this - unnested `std::array`s. @joebarker87, if you are interested, you could extend this in a new PR to provide nested arrays.

PR #129, opened 2018-09-16
Adding support for std::array

henryiii:
This should add support for std::array, which is missing (though inferior old C style arrays were supported!).

I tried to add a test, as well.

Edit: Addresses #100.

### DIFF
--- a/include/highfive/bits/H5Converter_misc.hpp
+++ b/include/highfive/bits/H5Converter_misc.hpp
@@ -181,16 +181,14 @@ struct data_converter<
     std::array<T, S>,
     typename std::enable_if<(
                          std::is_same<T, typename type_of_array<T>::type>::value)>::type> {
-    inline data_converter(std::array<T,S>& vec, DataSpace& space, size_t dim = 0) {
-        if(space.getDimensions().size() != 1) {
+    inline data_converter(std::array<T, S>& vec, DataSpace& space, size_t dim = 0) {
+        if (space.getDimensions().size() != 1) {
             throw DataSpaceException("Only 1D std::array supported currently.");
         }
-        if(space.getDimensions()[0] != S) {
+        if (space.getDimensions()[0] != S) {
             std::ostringstream ss;
-            ss << "Impossible to pair DataSet with "
-               << space.getDimensions()[0] << " elements into "
-                       "an array with "
-               << S << " elements.";
+            ss << "Impossible to pair DataSet with " << space.getDimensions()[0]
+               << " elements into an array with " << S << " elements.";
             throw DataSpaceException(ss.str());
         }
         (void)vec;
--- a/include/highfive/bits/H5Converter_misc.hpp
+++ b/include/highfive/bits/H5Converter_misc.hpp
@@ -182,14 +182,16 @@ struct data_converter<
     typename std::enable_if<(
                          std::is_same<T, typename type_of_array<T>::type>::value)>::type> {
     inline data_converter(std::array<T,S>& vec, DataSpace& space, size_t dim = 0)
-    : _space(&space), _dim(dim) {
-        assert(_space->getDimensions().size() >= dim);
+    : _space(&space) {
+        assert(_space->getDimensions().size() > 0);
+        assert(_space->getDimensions().at(0) == S);
         (void)vec;
+        (void)dim;
     }
     
     inline typename type_of_array<T>::type*
     transform_read(std::array<T, S>& vec) {
-        return &(vec[0]);
+        return vec.data();
     }
     
     inline typename type_of_array<T>::type*
@@ -200,7 +202,6 @@ struct data_converter<
     inline void process_result(std::array<T, S>& vec) { (void)vec; }
     
     DataSpace* _space;
-    size_t _dim;
 };
 
 #ifdef H5_USE_BOOST
@@ -306,6 +307,7 @@ struct data_converter<std::vector<T>,
     size_t _dim;
     std::vector<typename type_of_array<T>::type> _vec_align;
 };
+
 
 // apply conversion to scalar string
 template <>
--- a/include/highfive/bits/H5Converter_misc.hpp
+++ b/include/highfive/bits/H5Converter_misc.hpp
@@ -15,6 +15,7 @@
 #include <numeric>
 #include <sstream>
 #include <string>
+#include <array>
 
 #ifdef H5_USE_BOOST
 #include <boost/multi_array.hpp>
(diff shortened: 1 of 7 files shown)


henryiii:
Looks like I got unlucky and copied a test that wasn't run (missing BOOST macro). Will try to fix.

henryiii:
Fixed tests and code. Currently only supports 1D (non-nested) `std::array`s. This could be expanded similar to `std::vector`, but for now this is simpler and fits my use case. The test for vectors has been restored, as well.

include/highfive/bits/H5Utils.hpp
```diff
@@ -57,6 +58,12 @@ struct array_dims<T[N]> {
     static const size_t value = 1 + array_dims<T>::value;
 };
 
+// Only supporting 1D arrays at the moment
+template<typename T, std::size_t N>
+struct array_dims<std::array<T,N>> {
+    static const size_t value = 1;
```
hernando: Shoulnd't it be?

    static const size_t value = 1 + array_dims<T>::value;

hernando: All the const can be constexpr

include/highfive/bits/H5Converter_misc.hpp
```diff
@@ -174,6 +175,34 @@ struct data_converter<
...
+    
+    inline typename type_of_array<T>::type*
+    transform_write(std::array<T, S>& vec) {
+        return vec.data();
+    }
+    
+    inline void process_result(std::array<T, S>& vec) { (void)vec; }
+    
+    DataSpace* _space;
+    size_t _dim;
```
hernando: Members not needed, remove.

include/highfive/bits/H5Converter_misc.hpp
```diff
@@ -174,6 +175,34 @@ struct data_converter<
...
+// apply conversion to std::array
+template <typename T, std::size_t S>
+struct data_converter<
+    std::array<T, S>,
+    typename std::enable_if<(
+                         std::is_same<T, typename type_of_array<T>::type>::value)>::type> {
+    inline data_converter(std::array<T,S>& vec, DataSpace& space, size_t dim = 0)
+    : _space(&space), _dim(dim) {
+        assert(_space->getDimensions().size() >= dim);
+        (void)vec;
```
hernando: All other types are resizing the back storage according to `space` except in the case of plain C arrays, where not even the size can be checked. Since here you can't resize, but at least you can check the size, you should verify that space.getDimensions()[0] == S and throw DataSpaceException otherwise. Exercise this in the test.

include/highfive/bits/H5Utils.hpp
henryiii: Change the existing ones too? `static constexpr size_t`?

include/highfive/bits/H5Converter_misc.hpp
```diff
@@ -174,6 +175,34 @@ struct data_converter<
...
+    }
+    
+    inline typename type_of_array<T>::type*
+    transform_read(std::array<T, S>& vec) {
+        return &(vec[0]);
```
hernando: Why not vec.data()?

include/highfive/bits/H5Utils.hpp
hernando: Yes, I'd do it

include/highfive/bits/H5Converter_misc.hpp
henryiii: That's how `std::vector` was done. I changed one of them to `.data`, will do the other too.

include/highfive/bits/H5Utils.hpp
henryiii: Done.

tests/unit/tests_high_five_base.cpp
```diff
@@ -647,10 +647,12 @@ void readWriteVectorTest() {
     std::ostringstream filename;
     filename << "h5_rw_vec_" << typeid(T).name() << "_test.h5";
 
-    std::srand((unsigned int)std::time(0));
     const size_t x_size = 800;
     const std::string DATASET_NAME("dset");
     typename std::vector<T> vec;
+    vec.resize(x_size);
+    for(size_t i=0; i<x_size; i++)
```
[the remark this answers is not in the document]
henryiii: Assuming this is about the spacing, done. Also changed the spacing in the array test to match.

include/highfive/bits/H5Converter_misc.hpp
henryiii: Dropped the `_dim`. I think `_space` is still needed?

include/highfive/bits/H5Converter_misc.hpp
henryiii: Changed, see if this is better.

include/highfive/bits/H5Converter_misc.hpp
henryiii: Done (only changed the arrays, not the vectors).

include/highfive/bits/H5Converter_misc.hpp
hernando: You don't need to use 'at' if you have already checked that the size is not 0. But what I was suggesting is to have an if and throw an exception. In H5Slice_traits_misc.hpp you can see that the dimensions are checked and an exception is if they don't match. I don't see a good reason to not doing the same thing here as the penalty of checking that compared to the IO operation is going to be negligible. I've realized that the assert is not needed because of the check made by the caller, but I know the other converters are doing it.

tests/unit/tests_high_five_base.cpp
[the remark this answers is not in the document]
hernando: Yes, it was

include/highfive/bits/H5Converter_misc.hpp
hernando: In the other (when it's used) it appears in process_results, transform_read or transform_write. See the C array converter, it doesn't have anything.

henryiii: Removed.

include/highfive/bits/H5Converter_misc.hpp
henryiii: Is the new version better?

henryiii:
Great, thanks! If anyone is interested, a good next step would be to extend this to nested arrays. (For me, I'm using `multi_array` from Boost instead of nested arrays, so this is sufficient).

tests/unit/tests_high_five_base.cpp
```diff
@@ -667,11 +669,48 @@ void readWriteVectorTest() {
...
+    File file(filename.str(), File::ReadWrite | File::Create | File::Truncate);
+
+    // Create a dataset with double precision floating points
+    DataSet dataset = file.createDataSet<T>(DATASET_NAME, DataSpace::From(vec));
+
+    dataset.write(vec);
+
+    typename std::array<T, x_size> result;
+
+    dataset.read(result);
```
hernando: Here I forgot to ask for

    typename std::array<T, 1> tooSmall;
    BOOST_CHECK_THROW(dataset.read(tooSmall), DataSpaceException)

Do you mind creating another PR for it?

PR #134, opened 2018-09-19
Adding variadic DataSpace and Chunking constructor

henryiii:
This expands the current shortcut:

DataSpace(3) -> 1D dataspace: [3]

to work with any number of dimensions:

DataSpace(2,3,4) -> 3D dataspace: [2,3,4]

It also fixes the current odd behavior where two integer parameters, DataSpace(1,2) matches against the iterator constructor, which then matches against a different constructor in vector, causing a really odd result. This was done by not allowing an integer-like value through the template for the iterator constructor.

A similar fix was added to Chunking for consistency.

### DIFF
--- a/tests/unit/tests_high_five_base.cpp
+++ b/tests/unit/tests_high_five_base.cpp
@@ -657,6 +657,32 @@ BOOST_AUTO_TEST_CASE(DataSpaceVariadicTest) {
                                   space2b_ans.begin(), space2b_ans.end());
 }
 
+BOOST_AUTO_TEST_CASE(ChunkingConstructorsTest) {
+    Chunking first(1,2,3);
+
+    auto first_res = first.getDimensions();
+    std::vector<hsize_t> first_ans{1,2,3};
+
+    BOOST_CHECK_EQUAL_COLLECTIONS(first_res.begin(), first_res.end(),
+                                  first_ans.begin(), first_ans.end());
+
+    Chunking second{1,2,3};
+
+    auto second_res = second.getDimensions();
+    std::vector<hsize_t> second_ans{1,2,3};
+
+    BOOST_CHECK_EQUAL_COLLECTIONS(second_res.begin(), second_res.end(),
+                                  second_ans.begin(), second_ans.end());
+
+    Chunking third({1,2,3});
+
+    auto third_res = third.getDimensions();
+    std::vector<hsize_t> third_ans{1,2,3};
+
+    BOOST_CHECK_EQUAL_COLLECTIONS(third_res.begin(), third_res.end(),
+                                  third_ans.begin(), third_ans.end());
+}
+
 template <typename T>
 void readWrite2DArrayTest() {
     std::ostringstream filename;
--- a/tests/unit/tests_high_five_base.cpp
+++ b/tests/unit/tests_high_five_base.cpp
@@ -129,9 +129,7 @@ BOOST_AUTO_TEST_CASE(HighFiveBasic) {
     BOOST_CHECK_EQUAL(file.getName(), FILE_NAME);
 
     // Create the data space for the dataset.
-    std::vector<size_t> dims;
-    dims.push_back(4);
-    dims.push_back(6);
+    std::vector<size_t> dims = {4,6};
 
     DataSpace dataspace(dims);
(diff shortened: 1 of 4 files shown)


henryiii:
I also fixed `DataSpace({1,2,3})`, which was supported before on Clang, but not on GCC (ambiguous constructor). It is now explicitly handled correctly.

tests/unit/tests_high_five_base.cpp
```diff
@@ -129,9 +129,7 @@ BOOST_AUTO_TEST_CASE(HighFiveBasic) {
     BOOST_CHECK_EQUAL(file.getName(), FILE_NAME);
 
     // Create the data space for the dataset.
-    std::vector<size_t> dims;
-    dims.push_back(4);
-    dims.push_back(6);
+    std::vector<size_t> dims = {4,6};
```
hernando: You can also write dims{4, 6}

tests/unit/tests_high_five_base.cpp
```diff
@@ -598,6 +592,63 @@ BOOST_AUTO_TEST_CASE(DataSpaceTest) {
...
+    // Create 1D shortcut dataspace
+    DataSpace space(7);
+
+    BOOST_CHECK_EQUAL(space.getDimensions().size(), 1);
+    BOOST_CHECK_EQUAL(space.getDimensions()[0], 7);
+
+    // Initializer list (explicit)
+    DataSpace space3({8, 9, 10});
+
+    BOOST_CHECK_EQUAL(space3.getDimensions().size(), 3);
```
hernando: Use BOOST_CHECK_EQUAL_COLLECTIONS

henryiii: I just saw you use this in the last PR and already forgot about it... In my defense, I was inheriting code that was already there. :)

henryiii:
Currently, these are the supported constructors that take integers:
```cpp
DataSpace(1)      // Old shortcut
DataSpace(1,2)   // Used to match iterator constructor
DataSpace{1,2}   // New
DataSpace({1,2}) // Used to work on Clang, but not GCC (now works on both)
```
I think the (size_t, ...) constructor (line 2) is okay since it is an expansion of the old shortcut (line 1), but for simplicity and similarity to std::vector, it could be dropped in favor of only supporting line 3.

Whatever we do here probably should be done for Chunking as well, but that has never had a shortcut form (line 1), so we could only support lines 3-4 for Chunking. What do you think?

tests/unit/tests_high_five_base.cpp
henryiii: I've fixed the other uses of `push_back`, as well.

hernando:
>  I think the (size_t, ...) constructor (line 2) is okay since it is an expansion of the old shortcut (line 1), but for simplicity and similarity to std::vector, it could be dropped in favor of only supporting line 3.

I was thinking the same thing while reviewing and I discarded the idea thinking it would be an API breaking change, but it's not! because in fact you have added the constructor with the initializer list. Remove it then.

henryiii:
It does break usage of `DataSpace(3)`, unless an explicit constructor is added for it (like in the past). Then the question becomes why does `DataSpace(3,2)` not work (discoverability: of course, the solution is to use `DataSpace{3,2}` or `DataSpace({3,2})`. If you want me to add a non-variadic 1D constructor instead, like we've had in the past, I can. I assume this is less of an issue for Chunking since `Chunking(4)` was never supported.

hernando:
I don't get you mean. Removing only `explicit DataSpace(std::initializer_list<size_t> items);` from your additions wouldn't break DataSpace(1) or DataSpace(2, 3). I even tested to make sure.

henryiii:
Oh, we are talking about two different constructors. The `std::initializer_list<size_t> items` one makes sure that `DataSpace({1,2,3})` works - it's supposed to (and in fact `DataSpace({1,2,3}, {4,5,6})` does work), but due to C++ implicit constructors it fails to match against the single vector constructor because there are also auto-generated DataSpace move/copy constructors that are similar (I don't know why this is ambiguous, but it is), and fails.

I was talking about the variadic one, which I rather would like to keep.

Issue #135, opened 2018-09-21
Critical bug in broadcasting


henryiii:
I've been getting incomprehensible segfaults after converting to HighFive from the H5 C++ bindings, and I've traced it down to the following reproducible bug (I think/hope):

If you have a datatype stored as a column vector or row vector, HighFive correctly "broadcasts" any size one dimensions out. So `[3,1]` or `[1,3]` both should be able to be read into a vector and produce 3 elements. However, this does not work (silently reads 1 value in, since `checkDimensions` is correct but `data_converter` is not) for `[1,3]` (which, in my bad luck, is how the old bindings wrote vectors). Here's the test case that will fail currently:

```cpp
// Broadcasting is supported
BOOST_AUTO_TEST_CASE(ReadInBroadcastDims) {

    const std::string FILE_NAME("h5_missmatch1_dset.h5");
    const std::string DATASET_NAME("dset");

    // Create a new file using the default property lists.
    File file(FILE_NAME, File::ReadWrite | File::Create | File::Truncate);

    // Create the data space for the dataset.
    std::vector<size_t> dims_a{1,3};
    std::vector<size_t> dims_b{3,1};

    // 1D input / output vectors
    std::vector<double> some_data{5.0, 6.0, 7.0};
    std::vector<double> data_a;
    std::vector<double> data_b;

    DataSpace dataspace_a(dims_a);
    DataSpace dataspace_b(dims_b);

    // Create a dataset with double precision floating points
    DataSet dataset_a = file.createDataSet(DATASET_NAME + "_a", dataspace_a, AtomicType<double>());
    DataSet dataset_b = file.createDataSet(DATASET_NAME + "_b", dataspace_b, AtomicType<double>());

    dataset_a.write(some_data);
    dataset_b.write(some_data);

    DataSet out_a = file.getDataSet(DATASET_NAME + "_a");
    DataSet out_b = file.getDataSet(DATASET_NAME + "_b");

    out_a.read(data_a);
    out_b.read(data_b);

    BOOST_CHECK_EQUAL_COLLECTIONS(
            data_a.begin(), data_a.end(),
            some_data.begin(), some_data.end());

    BOOST_CHECK_EQUAL_COLLECTIONS(
            data_b.begin(), data_b.end(),
            some_data.begin(), some_data.end());
}
```

PR #136, opened 2018-09-21
Fix broadcasting on read

henryiii:
This is the test mentioned in #135, and a possible fix. Only fixes 1D vectors and arrays at the moment.

It works by giving the unnested vector and array converters a filtered version of the dimensions array, where all the 1s have been removed unless it is all 1s, in which case a single 1 is kept.

Higher dimensional arrays may still have issues where they don’t throw an error, but miss the correct non-1 dimensions.

### DIFF
--- a/include/highfive/bits/H5Converter_misc.hpp
+++ b/include/highfive/bits/H5Converter_misc.hpp
@@ -34,6 +34,18 @@ namespace HighFive {
 
 namespace details {
 
+inline bool is_1D(const std::vector<size_t>& dims)
+{
+    return std::count_if(dims.begin(), dims.end(),
+                         [](size_t i){ return i > 1; }) < 2;
+}
+
+inline size_t compute_total_size(const std::vector<size_t>& dims)
+{
+    return std::accumulate(dims.begin(), dims.end(), 1,
+                           std::multiplies<size_t>());
+}
+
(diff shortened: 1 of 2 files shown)


henryiii:
This is not necessarily the correct solution, though it is a solution for the cases I’ve tested. I can verify in a larger code base that it fixes the odd C-api segfaults when you don’t read the whole array many times. I’m not sure why this would be a C-api segfault, but it made it very hard to track down.

henryiii:
Is this a reasonable fix, or is something else better? Currently, HighFive cannot read files with scalars and 1D arrays created with a pretty normal setup with the old C++ HDF5 bindings, and the missmatch between the HighFive checker and reader causes segfaults instead of throwing an error.

hernando:
@henryiii Looks reasonable at a first glance, but I would like to understand better the problem myself before deciding whether this is the best solution at the moment or not. Regretfully I didn't have to look into it last week.

hernando:
I've slightly modified your solution. In particular, I have removed the getFilteredDimensions method as filter is more generic than removing dimensions which are one. It was only used in the implementation so I prefer keeping it there. I also changed the code to not create a new dims vector and simply process the given one to get exactly what is needed and nothing more.

henryiii:
Excellent, thank you!

PR #140, opened 2018-11-13
Fix missing include for old versions of BOOST

henryiii:
On some versions of Boost (1.64, at least), you can't import ublas without a specific header first. See [this issue](https://stackoverflow.com/questions/44534516/error-make-array-is-not-a-member-of-boostserialization). This fixes it by importing the extra required header first in HighFive.

Could it be possible to just add definitions instead of importing the boost headers? And then the imports would only be needed in user code if the user actually used the boost features. Not sure if it would work, but it would be much nicer if it did.

PS: Though the issue I linked didn't mention it, for me it seems to not be necessary on Boost 1.68; I only noticed this when running with 1.64.

### DIFF
--- a/include/highfive/H5DataSpace.hpp
+++ b/include/highfive/H5DataSpace.hpp
@@ -16,6 +16,10 @@
 #include <initializer_list>
 
 #ifdef H5_USE_BOOST
+
+// In some versions of Boost (starting with 1.64), you have to include the serialization header before ublas
+#include <boost/serialization/vector.hpp>
+
 #include <boost/multi_array.hpp>
 #include <boost/numeric/ublas/matrix.hpp>
 #endif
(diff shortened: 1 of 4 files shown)


henryiii:
I think older versions of boost don't have this header at all. So there's a specific range it needs to be added to. :/

hernando:
Yes, CI is not passing because we use an older version (1.62 IIRC)

henryiii:
To move forward, we'll need to see what versions of Boost need this, and just add it for those versions. :/.

henryiii:
Fixed it by including `boost/serialization/vector.hpp` instead; this header exists in older versions of Boost and includes the `array_wrapper` header on 1.64+ (where `array_wrapper` was added and the problem was introduced).

Issue #123, opened 2018-07-28
Support non-STL types


tdegeus:
I just came across your nice project. Is there any way that non-STL types that one would use for multi-dimensional arrays (e.g. Eigen, xtensor, ...) could we directly supported, without, at least seemingly, first passing through `std::vector`? I have done this in my own lightweight simple wrapper [HDF5pp](http://hdf5pp.geus.me) for Eigen (and my own n-d array library [cppmat](http://cppmat.geus.me)).

wolfv:
@tdegeus just FYI we have a PR in xtensor-io that adds support for HDF5 using this library

henryiii:
By the way, https://github.com/nlohmann/json has excellent support for adding converters (for json, in this case). I wonder if a similar system could be used for HighFive someday?

PR #141, opened 2018-11-18
H5Easy API

### LINKED ISSUE #139: Recursively create groups
I want to switch from my own wrapper [HDF5pp](https://github.com/tdegeus/HDF5pp) to using HighFive by contributing to a xtensor-io wrapper. The [current PR](https://github.com/QuantStack/xtensor-io/pull/18) appears to be too complicated as it wraps HighFive. I want to avoid this using the following syntax

```cpp
HighFive::File file("/tmp/file.h5", HighFive::File::ReadWrite);

xt::array<double> data = xt::load(file, "/path/to/data");

// ....

xt::dump(file, "/other/path", data);
```

However, I directly miss a feature that both the PR and my own library do have, and that I deem essential in use. And that is automatically recursively creating groups such that a DataSet with an arbitrary path can be created. I was looking to open a PR for HighFive, but I must say that I don't directly see where to place such a function, and how to make `createDataSet` use it.

Are you open to such a feature? How can I integrate it?

@hernando @wolfv

tdegeus:
Proposition to add one-line support, allowing a syntax like

```cpp
#include <Eigen/Eigen>
#include <HighFive/H5OneLine.hpp>

int main()
{
  HighFive::File file("example.h5", HighFive::File::Overwrite);

  std::vector<double> A = {1., 2., 3.};

  HighFive::dump(file, "/path/to/A", A);
  HighFive::dump(file, "/path/to/A", A, HighFive::Mode::Overwrite);

  int B = 10;

  HighFive::dump(file, "/path/to/B", B);

  HighFive::dump(file, "/path/to/C", 10, {0});
  HighFive::dump(file, "/path/to/C", 11, {1});
  HighFive::dump(file, "/path/to/C", 12, {3});

  Eigen::MatrixXd D = Eigen::MatrixXd::Random(10,5);

  HighFive::dump(file, "/path/to/D", D);

  A = HighFive::load<std::vector<double>>(file, "/path/to/A");

  B = HighFive::load<int>(file, "/path/to/B");

  std::vector<int> Cr = HighFive::load<std::vector<int>>(file, "/path/to/C");

  D = HighFive::load<Eigen::MatrixXd>(file, "/path/to/D");

  return 0;
}
```

whereby `dump` and `load` feature many more templated overloads (*scalar, string, vector, xtensor, Eigen*). In addition the free functions `createGroups` (to recursively create groups), `exist` (to check existence without throwing), `shape` and `size` (to get the shape and size of a DataSet) are proposed. 

I am curious to your comments. 

To-do remains:

* Including headers in CMake installation.
* Documentation.
* ... (feel free to add!)

> Closes #139 and #123


henryiii:
On the writing syntax, is it really better than:

```cpp
HighFive::File file("example.h5", HighFive::File::Overwrite);
std::vector<double> A = {1., 2., 3.};
file.createGroup("/path");
file.createGroup("/path/to");
file.createDataSet("/path/to/A", A);
```

? Other than group creation, it's just about as concise. If the rest is accepted, of course, it's consistent with that. I like how it's a stand-alone addition though; in theory it could even be a separate library with a mention in the docs. I don't see any tests (and a few examples would also be nice).

(Just my thoughts, I'm just a fellow contributor to HighFive, and am not supporting or opposing the addition. I like short syntax, so *maybe* mild support so far.)

tristan0x:
I am not sure to see the benefits of this syntax compared to what we have currently. Furthermore, I am not in favor of having several ways to express the same thing, for the sake of clarity and maintenance over time.

tdegeus:
@tristan0x @henryiii I think the discussion should be broken-up in the following three parts. For each of which there exists I believe other arguments.

1.  General usability can be aided by having:

    * A recursive `createGroup`.
    * An `exist` that returns a `bool` on **any** path instead of throwing.

2.  Intuitive support functions:

    * `size`
    * `shape`

3.  One-line, templated, read-write support for many common objects (like scalars, vector, xtensor, Eigen).

ferdonline:
> A recursive createGroup.

I agree the current function should support that

> An exist that returns a bool on any path instead of throwing.

That exists already. Check highfive/bits/H5Node_traits.hpp for `exist`

> Intuitive support functions: size, shape

`H5DataSpace:: getNumberDimensions ` and `H5DataSpace::getDimensions()` dont suit you?

> One-line, templated, read-write support for many common objects (like scalars, vector, xtensor, Eigen).

There were recent additions for short syntax, but eventually there are cases to be improved. I don't believe we should attempt to support conversion to several formats in this core lib. That sounds to me more like a lib which uses HighFive, or a plugin (though we don't support plugins yet)

tdegeus:
@ferdonline 

This for me *throws*:

```cpp
HighFive::File file("test.h5", HighFive::File::Overwrite);
file.exist("/path/to/group");
```

The `size` and `shape` that I proposed differ in the sense that they only require the `path` (and an open file). They wrap around `DataSet` and `DataSpace`.

ferdonline:
@tdegeus I'm afraid that full paths are not recursively checked and thus it throws. I don't really agree with that behavior. Feel free to open an issue with your test and results. Thanks

adevress:
Hi guys,

I would be for the inclusion of this API and as one line load / dump API.

This would allow to provide a numpy/pickle/joblib load/save style API which is very handy for any C++ tensor library ( xtensor, Eigen ) that can star to user HDF5 as any other serialization format.

Easy and simple API for easy and simple tasks are always good to take.

I would however rename the name of the header to something else.

tdegeus:
@adevress 

What name would you suggest?

It seems that the functions `exist` and `createGroups` potentially should be 'fixed' in the core implementation, see also #149 and #139. If we decide to do this, maybe you guys can tell me where, so that I can include these fixes in this PR.

adevress:
@tdegeus  Something in the spirit of the CURL "easy" API.

To not make it confusing with the HDF5 HL, something like <HighFive/H5Easy.hpp>

tdegeus:
Sounds good. I will update the PR once #152 is integrated. I will include tests and some documentation also.

ferdonline:
@tdegeus This contribution is nice for the new formats. The "improvements" (point 1 and point 2) should simply become part of the core. (since #152 was just merged we just miss point 2)

Regarding point 2. I like the idea of `size()` but eventually it can be implemented  simply using `std::accumulate` with `std::multiplies`. In turn, `shape()` I see it as just providing a different API for getDimensions() but doing the same. I believe that is not so important now.

For your point 3, please go ahead with the changes as suggested. I'd split the files though.

tdegeus:
@ferdonline #152 is very nice, that really smoothens usage. I will update the PR soon. 

(I just didn't understand by "I'd split the files though.")

ferdonline:
You have API and implementation for both formats all in the same file. I find it preferable to split, as you see throughout the project, even if some files are relatively small. It helps to track changes and possibly parallel work.

tdegeus:
If I understand you correctly I should emphasise that I already have a `bits/fname.hpp` and `fname.hpp`.

ferdonline:
The implementation (bits) gets quite long, so personally I'd create a folder and then have separation per output format (or namespace).
I believe to the user it could be nice to do specific import, e.g.:
```
#include <highfive/adapters/xtensor.hpp
```

but I'd invite other maintainers to give their thoughts on that.
It's also something we can decide later.

tdegeus:
I have overhauled this PR with the suggestions above. I have started documentation, but am not sure if it is in the right location. A readthedocs would make it simpler to document nicely. A point of discussion is testing, and in particular the optional Eigen and xtensor dependencies. Maybe you can give your opinion?

@ferdonline @adevress @tristan0x @henryiii

tdegeus:
It has been awfully quiet on the developers front...

include/highfive/H5Easy.hpp
```diff
@@ -0,0 +1,217 @@
...
+    inline void createGroupsToDataSet(HighFive::File& file, const std::string& path);
+
+    ///
+    /// Get the size of an existing DataSet in an open HDF5 file.
+    ///
+    /// @param file opened HighFive::File
+    /// @param path path of the DataSet
+    ///
+    /// @return size the size of the HighFive::DataSet
+    inline auto size(const HighFive::File& file, const std::string& path);
```
tristan0x: HighFive is C++11 compatible. I am afraid it does not support `auto` return type.

include/highfive/bits/H5Easy_Eigen.hpp
```diff
@@ -0,0 +1,190 @@
...
+ */
+#ifndef H5EASY_BITS_EIGEN_HPP
+#define H5EASY_BITS_EIGEN_HPP
+
+#include "../H5Easy.hpp"
+#include "H5Easy_misc.hpp"
+
+#ifdef HIGHFIVE_EIGEN
+
+namespace HighFive
```
tristan0x: Please follow coding style of existing code base, i.e:

```cpp
namespace A {
namespace B {

foo my_function() {
  instruction1; // no extra newline between every instruction
  if (test) {
    instruction2;
  }
}

} // namespace B
} // namespace A
```

README.md
```diff
@@ -80,6 +80,104 @@ See [src/examples/](src/examples/)  sub-directory for more infos
...
+    -   vectors (of standard types),
+    -   [Eigen::Matrix](http://eigen.tuxfamily.org) (optional, enabled by including Eigen before including HighFive, or by `#define HIGHFIVE_EIGEN` before including HighFive),
+    -   [xt::xarray](https://github.com/QuantStack/xtensor) and [xt::xtensor](https://github.com/QuantStack/xtensor) (optional, enabled by including xtensor before including HighFive, or by `#define HIGHFIVE_XTENSOR` before including HighFive).
+
+*   Direct access to the `size` and `shape` of a DataSet.
+
+Consider this example:
+
+```cpp
+#include <Eigen/Eigen>
```
tristan0x: Nice example, but it should be in a dedicated file in `src/examples/`, executed by CI instead.

include/highfive/bits/H5Easy_Eigen.hpp
```diff
@@ -0,0 +1,190 @@
...
+            template <class C, int Rows, int Cols, int Options>
+            void write(HighFive::DataSet& dataset, const Eigen::Matrix<C,Rows,Cols,Options>& data)
+            {
+                if (data.IsRowMajor)
+                {
+                    dataset.write(data.data());
+                }
+                else
+                {
+                    Eigen::Matrix<C,Rows,Cols,Eigen::RowMajor> tmp = data;
```
tristan0x: Is a copy required here?

tdegeus: Maybe the `Map` functionality of Eigen can provide a way out here.

tdegeus:
@tristan0x Thanks for your comments. Regarding the CI, see my question on how I should handle the optional Eigen and xtensor dependencies.

include/highfive/H5Easy.hpp
tdegeus: ...

That's really a pity.

include/highfive/bits/H5Easy_Eigen.hpp
ferdonline: We have the `.clang-format` file in the root, just apply the style.

README.md
ferdonline: Agreed. README.md should be a minimal introduction.

include/highfive/bits/H5Easy_misc.hpp
```diff
@@ -0,0 +1,97 @@
...
+        auto dataspace = dataset.getSpace();
+
+        auto dims = dataspace.getDimensions();
+
+        using T = decltype(dims)::value_type;
+
+        return std::accumulate(dims.begin(), dims.end(), 1, std::multiplies<T>());
+    }
+
+    inline auto size(const HighFive::DataSet& dataset)
```
ferdonline: I am not convinced on having these helper functions. To me it seems like going away from an OO paradigm. They very useful shortcuts though, I would implement them in the core H5DataSet (and probably internally cache the results). What do you think? @tristan0x

tristan0x: I think the load, dump, shape, and size are simple operations that deserve to be in this simple API.
On the other hand, I am not completely convinced for `groupName` and `createGroupsToDataSet`:
* `groupName` is a very basic function with no H5 specificities.
* `createGroupsToDataSet` is not meant to be used in simple use-cases like the load/dump functions for instance. So this operation should be done using the OO API.

include/highfive/bits/H5Easy_Eigen.hpp
tdegeus: @ferdonline Cool, do I have to run `clang-format -style=file include/highfive/H5Easy.hpp` from the project's root?

tristan0x: `-style=file` is redondant. `clang-format` looks for `.clang-format` in parent directories if not found where source is, so it can be run from anywhere within project directory.

use option `-i` option to inplace edit the file.

include/highfive/bits/H5Easy_misc.hpp
tristan0x: If `createGroupsToDataSet` is internal, then it should not be usable.

> Can you specify what you mean by "So this operation should be done using the OO API."
I meant that this helper is not necessary and if the user needs to do it, it should use what current API already provides.

tdegeus: I see. Is there a function that does exactly this in the API? I.e. recursively create all groups needed to create a DataSet. E.g. recursively create groups `/path/to/a/certain` if the input path to the DataSet is `/path/to/a/certain/dataset`?

include/highfive/bits/H5Easy_Eigen.hpp
tdegeus: @tristan0x Can you tell me if there is a way to write using to iterator (that loops over non-contiguous data)?

include/highfive/bits/H5Easy_vector.hpp
```diff
@@ -0,0 +1,105 @@
...
+{
+    namespace detail
+    {
+        namespace vector
+        {
+            template <class T, class E = void>
+            struct dump_impl
+            {
+                template <class C>
+                static HighFive::DataSet run(HighFive::File& file, const std::string& path, const std::vector<C>& data)
```
ferdonline: I find the implementation quite complicated. With such level of abstraction I would expect more code reuse. Eventually there is not the need for so many templates since it seems to me that T and C are related. Could you maybe explain better this design?
@tristan0x do you see simplifications that could be done here?

tdegeus:
Thanks a lot for your additions @ferdonline! I have build upon this with:
- Some minor bug-fixes that got introduced.
- Several minor implementation simplifications. 
- Some manual formatting updates, because the automatic re-formatting made the code quite unreadable. 

**To-do remains testing. I will try to get to it by Sunday, but should you want to work on it in the mean-time be my guest of course. Do there have to be special measures to test the Eigen and xtensor plugins?**

include/highfive/bits/H5Easy_misc.hpp
tdegeus: @ferdonline The problem is that `createGroup` throws if called for an existing group. Otherwise I would do what you suggest.

include/highfive/bits/H5Easy_Eigen.hpp
```diff
@@ -0,0 +1,190 @@
...
+            {
+                return detail::eigen::load_impl<Eigen::Matrix<T,Rows,Cols,Options>>::run(file, path);
+            }
+        };
+    }
+
+    template <class T, int Rows, int Cols, int Options>
+    inline HighFive::DataSet dump(HighFive::File& file, const std::string& path, const Eigen::Matrix<T,Rows,Cols,Options>& data, HighFive::Mode mode)
+    {
+        if ((mode == HighFive::Mode::Create) or (mode == HighFive::Mode::Overwrite and !file.exist(path)))
```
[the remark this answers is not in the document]
tdegeus: Done.

include/highfive/H5Easy.hpp
```diff
@@ -0,0 +1,217 @@
...
+#define HIGHFIVE_EIGEN
+#endif
+
+#ifdef HIGHFIVE_EIGEN
+#include <Eigen/Eigen>
+#endif
+
+namespace HighFive
+{
+    enum Mode
```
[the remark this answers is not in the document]
tdegeus: Done.

include/highfive/bits/H5Easy_misc.hpp
```diff
@@ -0,0 +1,97 @@
...
+#ifndef H5EASY_BITS_MISC_HPP
+#define H5EASY_BITS_MISC_HPP
+
+#include "../H5Easy.hpp"
+
+namespace HighFive
+{
+    namespace detail
+    {
+        inline auto error(const HighFive::File& file, const std::string& path, std::string message)
```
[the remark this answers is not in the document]
tdegeus: Thanks, much better, I was not aware! I think that in the spirit of `HighFive::Exception` it would be best to return a `HighFive::Exception` and throw where invoked. This is what I have now done.

include/highfive/bits/H5Easy_vector.hpp
tdegeus: Maybe it was not the most logical choice, but I kept the structure the same everywhere. The structure is useful for Eigen/xtensor. Got rid to the `C` where I could.

include/highfive/bits/H5Easy_Eigen.hpp
```diff
@@ -0,0 +1,153 @@
...
+        return {static_cast<size_t>(data.rows())};
+    } else {
+        return {static_cast<size_t>(data.rows()), static_cast<size_t>(data.cols())};
+    }
+}
+
+// write to open DataSet of the correct size
+// if the "Eigen::Matrix" is column-major reordering has to be done first; HDF5 is always row-major
+template <class C, int Rows, int Cols, int Options, int MaxRows, int MaxCols>
+void write(HighFive::DataSet& dataset, const Eigen::Matrix<C,Rows,Cols,Options,MaxRows,MaxCols>& data)
```
ferdonline: This line is way too long. It is typically a bad idea to do "custom" formatting. Of course clang-format sometimes might generate ugly things that we can fix, but here I wonder what you had against one argument per line. That's standard in this project.

include/highfive/bits/H5Easy_Eigen.hpp
```diff
@@ -0,0 +1,153 @@
...
+    }
+}
+
+// create DataSet and write data
+template <class T>
+struct dump_impl
+{
+    template <class C, int Rows, int Cols, int Options, int MaxRows, int MaxCols>
+    static HighFive::DataSet run(HighFive::File& file, const std::string& path,
+        const Eigen::Matrix<C,Rows,Cols,Options,MaxRows,MaxCols>& data)
```
ferdonline: Format: args just got same indentation as body. inconsistent and puzzling. Please keep the default.

tests/unit/tests_high_five_base.cpp
```diff
@@ -1406,3 +1411,128 @@ BOOST_AUTO_TEST_CASE(HighFiveRecursiveGroups) {
     // Using root slash
     BOOST_CHECK(file.exist(std::string("/") + DS_PATH));
 }
+
+BOOST_AUTO_TEST_CASE(H5Easy_scalar)
```
ferdonline: Please move the "H5Easy" tests to another source. `tests_high_five_base.cpp` is for core and is way too long already, should be refactored.

tests/unit/tests_high_five_base.cpp
```diff
@@ -1406,3 +1411,128 @@ BOOST_AUTO_TEST_CASE(HighFiveRecursiveGroups) {
...
+    BOOST_CHECK_EQUAL(b == b_r, true);
+    BOOST_CHECK_EQUAL(c == c_r, true);
+}
+
+#ifdef HIGHFIVE_XTENSOR
+BOOST_AUTO_TEST_CASE(H5Easy_extend1d)
+{
+    HighFive::File file("test.h5", HighFive::File::Overwrite);
+
+    for (std::size_t i = 0; i < 10; ++i)
```
ferdonline: In the new source apply the formatter please

tdegeus:
@ferdonline Thanks. I have updated the PR.

tests/unit/CMakeLists.txt
```diff
@@ -15,20 +15,41 @@ endif()
...
-    target_link_libraries(tests_high_five_cpp11_bin ${MPI_C_LIBRARIES})
+  target_link_libraries(tests_high_five_cpp11_bin ${MPI_C_LIBRARIES})
 endif()
 
 target_compile_options(tests_high_five_cpp11_bin PRIVATE "${HIGHFIVE_CPP_STD_FLAG}")
 
 add_test(NAME tests_high_five_cpp11 COMMAND tests_high_five_cpp11_bin)
 
+if(USE_EIGEN)
+  list(APPEND CMAKE_MODULE_PATH "/usr/share/cmake-2.8/Modules/")
```
tristan0x: Why is this `list(APPEND ...)` required? This path is already in the searched paths. Please remove it.

include/highfive/bits/H5Easy_scalar.hpp
```diff
@@ -0,0 +1,170 @@
...
+#define H5EASY_BITS_SCALAR_HPP
+
+#include "../H5Easy.hpp"
+#include "H5Easy_misc.hpp"
+
+namespace HighFive {
+
+namespace detail {
+
+namespace scalar {
```
ferdonline: are these xxx_impl methods supposed to be available elsewhere in the project? If not this should be a simple unnamed namespace.

tests/unit/CMakeLists.txt
tdegeus: The only reason that this is included is that, without this, the default build environment is failing to find Eigen.

tdegeus:
I am very much against putting the just two functions and one enum in a separate namespace and I will explain why. The easy-routines are specifically constructed to allow as simple usage as possible, providing accessibility to even to most inexperienced users. By putting the few functions in a separate namespace, the user is required to deal with two namespaces at all times, even for this easy-API (as the user has to provide a `HighFive::File`).

include/highfive/bits/H5Easy_xtensor.hpp
```diff
@@ -0,0 +1,135 @@
...
+{
+    return std::vector<size_t>(data.shape().cbegin(), data.shape().cend());
+}
+
+// create DataSet and write data
+template <class T>
+struct dump_impl
+{
+    template <class C>
+    static HighFive::DataSet run(HighFive::File& file, const std::string& path,
```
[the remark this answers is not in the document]
tdegeus: I will look into it.

include/highfive/bits/H5Easy_scalar.hpp
tdegeus: I don't understand your question, nor do I understand your suggestion. Sorry.

ferdonline: As it is, the entities defined inside this HighFive::detail::scalar namespace are visible outside the "source file" which is know as translation unit in c world. Since I believe they shall not be used anywhere else (they should be involked always from the "public" funcs you define in the bottom) they shoud  be in a special namespace which doesn't create external linkage. Please look up "unnamed namespaces".

ferdonline:
>the user is required to deal with two namespaces at all times, even for this easy-API (as the user has to provide a `HighFive::File`).

We understand that point but we must keep the core namespace clean, specially if we start receiving other contributions. To improve the situation we shall document it and suggest the user to do:
```
HighFive::Easy::load(HighFive::File(), ...)
```
or 
```
using H5Easy=HighFive::Easy;
H5Easy::load(HighFive::File(), ...)
```

I believe the situation would only change if this contribution became part of the core, something we decided not to do. 
@adevress @hernando Do you want do share your point of view?

tests/unit/CMakeLists.txt
ferdonline: I agree this is a hack that we should try to remove. I can try to look it up

include/highfive/bits/H5Easy_scalar.hpp
tdegeus: The idea of unnamed/anonymous namespaces it quite neat, I didn't know. However, this quiet opposes your earlier wish to separate in different files, one for each type. For consistency one would like to also place the functions in `bits/H5Easy_misc.hpp` in an unnamed namespace. In theory that is fine since with all the includes effectively one is creating one translation unit. But I do think that this is not very transparent. For this reason, and for the reason that `bits/H5Easy_scalar.hpp` has to be loaded first, I think that the entire Easy implementation should be in one file (as it originally was).

ferdonline: The suggestion would be to use the `details` namespace for internal but shared stuff, which is the case of your `easy_misc`. But again, we can drop the idea, as per the updated comment (i.e.: we are implementing stuff in headers)

ferdonline:
I don't see much of a problem for a user to type an extra short namespace. Consider even the libcurl example that inspired this API, functions are named `curl_easy_xxx`.
But mainly the no-go thing for me is the `Mode` class whose name is extremely generic and yet can only be used in a very particular case in a new part of the API. Otherwise I'm not against of having only `dump` and `load` in the main namespace, even though that's an exception for an high-level contribution which other maintainers would have to agree.

include/highfive/H5Easy.hpp
```diff
@@ -0,0 +1,212 @@
...
+#ifdef HIGHFIVE_EIGEN
+#include <Eigen/Eigen>
+#endif
+
+namespace HighFive {
+
+///
+/// Write mode for DataSets
+///
+enum class Mode
```
hernando: DumpMode

tdegeus: This I can live with. I will update the PR.

tests/unit/CMakeLists.txt
tristan0x: Could you please detail in which conditions you need this `APPEND CMAKE_MODULE_PATH`? at least the following information, Linux distribution, cmake version, eigen installation and version, cmake command to build highfive

tdegeus: For the HighFive's Travis/Linux environment for which no compiler is specified (i.e not `GCC_VERSION=7` or `CLANG_VERSION=...`).

I cannot reproduce a failure to find Eigen on any of my systems.

hernando:
Taking into account that Mode has been renamed to DumpMode, I don't have any particular feeling about having a different namespace or not for this API. In any case, I'd prefer something shorter such as EasyFive or H5Easy rather than HighFive::Easy (I would prefer having lower case namespaces everywhere as well, but I guess it's a bit late for that). Regarding the problem of having to remember 2 namespace, I think that's easily solvable adding using declarations inside the Easy namespace for the core types that it requires (e.g. `using HighFive::File`). This way, users of the easy API won't need to deal with the HighFive namespace at all.

tests/unit/CMakeLists.txt
hernando: That may be solvable with an environmental variable configured in the `.travis.yml` file rather than hardcoding it here.

henryiii:
Adding things to an existing namespace in separate files is not bad (after all, all the standard library files add things to `std`), the important thing is that you don't conflict (that's what namespaces are for). So as long as the new names are not things you eventually want to use in HighFive core, I would think using the same namespace would be simpler and okay to do. Just my 2 cents.

ferdonline:
@tdegeus thanks for the change. I also agree with all suggestions here: (thanks @hernando) in case of having a separate namespace it should be short and should include the core namespace so that, as mentioned, we solve in a simple way the "using two namespaces problem".
I gave it some thought and put the argument side-by-side:

Separate namespace:
- It makes it very clear it is at a high-level layer, and the core will not need it.
- The functions have a generic different signature: Are not OO, receiving the h5File object to work

Same namespace:
- User doesn't need to change namespaces.
- No changes from current implementation

I have a light preference to separate while I see some light preference to keep as is. @tristan0x  I would like to hear from you to disambiguate. Since this is API it would be great to agree and not change later.

tristan0x:
Thank you for summarizing this topic @ferdonline 
I am also in favor of separating namespaces, because IMHO:
* what makes h5easy simpler to use is also that you don't have to deal with all HighFive symbols.
* this PR provides an extension that should remain out of the core.

tdegeus:
@adevress @henryiii @ferdonline @tristan0x  

I have doubted if and how to respond as I have given my opinion before. Nonetheless I want to share some thoughts:

* The philosophy of the easy-API is primarily to do standard and often repeating operations in a single line, facilitating well readable and well maintainable end-user-code. 
* Beyond making standard things easier for the user, for me the easy-API is meant as an invitation into HighFive should she/he want or need to (this is also the way how I sometimes use more detailed features of `h5py` but mostly use simple read and write). I think that the mental barrier is higher when the easy-API is an extension, fork, or wrapper. 
* Making the easy-API part of the main namespace allows a seamless transition between doing something standard and adding some customisation. For example, with `dump` the `DataSet` is returned such that the user could add metadata. With two namespaces this would directly become less transparent. One could, of course, add decelerations in an easy-namespace, but I doubt that the user will be less confused. Why would she/he expect the objects from a different namespace to be same?
* The reason that the easy-API's functions are not object-oriented, but free functions, is to allow customised extensions as overloads. In contrast to classes that cannot be extended, the free functions allow the user to just add overloads in her/his code without having to bother to invent a custom API, or having the need to add existing functionality to such a custom API.
* Like @henryiii says, the `std` namespace (but also many other namespaces) feature a large number of functions without causing collisions or confusion, quite the contrary I think.

To understand better your opposing views it would help if you would explain why the "core" needs special protection. There is probably some philosophy that I am not aware of.

tests/unit/CMakeLists.txt
tdegeus: I'm all for it! I could not find how to, do you know?

ferdonline:
@tdegeus I believe this issue has gone far in terms of time. We contribute to it on a best-effort basis and consider that your contribution doesn't follow previous code conventions (so pls dont compare to `std`), nor was critical to our roadmap. In the past such contributions would be totally refactored out and incorporated in the core by some main developer (been there!). So yes, we have to protect the core out of inconsistencies to make it a better world for any maintainer with very limited time. 
I believe the solution with an own namespace was a good compromise. My position is that, besides the technical stuff, your API is by definition not core API. If you or any other maintainer wants to refactor that, fine with me.

tdegeus:
The H5Easy-API is now on the H5Easy namespace. `H5Easy::File` is introduced as derived class of `HighFive::File` to permit the user to use only one namespace.

henryiii:
Can't you just export `using namespace HighFive::File` inside `H5Easy` instead of making that a new class? And you could even export the entire namespace inside `H5Easy` with `using namespace HighFive` so a user only needs one namespace...

tdegeus:
@henryiii I also liked that better, but I thought the worry was conflicts ;). I now refactored to `using HighFive` such that the `H5Easy` contains everything from HighFive, and the user will never have to switch.

Also the Xcode-CI has been broken by something that is not related to this PR.

tdegeus:
Also the CI now errors over an unused variable warning in the example. I'm not sure how to treat that. Removing the variable reduces the readability, adding some void operations does too.

hernando:
Don't assign the result of the function calls to anything. It should be obvious what you're trying to show.

tristan0x:
Please stop force pushing to your branch @tdegeus 
Consider using `git push --force-with-lease` instead

tristan0x:
Hello @tdegeus 
I noticed you own the copyright of H5Easy files. I am very sorry, but we cannot accept a contribution we don't own. Can you please set `Adrien Devresse <adrien.devresse@epfl.ch>` as copyright holder?

tdegeus:
Changed. Also the CI has to be reconfigured, or Eigen (and probably xtensor) have to be excluded.

tristan0x:
@tdegeus your 2 last commits didn't fix Travis, it disabled testing of eigen.
I am working on it, please stop editing .travis.yml

tdegeus:
PING

ferdonline:
@tristan0x I'm ok with the current state, improvements may be done separately. Are you still planning some changes here, or could we merge?

PR #147, opened 2018-12-04
Added missing default constructor HighFive::File

tdegeus:
This constructor is needed for example to use `HighFive::File` as a class variable. E.g.

```cpp
class Foo
{
private:

    HighFive::File data;

public:

    Foo(const std::string &fname)
    {
        data = HighFive::File(fname, HighFive::File::ReadOnly);
    }
};

### DIFF
--- a/include/highfive/H5File.hpp
+++ b/include/highfive/H5File.hpp
@@ -40,9 +40,11 @@ class File : public Object,
     static const int Create = 0x10;
     /// Derived open flag: common write mode (=ReadWrite | Create | Truncate)
     static const int Overwrite = Truncate;
-    /// Derived open flag: Opens RW or exclusivelly creates
+    /// Derived open flag: Opens RW or exclusively creates
     static const int OpenOrCreate = ReadWrite | Create;
 
+    // Default constructor
+    explicit File() = default;
 
     ///
     /// \brief File

### CODE REVIEW

**Comment by hernando** [comment]
Not having a default constructor is by-design. In principle there are alternatives, but typically they require pointers. On the other hand, I would make the class movable.

**Comment by tdegeus** [comment]
@hernando What would that imply for my example?

**Comment by tristan0x** [comment]
I agree with @hernando : default constructor is missing by-design.
Why don't you use constructor brace initializer in your example?

```cpp
class Foo
{
private:

    HighFive::File data;

public:

    Foo(const std::string &fname)
      data(fname, HighFive::File::ReadOnly)
    { }
};
```

tdegeus:
@tristan0x Because this example is simplified ;) In fact I derive `fname` from the input only inside the constructor, so I cannot use the brace initializer. Do you have other nice solutions.

Also, can you enlighten why there is no default constructor?

tdegeus:
@ferdonline I'm not very successful with the `unique_ptr` and the optional argument, as this

```cpp
std::unique_ptr<HighFive::File> data = std::make_unique<HighFive::File>(fname, HighFive::File::Overwrite);
```

results in a linker error:

```bash
Undefined symbols for architecture x86_64:
  "HighFive::File::Overwrite", referenced from:
      Foo::Foo(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) in test-90f23c.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

hernando:
The default constructor is not included because the design follows RAII principles. I wouldn't make an exception just in this class and as @ferdonline pointed out, having to deal with the uninitialized state introduced by the default constructor in all classes requires changes in too many places.

By using pointers I was referring to either std::unique_ptr or std::shared_ptr. I'm surprised about the linking error, I'll try to reproduce it myself, but if you can provide a complete reproducer that would help debugging.

Initially, I was not happy about not having a default constructor myself, but with unique_ptr I've been able to stay without out. Having a movable object would allow using lambda functions when direct default initialization is not possible.

tdegeus:
I have converted the PR to an issue concerning the `static const` problem.

hernando:
I was assuming HighFive::File::Overwrite was an enum value. The most probably solution is using constexpr instead of static

tdegeus:
@hernando In my personal opinion I would rather change it to an `enum`, as it is much more robust also in terms of avoiding unexpected inputs. The current implementation seems a bit to 'reinvent the wheel'.

hernando:
I think it's not an enum because it's a bitmask and enums as bitmasks are not a good idea, but @ferdonline can provide the reasons behind that better than me.

PR #152, opened 2018-12-05
createGroup and exist can be recursive

### LINKED ISSUE #139: Recursively create groups

### LINKED ISSUE #149: exist not recursive
The following

```cpp

int main()
{
    HighFive::File file("test.h5", HighFive::File::Overwrite);
    file.exist("/path/to/group");

    return 0;
}
```

*throws* in-stead of returning `false`. The output:

```bash
HDF5-DIAG: Error detected in HDF5 (1.10.4) thread 0:
  #000: H5L.c line 815 in H5Lexists(): unable to get link info
    major: Links
    minor: Can't get value
  #001: H5L.c line 3095 in H5L__exists(): path doesn't exist
    major: Links
    minor: Object already exists
  #002: H5Gtraverse.c line 851 in H5G_traverse(): internal path traversal failed
    major: Symbol table
    minor: Object not found
  #003: H5Gtraverse.c line 741 in H5G__traverse_real(): component not found
    major: Symbol table
    minor: Object not found
libc++abi.dylib: terminating with uncaught exception of type HighFive::GroupException: Invalid link for exist()  (Symbol table) Object not found
```

ferdonline:
In recursive mode group paths are correctly handled when checking
with exist(), and intermediate groups are created as required

Fixes #139 #149


include/highfive/bits/H5Node_traits_misc.hpp
```diff
@@ -180,6 +197,35 @@ inline bool NodeTraits<Derivate>::exist(const std::string& node_name) const {
     return (val > 0 );
 }
 
+template <typename Derivate>
+inline bool NodeTraits<Derivate>::exist(const std::string& group_path,
+                                        bool recursive) const {
+    if (recursive) {
+        std::string path = group_path;
```
adevress: You do not really need memory allocation nor string assignment here. You can do the "view" and the recursion entirely with Iterators.

include/highfive/bits/H5Node_traits_misc.hpp
```diff
@@ -180,6 +197,35 @@ inline bool NodeTraits<Derivate>::exist(const std::string& node_name) const {
...
+        size_t split_pos = path.find('/');
```
adevress: CF previous comment, with iterator you can use std::find and avoid to play with offset.

include/highfive/bits/H5Node_traits_misc.hpp
```diff
@@ -180,6 +197,35 @@ inline bool NodeTraits<Derivate>::exist(const std::string& node_name) const {
...
+        if (split_pos == 0) {
+            if (!std::is_same<Derivate, File>::value) {
+                throw GroupException("Attempt to check absolute path from within a group");
+            }
+            // Search again, skip leading '/'
+            path = group_path.substr(1);
+            split_pos = path.find('/');
+        }
+        if (split_pos != std::string::npos) {
+            std::string cur_node_name = path.substr(0, split_pos);
```
adevress: same here

ferdonline:
@adevress! Welcome back! :D Thanks for the review! btw if you have some time give a check at #78 since that's a bit outside my area of comfort ;)

include/highfive/bits/H5Node_traits.hpp
```diff
@@ -107,13 +114,21 @@ class NodeTraits {
...
     /// false
     bool exist(const std::string& node_name) const;
 
+    /// \brief Checks if a node / path exist under the current node / group.
+    /// \param node / path name to check
+    /// \param Whether recursive checking shall be activated. When disabled
+    ///        plain H5Lexists is used, which fails if a part of the path 
+    ////       doesnt exist
+    /// \return true if the given node name / path exists
+    bool exist(const std::string& node_path, bool recursive) const;
```
tristan0x: Does it make sense to only have one `exist` member method with an `recursive` parameter with default value `false`?

```cpp
bool exist(const std::string& node_path, bool recursive = false) const;
```

Besides Doxygen parameter documentation is not consistent between the 2 `exist` member methods.

tdegeus:
Nice! Also fixes #139.

include/highfive/bits/H5Node_traits.hpp
hernando: According to this page: https://support.hdfgroup.org/HDF5/doc/H5.user/Groups.html, which I assume to be latest version of the library, slashes are not allowed as part of any component name. I would make have a single function that handles recursion by default. That shouldn't have any noticeable performance impact.

ferdonline: Thanks @hernando for the link, I was missing verifying that info.
Information in https://support.hdfgroup.org/HDF5/doc/RM/RM_H5L.html#Link-Exists is a bit misleading. They even made `exist("/")` at the file and root node level to return 1.
I agree that having overloads might not be very elegant. Putting together your suggestions I feel we should have a different function name for the base version to become obviously non-recursive, like `exist_node` and turn `exist` always recursive. What you think?

ferdonline: Reading  bit more Groups.html it is understandable that names can be composed and have slashes, and H5Lexists will respect that. The only problem is indeed when middle components don't exist, where instead of returning false it throws an exception.
Without knowing the motivation behind that, we could simply fix that case. 
I'll make some changes.

ferdonline:
@tristan0x @hernando Do you agree with the current solution?

include/highfive/bits/H5Node_traits.hpp
```diff
@@ -116,7 +117,17 @@ class NodeTraits {
...
   private:
     typedef Derivate derivate_type;
+
+    ///
+    /// \brief create a new group with the name group_name
+    /// \param group_name
+    /// \return the group object
+    Group _createGroup(const std::string& group_name);
+
+    // A wraper over the low-level H5Lexist
```
tristan0x: s/wraper/wrapper/

ferdonline: fixed :)

include/highfive/bits/H5Node_traits_misc.hpp
```diff
@@ -105,6 +106,24 @@ inline Group NodeTraits<Derivate>::createGroup(const std::string& group_name) {
...
+        if (split != group_path.end()) {
+            std::string cur_node_name(group_path.begin(), split);
+            Group cur_g = exist(cur_node_name)? getGroup(cur_node_name)
+                                              : createGroup(cur_node_name);
+            return cur_g.createGroup(std::string(split + 1, group_path.end()),
+                                     true);
+        }
+    }
+    // No recursion
+    return _createGroup(group_path);
```
hernando: I'd insist on not having the recursive parameter and write this:

    const auto split = group_path.find_first_of('/');
    if (split != std::string::npos) {
        const auto cur_node_name = group_path.substr(0, split);
        Group cur_g = exist(cur_node_name)? getGroup(cur_node_name)
            : createGroup(cur_node_name);
        return cur_g.createGroup(group_path.substr(split + 1));
    }
    else
        return _createGroup(group_path);

That's less code, fewer parameters and less documentation.

You also have to deal with the case of a group name with a trailing slash or two consecutive slashes. None of the cases should be accepted, it should throw and in my opinion no group should be created if the group name is ill formed (for exception safety reasons).

tdegeus: I agree on the trailing slash(es) that should throw. However, based on HDF5's documentation as referenced above it should be concluded that any number of repeated slashes are legal and should be simply interpreted as a single slash.

tdegeus: On second thought, I am even doubting about the trailing slash(es). The unix command `mkdir /path/to/new/dir/` would also be acceptable, I see no reason why it should be any different here.

hernando: The trailing `/` still has to be handled correctly by the implementation, and right now it's not the case as it tries to create a leaf group with an empty name.
I wouldn't take other tools as an example as I'm aware that some UNIX tools behave differently when there's a trailing / and the end. For example `ln -s foo bar`and `ln -s foo bar/` don't mean the same. On the other hand, h5 command line tools accept trailing slashes, so I won't oppose it.

ferdonline: Hi guys. Thanks for your input.
My reasons to prefer recursive off by default are:

- we dont change hdf5 lib way of working which could mislead people
- dont break compat
- mimick better filesystem mkdir

in terms of code size we only saved 1 `if`. I'd care more for the whole indentation, which might be possible to improve.

@hernando you raised an important point, that we should leave things clean if it fails. I also like the idea of validating the path before blindly start the process. I'll try to find some time to work on it

hernando: I was about to reason in favor of having recursive on by default, but before that I wanted to test what happens if you try to create a group with a slash in the name in the current master. To my surprise, I've discovered that there is a test for this, that means that nested groups are probably handled by the HDF5 lib already and this change is unnecessary. Look in test_high_five_base.cpp 247

hernando: And according to my last observation, maybe that property is on by default.

ferdonline: in test_high_five_base.cpp 247 they have already GROUP_CREATED by line 245. 
I'm pretty sure recursive is off by default, that's also why so many people complain about this and think hdf5 doesnt do it (including me).

hernando: I didn't pay attention to that. Then I'll write back what I didn't send yesterday.

I'd insist on making recursive the default and only behaviour (either with H5Pset_create_intermediate_group or manually implemented, I don't care). Consider that slashes are not allowed in group names, so this code:

    auto foo = file.createGroup("/root/foo");

fails at the moment. Adding the recursive parameter won't change the fact that any casual reading of code using HighFive which hasn't read the documentation (and very little people read docs these days, they only read stackoverflow) will be surprised about this behaviour, moreover if you consider that h5py creates the intermediate groups.

My point here is that:

> * we dont change hdf5 lib way of working which could mislead people

Is a non-reason. You are not going to mislead people, when in fact people are complaining that `file.createGroup("/root/foo")` doesn't work. Having to add the additional bool parameter doesn't make the code intent readable either, because there's no way to understand what the unnamed boolean parameter (which are always discouraged) without reading the code. I always advocate for self-documenting code, which one you think is more obvious `file.createGroup("/root/foo")` or  `file.createGroup("/root/foo", true)`?

> * dont break compat

It's a another non-reason, as you're relaxing the preconditions, not changing the semantics. Old code will still work the same. The only difference is that what before was an exception that probably nobody was handling, but rather avoiding, now would work by default.

tdegeus: Couldn't agree more, the recursive version should be the default and only. If a user specifically wants to avoid this, she/he can always call `exist` before calling `createGroup`. The same argumentation holds for `exists` by the way.

ferdonline: Guys, I think we are making a big deal out of this option. I don't care, one way or the other. But for some reason many libs don't do recursive magic by default to be safe. And if people don't read the docs where a signature has a `recursive=false`, I'm sorry maybe they shouldn't program :p
@adevress @tristan0x What is your view? Yes (recursive default) or No? (we dont need more arguments)

arsenius7: My 2 cents: it depends where one puts HighFive on the scale between "wrapper around HDF5 C interface" and "also, simplify usage for most often usecases at the cost of diverging from HDF5 C interface defaults".
My personal (slight) preference would be the latter one.
Especially for `exists`, I can't imagine why would I want an exception because of a missing intermediate group (for `createGroup` and `mkdir` there are different approaches out there).

ferdonline:
@tristan0x @hernando Is there anything else you see which could be improved?

ferdonline:
@tristan0x I dont really understand why it succeeded before and fails now with native OSX. Is it possible to use a more recent image? xcode clang is currently v10 and this is using v9. I can't reproduce it.

include/highfive/bits/H5Node_traits_misc.hpp
```diff
@@ -94,10 +95,15 @@ NodeTraits<Derivate>::getDataSet(const std::string& dataset_name,
 }
 
 template <typename Derivate>
-inline Group NodeTraits<Derivate>::createGroup(const std::string& group_name) {
+inline Group NodeTraits<Derivate>::createGroup(const std::string& group_name,
+                                               bool parents) {
+    RawPropertyList lcpl(Properties::LINK_CREATE);
+    if(parents) {
```
hernando:
```suggestion
    if (parents) {
```

hernando: I would be helpful to have a clang-format file to make all style consistent (even if clang-format does ugly things sometimes).

include/highfive/bits/H5PropertyList_misc.hpp
```diff
@@ -15,11 +15,55 @@
...
+        return H5P_OBJECT_COPY;
+    case Properties::LINK_CREATE:
+        return H5P_LINK_CREATE;
+    case Properties::LINK_ACCESS:
+        return H5P_LINK_ACCESS;
+    default:
+            HDF5ErrMapper::ToException<PropertyException>(
+                "Unsupported property list type");
+    }
+    return -1;
```
hernando: I wonder whether ToException should become a macro to avoid the spurious return -1 at the bottom.  Adding [[noreturn]] to the function declaration may also suppress the compilation warning but I need to test it (see https://en.cppreference.com/w/cpp/language/attributes/noreturn).

hernando: In fact that works, but you have to be careful and place [[ noreturn ]] at the very beginning of the function declaration, otherwise you get a warning from gcc. I wonder whether we should add it anyway as it's C++11 specific.

ferdonline: IMHO it's avoiding one slightly ugly thing with another (macros) where the ugliness gets hidden. I'd prefer to keep it obvious.

PR #156, opened 2018-12-10
More compiler flags

tristan0x:


src/examples/boost_multi_array_2D.cpp
```diff
@@ -18,8 +18,8 @@ using namespace HighFive;
 
 const std::string FILE_NAME("boost_multiarray_example.h5");
 const std::string DATASET_NAME("dset");
-const size_t size_x = 10;
-const size_t size_y = 3;
+const long size_x = 10;
```
ferdonline: Feeling puzzled. It doesnt warn without the L like `10l`
Anyway why not use the good-old & fast `int` type for small loops?

tests/unit/tests_high_five_base.cpp
```diff
@@ -114,7 +114,7 @@ struct ContentGenerate<std::string> {
         std::string random_string;
         std::mt19937_64 rgen;
         rgen.seed(88);
-        std::uniform_int_distribution<int> int_dist(0, 1000);
+        std::uniform_int_distribution<unsigned int> int_dist(0, 1000);
```
ferdonline: For consistency, all unsigned's with or without `int` (or uint :)

tests/unit/tests_high_five_base.cpp
```diff
@@ -995,7 +995,7 @@ void MultiArray3DTest() {
     std::ostringstream filename;
     filename << "h5_rw_multiarray_" << typeid(T).name() << "_test.h5";
 
-    const size_t size_x = 10, size_y = 10, size_z = 10;
+    const long size_x = 10, size_y = 10, size_z = 10;
```
ferdonline: int?

include/highfive/H5PropertyList.hpp
```diff
@@ -147,15 +144,15 @@ class Caching {
     /// details.
     Caching(const size_t numSlots,
             const size_t cacheSize,
-            const double w0 = H5D_CHUNK_CACHE_W0_DEFAULT)
+            const double w0 = static_cast<double>(H5D_CHUNK_CACHE_W0_DEFAULT))
```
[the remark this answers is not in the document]
tristan0x: IMHO one should always use proper type conversion mechanisms provided by C++ instead of c-style cast that can convert carrots to apples without complaining.

include/highfive/bits/H5Annotate_traits_misc.hpp
```diff
@@ -80,7 +80,7 @@ inline size_t AnnotateTraits<Derivate>::getNumberAttributes() const {
         HDF5ErrMapper::ToException<AttributeException>(std::string(
             "Unable to count attributes in existing group or file"));
     }
-    return res;
+    return static_cast<size_t>(res);
```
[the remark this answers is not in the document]
tristan0x: H5 uses negative values to handle error cases, which are translated to C++ exceptions in this member method. So the returned value is always positive. To me `size_t` is the proper type for this.

include/highfive/bits/H5File_misc.hpp
```diff
@@ -22,8 +22,8 @@ namespace {
 
 // libhdf5 uses a preprocessor trick on their oflags
 // we can not declare them constant without a mapper
-inline int convert_open_flag(int openFlags) {
-    int res_open = 0;
+inline unsigned convert_open_flag(unsigned openFlags) {
```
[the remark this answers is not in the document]
tristan0x: `unsigned` is specified in the C++ standard. I am against introducing another type forcing the user to take a look at its definition to ensure what it is.

include/highfive/bits/H5Converter_misc.hpp
```diff
@@ -98,7 +98,7 @@ single_buffer_to_vectors(typename std::vector<T>::iterator begin_buffer,
                          std::vector<T>& vec_single_dim) {
     const size_t n_elems = dims[current_dim];
     typename std::vector<T>::iterator end_copy_iter =
-        std::min(begin_buffer + n_elems, end_buffer);
+        std::min(begin_buffer + static_cast<long>(n_elems), end_buffer);
```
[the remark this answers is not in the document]
tristan0x: Again, I am against C-style cast in C++.

include/highfive/H5PropertyList.hpp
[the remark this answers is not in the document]
ferdonline: I agree, my thought was that eventually we could use the short syntax only for primitive types. But it's fine for me.

include/highfive/H5File.hpp
```diff
@@ -28,17 +28,17 @@ class File : public Object,
  public:
     enum : int {
         /// Open flag: Read only access
-        ReadOnly = 0x00,
+        ReadOnly = 0x00u,
```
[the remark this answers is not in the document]
tristan0x: I changed to `unsigned`.

include/highfive/bits/H5Node_traits_misc.hpp
```diff
@@ -140,12 +140,12 @@ inline std::string NodeTraits<Derivate>::getObjectName(size_t index) const {
         HDF5ErrMapper::ToException<GroupException>(
             "Error accessing object name");
     if (length <= maxLength)
-        return std::string(buffer, length);
-    std::vector<char> bigBuffer(length + 1, 0);
+        return std::string(buffer, static_cast<std::size_t>(length));
```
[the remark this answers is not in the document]
tristan0x: No that wasn't intentional. Thank you for spotting this one!

src/examples/boost_multi_array_2D.cpp
tristan0x: It doesn't warn because an integer fits in a long, and the compiler makes the type conversion implicitly.

tristan0x: Besides, I changed to `int` as suggested

tests/unit/tests_high_five_base.cpp
tristan0x: absolutely!

tests/unit/tests_high_five_base.cpp
```diff
@@ -78,7 +78,7 @@ struct ContentGenerate {
 
     T operator()() {
         T ret = _init;
-        _init += _inc;
+        _init = static_cast<T>(_init + _inc);
```
ferdonline: Sorry a last remark here. The compiler used to warn on an assignment but doesn't warn if we sum? When we sum it should anyway use the longest type and so the cast is not needed.

tristan0x: gcc emits a warning in such case because as addition might overflow, the compiler stores addition result into an int, which then fails when trying to put it in `k`:

```
cat <<EOF | g++ -Wconversion -x c++ -
int main() {
        unsigned char i = 0, j = 0;
        unsigned char k = i + j;
        return k;
}
EOF
<stdin>: In function ‘int main()’:
<stdin>:3:22: warning: conversion to ‘unsigned char’ from ‘int’ may alter its value [-Wconversion]
```

tristan0x:
I squashed last commits and it discarded your review. Can you please re-approve @ferdonline

ferdonline:
thanks @tristan0x !

Issue #112, opened 2018-05-18
Collective read


iamholger:
Hi,
I am happily using HighFive at NERSC to read HDF5 files. Currently, I am encountering a
performance bottleneck with parallel read and was wondering if it is possible to configure
HighFive to do collective read.

Thank you very much,
Holger

roblatham00:
Can I expand the scope of this issue to collective read and write?   https://github.com/BlueBrain/HighFive/blob/master/include/highfive/bits/H5Slice_traits_misc.hpp#L252  passes a hard coded H5P_DEFAULT to H5Dwrite, which means no matter what property list folks set up it will never request collective I/O.  Same is true for the call to H5Dread in `SliceTraits<Derivate>::read()` https://github.com/BlueBrain/HighFive/blob/master/include/highfive/bits/H5Slice_traits_misc.hpp#L214

1uc:
Thank you for reporting the issue. I need a similar feature myself. Therefore, this will likely get looked at and fixed in the next few weeks.

PR #157, opened 2018-12-17
Refactoring property lists

ferdonline:
To address requests/issues related to Properly lists (e.g. #112) I am refactoring a bit the Properties code. 
This change is the basis to have easily all propertyList types. 

Eventually we could extend the concept further to group several properlylists in an Options object, so that, e.g. DatasetAccess, Create, Xfer that can go to dataset create together as a single object

@hernando since you implemented this part in the first place, do you foresee any problem with this implementation?

### DIFF
--- a/include/highfive/bits/H5FileDriver_misc.hpp
+++ b/include/highfive/bits/H5FileDriver_misc.hpp
@@ -42,18 +42,13 @@ private:
   Info _info;
 };
 
-// depecrated, use Properties(Properties::FILE_ACCESS) instead
-class DefaultFileDriver : public FileDriver {
-};
-}
-
-// file access property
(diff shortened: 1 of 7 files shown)


include/highfive/bits/H5FileDriver_misc.hpp
```diff
@@ -42,18 +42,13 @@ class MPIOFileAccess
   Info _info;
 };
 
-// depecrated, use Properties(Properties::FILE_ACCESS) instead
-class DefaultFileDriver : public FileDriver {
-};
```
hernando: Keep for backwards compatibility?

ferdonline: @adevress Any notes on this?

PR #159, opened 2019-01-14
Set defaults for non-install or non-standard options to FALSE

tdegeus:


tdegeus:
Fixes #196

tdegeus:
I would be in favour of at l

PR #163, opened 2019-02-07
Fixed several minor issues for newer cmake

emmenlau:
This PR adds a few minor improvements in cmake, and removes some complaints from cmake 3.13.4 mostly on MSVC. Please review.


CMakeLists.txt
```diff
@@ -1,11 +1,22 @@
-project(HighFive)
 cmake_minimum_required(VERSION 3.0)
 
-enable_testing()
+if(POLICY CMP0048)
+  cmake_policy(SET CMP0048 NEW)  # package version behavior added in cmake 3.0
```
tristan0x: I do not understand why you enable these policies explicitly. To my understanding, they are already enabled if available.

emmenlau:
Actually I think the default for policies is not to set them unless a very new cmake version is requested in CMakeLists.txt. I maybe wrong though. But cmake 3.13.4 (current newest) printed the following warnings:

```
CMake Warning (dev) at CMake/config/CompilerFlagsHelpers.cmake:31 (elseif):
  Policy CMP0054 is not set: Only interpret if() arguments as variables or
  keywords when unquoted.  Run "cmake --help-policy CMP0054" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  Quoted variables like "MSVC" will no longer be dereferenced when the policy
  is set to NEW.  Since the policy is not set the OLD behavior will be used.
Call Stack (most recent call first):
  CMake/config/ReleaseDebugAutoFlags.cmake:36 (include)
  CMakeLists.txt:15 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.
```
and
```
CMake Warning (dev) at CMakeLists.txt:31 (find_package):
  Policy CMP0074 is not set: find_package uses <PackageName>_ROOT variables.
  Run "cmake --help-policy CMP0074" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  CMake variable HDF5_ROOT is set to:
    D:/Debug

  For compatibility, CMake is ignoring the variable.
This warning is for project developers.  Use -Wno-dev to suppress it.
```
Both warnings where fixed by explicitly setting the policies to `NEW`. Should I investigate more or does that seem reasonable?

tristan0x:
Can you please share with me how you obtain these warnings?
I can't manage to reproduce them.

```sh
$ mkdir _build
$ cd _build
$ cmake --version
cmake version 3.13.4

CMake suite maintained and supported by Kitware (kitware.com/cmake).
$ cmake -Wdev -Wdeprecated ..
-- The C compiler identification is GNU 5.5.0
-- The CXX compiler identification is GNU 5.5.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Setting build type to 'RelWithDebInfo' as none was specified.
-- HDF5: Using hdf5 compiler wrapper to determine C configuration
-- Configuring done
-- Generating done
-- Build files have been written to: /goinfre/src/github.com/BlueBrain/HighFive/_build
$
```

emmenlau:
I can only guess right now because I lack the time for thorough investigation. Here are my guesses:

1. The policy `CMP0054` complaint about the string `"MSVC"`. I received this warning exclusively on Windows with Visual Studio Compiler. According to the cmake documentation (https://cmake.org/cmake/help/latest/variable/MSVC.html), `MSVC` happens to be a cmake variable that evaluates to `true` when the compiler is some version of Microsoft Visual C++. My guess is that this warning (and the problem) are specific to Windows and the Visual Studio Compiler. Do you think that may be possible?
2. The policy `CMP0074` warning was likely caused by me setting `-DHDF5_ROOT=...` or `-DBOOST_ROOT=...`. Can you try if this causes the warning for you too?

For what its worth, here are my cmake options:
```
cmake "D:/builds/HighFive" \
  -G"Ninja" \
  -DCMAKE_VERBOSE_MAKEFILE="ON" \
  -DCMAKE_PREFIX_PATH="D:/artifacts/Debug" \
  -DCMAKE_INSTALL_PREFIX="D:/artifacts/Debug" \
  -DCMAKE_INSTALL_LIBDIR="lib" \
  -DCMAKE_INSTALL_DATADIR="lib" \
  -DCMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY="ON" \
  -DCMAKE_BUILD_TYPE="Debug" \
  -DCMAKE_POSITION_INDEPENDENT_CODE="ON" \
  -DBUILD_SHARED_LIBS="OFF" \
  -DBoost_USE_STATIC_LIBS="ON" \
  -DBOOST_ROOT="D:/artifacts/Debug" \
  -DHDF5_USE_STATIC_LIBS="ON" \
  -DHDF5_ROOT="D:/artifacts/Debug" \
  -DHDF5_FIND_DEBUG="ON" \
  -DUSE_BOOST="ON" \
  -DHIGHFIVE_UNIT_TESTS="OFF" \
  -DHIGHFIVE_EXAMPLES="OFF" \
  -DHIGHFIVE_PARALLEL_HDF5="OFF"
```

CMakeLists.txt
henryiii: I would recommend following [my advice here](https://cliutils.gitlab.io/modern-cmake/chapters/basics.html), and doing something like this:

```cmake
cmake_minimum_required(VERSION 3.0)

if(${CMAKE_VERSION} VERSION_LESS 3.13)
    cmake_policy(VERSION ${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION})
else()
    cmake_policy(VERSION 3.13)
endif()
``` 

Then test it with the highest version listed (3.13) and make sure it works. When new CMakes come out, you can update the number here after testing the new version. If you find you depend on any old policies, try to fix the issue. Piecemeal setting policies is not a good idea.

This is now supported in CMake as a language feature, but due to a MSVC error, keeping it written this way is probably best for now.

CMakeLists.txt
```diff
@@ -1,11 +1,22 @@
...
+  cmake_policy(SET CMP0054 NEW)  # Only interpret if() arguments as variables or keywords when unquoted
+endif()
+if(POLICY CMP0074)
+  cmake_policy(SET CMP0074 NEW)  # find_package behavior added in cmake 3.12
+endif()
 
 set(HIGHFIVE_VERSION_MAJOR 2)
 set(HIGHFIVE_VERSION_MINOR 0)
 
+project(HighFive VERSION "${HIGHFIVE_VERSION_MAJOR}.${HIGHFIVE_VERSION_MINOR}")
```
henryiii: Setting the version automatically creates these variables for you (at least, `HighFive_VERSION_MAJOR`, etc, that's the point). If you rename the project to HIGHFIVE here, then `HIGHFIVE_VERSION_MAJOR` will be set.

CMakeLists.txt
emmenlau: Very interesting, I was not aware of this! But it seems this may require more extensive testing of cmake versions? At least I do not feel comfortable to set any specific version as supported. But please fork my PR and exchange this part?

emmenlau:
I have found and fixed another issue with C++11 detection. It seems that the current implementation does not work in multiple ways :-) :-) A proposed fix is included in this PR. Please review.

emmenlau:
Furthermore I have some issues with finding HDF5 on MSVC (see https://gitlab.kitware.com/cmake/cmake/issues/18916). Therefore I have removed the `QUIET` option from all the `find_package()` calls, and have fixed the required HDF5 `COMPONENTS` to `C`. Please review.

tristan0x:
Thank you very much Mario. I currently lack bandwidth to review your contribution. I hope to have time to do it by the end of this week.
Best

emmenlau:
Dear @tristan0x I've just re-read my changes and found something I overlooked before: You have a special construct in `CMake/portability/CheckCXX11Portability.cmake` to not use `c++11` on a compiler that is identified by `CMAKE_CXX_COMPILER_ID STREQUAL "XL"`.

Is that code that is currently still in use and required? Would HighFive even build without `c++11`? If not, then I would just remove this part. But if yes, then I would adapt my suggested change and put it in an:
```
if(NOT CMAKE_CXX_COMPILER_ID STREQUAL "XL")
    ...
endif()
```

tristan0x:
> Is that code that is currently still in use and required?

This is for IBM XLC compiler. I am afraid it is still supported.

> Would HighFive even build without `c++11`?

Yes, HighFive is compatible with C++98.

emmenlau:
Dear @tristan0x, I've updated the PR again. This update includes the re-added support for XLC compiler, and a new fix for the compiler detection of MSVC in `CMake/config/CompilerFlagsHelpers.cmake`. The previous compiler detection did not work for me on MSVC 2015 and MSVC 2017, and there where minor issues that I resolved.
For example, the `else` the following section seems incorrect and should be replaced by `elseif`, and the check for MSVC is inconsistent with the other compiler checks:
```
	if(CMAKE_${COMPILER_LANGUAGE}_COMPILER_ID STREQUAL "XL")
	  set(CMAKE_${COMPILER_LANGUAGE}_COMPILER_IS_XLC ON)
	elseif(CMAKE_${COMPILER_LANGUAGE}_COMPILER_ID STREQUAL "Intel")
	  set(CMAKE_${COMPILER_LANGUAGE}_COMPILER_IS_ICC ON)
	elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
	  set(CMAKE_${COMPILER_LANGUAGE}_COMPILER_IS_MSVC ON)
	elseif(${CMAKE_${COMPILER_LANGUAGE}_COMPILER_ID} STREQUAL "Clang")
	  set(CMAKE_${COMPILER_LANGUAGE}_COMPILER_IS_CLANG ON)
	else(CMAKE_${COMPILER_LANGUAGE}_COMPILER_ID STREQUAL "GNU")
	  set(CMAKE_${COMPILER_LANGUAGE}_COMPILER_IS_GCC ON)
CMAKE_${COMPILER_LANGUAGE}_COMPILER_ID = '${CMAKE_${COMPILER_LANGUAGE}_COMPILER_ID}'")
	endif()
```

emmenlau:
In the latest PR, all issues should be resolved. Can you please review or proceed to merging?

ferdonline:
@emmenlau Some of these issues were already addressed in another PR.
However some of your fixes are still valid and it would be worth bringing them to the project.

I would prefer no changes to policies and fix the warnings by creating "universal" CMake rules.

ferdonline:
@emmenlau I tried to bring this PR up to date, but I get permission denied when pushing. Would you accept additional commits?

ferdonline:
This is been mostly superseded by #202 
Suggestions by @henryiii are valuable and should be integrated maybe with other CMake changes

PR #177, opened 2019-05-15
CMakeLists.txt: Added c++ version support check for MSVC

emmenlau:
The current c++ standards check in CMakeLists.txt does not work for MSVC. This PR adds an improved check that supports various c++ compilers and already introduces support variables for c++17.


ferdonline:
@emmenlau Thanks for the contribution.
Nevertheless adding all these cases is not very future proof, and cmake3.1 actually has good support for such checks using CMAKE_CXX_STANDARD_REQUIRED and CMAKE_CXX_STANDARD. 
Could you please use them instead?

emmenlau:
Dear @ferdonline, I'm not very experienced with header-only libraries. AFAIK the flags in CMAKE_CXX_STANDARD_REQUIRED and CMAKE_CXX_STANDARD will not be respected if no source files are to be compiled.

I was under the assumption that this is the reason why you currently use the explicit compiler checks instead of using the more standard approach. And this is also why I kept your current logic of explicit compiler checks intact.

PR #169, opened 2019-02-26
Fix MSVC warnings

tklauser:
This fixes all remaining MSVC type conversion warnings (both x64 and x32
build) and one "code unreachable" warning (by marking
HDF5ErrMapper::ToException as [[noreturn]]).
    
Also, in DataSpace::getMaxDimensions it fixes the H5S_UNLIMITED ->
DataSpace::UNLIMITED replacement to work on the correct vector.
    
Fixes #164

PR #179, opened 2019-05-28
Fix another MSVC warnings when building for win32

tklauser:
After PR #169 there is still a warning remaining when building for win32
where sizeof(size_t) > sizeof(unsigned int). Fix it by using brace
initialization which requires that the value inside is exactly
representable in size_t. This is better than the static_cast used in the
initial version of PR #169.


ferdonline:
Thanks @tklauser. There should be at least another similar case in bits/H5Dataspace_misc.hpp. Could you please convert that one too to stay consistent?

tklauser:
The case in `bits/H5Dataspace_misc.hpp` is now also converted, thanks @ferdonline

PR #181, opened 2019-07-03
Fixes for building with Visual Studio CMake integration with MSVC compiler

contre:
Several fixes for using the CMake tooling as part of Visual Studio:
* remove and fix up the C++ standards detection code to make use of CMAKE_CXX_STANDARD flags. This is partially based on PR BlueBrain/HighFive#163 as these changes were necessary to get MSVC working. This also removes the need for HIGHFIVE_CPP_STD_FLAG
* Don't include the tests sub dir if you're not building with the tests on. This is currently handled in the tests cmakelists but I figured it was better to be explicit in the top level
* several unit tests fail due to invalid characters in the filenames generated from typenames. Windows doesn't allow "<>:" in paths so replace them with _. Do the same with spaces to make sure we don't cause any issues with that. Added a helper function to handle this issue with #if check for win32.
* Boost Test does not like std::string being in the dataset_test_types typelist used for the BOOST_AUTO_TEST_CASE_TEMPLATE test case generator. This ends up with tests not being run due to their being spaces in the generated typenames from typeid(T).name(). This occurrs on both gcc and MSVC. Added separate tests for the string versions of these tests instead
* Add ignore for .vs Visual Studio temporary folder


contre:
The CI build failed due to an issue with Travis only on OSX. I've created a separate PR #182 to try and workaround this issue.

PR #184, opened 2019-07-05
Boost requirements depend on tests

### LINKED ISSUE #176: Dependency to boost_unit_test_framework should be active only with HIGHFIVE_UNIT_TESTS=On
Hey,

I wanted to use HighFive via CMake's `ExternalProject_Add` and I have to provide `boost_unit_test_framework` module in order to configure it. It shouldn't be needed. (Note: `enable_testing` should be conditional as well)

As HighFive is header only I guess I could skip configuration step at all, but I want to have HDF5 dependency checked.

I will try to provide PR later.

ferdonline:
Boost testing framework shall not be required if not building tests
If user attepts to disable boost and keep tests witll emmit an early
error
H5Easy test module is now independent

Fixes #176


CMakeLists.txt
```diff
@@ -25,12 +24,16 @@ if(HIGHFIVE_PARALLEL_HDF5)
...
-find_package(HDF5 QUIET REQUIRED)
+find_package(HDF5 REQUIRED)
 
 set(Boost_NO_BOOST_CMAKE TRUE)
 
 if(USE_BOOST)
-  find_package(Boost QUIET REQUIRED COMPONENTS system unit_test_framework)
+  set(BOOST_REQ_COMPONENTS system)
+  if(HIGHFIVE_UNIT_TESTS)
+    set(BOOST_REQ_COMPONENTS ${BOOST_REQ_COMPONENTS} unit_test_framework)
```
tristan0x: What about using `list(APPEND ...)` instead of `set(BOOST_REQ_COMPONENTS ${BOOST_REQ_COMPONENTS} unit_test_framework)`?

ferdonline: There is a strange case when it shows a `;` instead of being treated as 2+ values... Maybe when we use quotes?

ferdonline: I just tried and it seems to work fine. Thanks for the hint, indeed it's neater

PR #185, opened 2019-07-05
Tests: Travis to keep Homebrew caches

ferdonline:


ferdonline:
I don't get the expected performance improvement... :(
I'm tempted to cache the updated brew and don't ask the update

ferdonline:
@tristan0x we are back to 6min OSX builds. :)
Do you agree with this solution?

ferdonline:
@tristan0x I'm a bit concerned with the reliability of this solution. It seems caches are not supposed to be shared, and without the `update:true` we kind of rely on that.
I would maybe prefer #186 but for some strange reason unrelated OS images are failing...

PR #187, opened 2019-07-09
H5Easy support loading vector<vector>

gaehlerm:
I hope this break doesn't break anything, but it seems to be working.

### DIFF
--- a/include/highfive/bits/H5Easy_vector.hpp
+++ b/include/highfive/bits/H5Easy_vector.hpp
@@ -53,6 +53,23 @@ struct overwrite_impl
         return dataset;
     }
 };
+    
+// load "std::vector<std::vector>" from DataSet
+template <class T>
+struct load_impl<std::vector<std::vector<T>>>
+{
+    static std::vector<std::vector<T>> run(const File& file, const std::string& path)
+    {
+        DataSet dataset = file.getDataSet(path);
+        std::vector<size_t> dims = dataset.getDimensions();
+        if (dims.size() != 2) {
+            throw detail::error(file, path, "H5Easy::load: Field not rank 2");
+        }
+        std::vector<std::vector<T>> data;
+        dataset.read(data);
+        return data;
+    }
+};
 
 }  // namespace vector
 


include/highfive/bits/H5Easy_vector.hpp
```diff
@@ -53,6 +53,23 @@ struct overwrite_impl
...
+        if (dims.size() != 2) {
+            throw detail::error(file, path, "H5Easy::load: Field not rank 2");
+        }
+        std::vector<std::vector<T>> data;
+        dataset.read(data);
+        return data;
+    }
+};
 
 }  // namespace vector
```
tdegeus: The reason your code is failing is that you are addition your overload to the `H5Easy::detail::vector` namespace, whereas the overloads are selected from the `H5Easy::detail` namespace. Practically this means that you need to reposition this closing bracket (`}  // namespace vector`) above your added code.

tdegeus: A side note is that I do think it is important to keep this overload before the `std::vector<T>` overload (as you already did).

include/highfive/bits/H5Easy_vector.hpp
```diff
@@ -53,6 +53,23 @@ struct overwrite_impl
...
+    
+// load "std::vector<std::vector>" from DataSet
+template <class T>
+struct load_impl<std::vector<std::vector<T>>>
+{
+    static std::vector<std::vector<T>> run(const File& file, const std::string& path)
+    {
+        DataSet dataset = file.getDataSet(path);
+        std::vector<size_t> dims = dataset.getDimensions();
+        if (dims.size() != 2) {
```
tdegeus: In fact your addition is entirely obsolete. If the dimension check is removed or made more intelligent in the existing overload the code will readily work (`std::vector<std::vector<double>>` is a valid specialisation of `std::vector<T>`).

What surprises me a little however, it that I would expect the same for the dump overload, but that one does seem to require an explicit `std::vector<std::vector<T>>` overload.

tdegeus:
Having such an overload seems to be a relevant extension. However, the approach in my opinion can be a but different. I've made some comments with the code.

include/highfive/bits/H5Easy_vector.hpp
gaehlerm: I'm fine with any of the suggested changes, as long as I can load vector<vector<T>>. Removing the `if (dims.size() != 1) {` seems like the simplest solution to me. But I didn't want to remove it as I thought it might be there for some reason. But possibly the author just didn't consider that T may be a vector itself and everything still works out.

tdegeus: Since I am the author I can confirm ;). I would suggest that `if (dims.size() != 1) {` be replaced by `if (dims.size() != detail::vector::getDims(data)) {`. Whereby:

* The check should be placed after the allocation of `data`,
* `getDims(data)` should be on the lines of [this answer](https://stackoverflow.com/a/28942971/2646505).

tdegeus:
@gaehlerm Are you still interested in making this addition? If yes, it would be great to discuss my propositions.

gaehlerm:
@tdegeus Yes I am. As far as I remember it was about loading vector<vector<T>>. I just need a fix to make it work, preferably with the following syntax:
`auto valueMatrix = H5Easy::load<std::vector<std::vector<double>>>(file, valueName);`
As far as I remember, my fix enabled this. But better solutions are of course welcome. I did not read the code in depth nor am I skilled enough to judge the quality of the fix.
Thanks a lot for your efforts!

tdegeus:
@gaehlerm See #204 .

My initial assessment was not entirely correct, in the sense that there was no missing overload for writing, it was simply a matter of extracting the inner-most typename. For reading 1-d and 2-d vectors now work, but there is still an issue with n-d vectors with n >= 3.

ferdonline:
@tdegeus I find it preferable to work on the same PR until is it ready to avoid fragmenting the reviews and discredit the original work. Even if the original author is not responding, in most cases you can still commit to his branch, which seems to be the case here. Opening a new PR should be an absolute last resort.
@gaehlerm Since the original fix seems to be very different, do you agree we move to the new PR, making sure you and this PR are appropriately credited?
Cheers

gaehlerm:
@ferdonline I'm fine with whatever you do, as long as my code mentioned above works.

ferdonline:
Thanks @gaehlerm. Then could you please verify if the tests added in #204 cover your use case?

tdegeus:
OK @ferdonline , next time I will amend existing PRs, even if they change drastically

tdegeus:
@gaehlerm The "``H5Easy_vector2d``" test should be your use-case if I understood well.

gaehlerm:
> @gaehlerm The "`H5Easy_vector2d`" test should be your use-case if I understood well.

@tdegeus exactly

ferdonline:
Thank you guys

tdegeus:
@gaehlerm Please check-out the current master: you should be good to go.

gaehlerm:
looks great, thanks a lot guys!

Issue #96, opened 2018-01-29
Dataset read error with 3D vector


Mrporkk:
When I compile this code I get error with this,but if I notes "dataset1.read(data);" this code,it compiles successful, and show label.size() is right. How to solve this problem?

# SOURCE CODE
std::vector<std::vector<std::vector<float>>> data;
std::vector<int> label;
std::string filename = "/path/to/data/h5_data/110020171219_h5/0.h5";
File file(filename, File::ReadOnly);
DataSet dataset1 = file.getDataSet("data");
dataset1.read(data);
DataSet dataset2 = file.getDataSet("label");
dataset2.read(label);
std::cout << label.size() << std::endl;

# ERROR
In file included from /path/to/CLionProjects/Create_hdf5/src/highfive/bits/../bits/H5Attribute_misc.hpp:29:0,
                 from /path/to/CLionProjects/Create_hdf5/src/highfive/bits/../H5Attribute.hpp:73,
                 from /path/to/CLionProjects/Create_hdf5/src/highfive/bits/H5Annotate_traits_misc.hpp:18,
                 from /path/to/CLionProjects/Create_hdf5/src/highfive/bits/H5Annotate_traits.hpp:74,
                 from /path/to/CLionProjects/Create_hdf5/src/highfive/H5File.hpp:17,
                 from/path/to/CLionProjects/Create_hdf5/src/main.cpp:7:
/path/to/CLionProjects/Create_hdf5/src/highfive/bits/../bits/H5Converter_misc.hpp: In instantiation of ‘typename std::vector<_RealType>::iterator HighFive::details::single_buffer_to_vectors(typename std::vector<_RealType>::iterator, typename std::vector<_RealType>::iterator, const std::vector<long unsigned int>&, std::size_t, std::vector<WeightType>&) [with T = float; U = std::vector<std::vector<float> >; typename std::vector<_RealType>::iterator = __gnu_cxx::__normal_iterator<float*, std::vector<float> >; std::size_t = long unsigned int]’:
[...]

PR #191, opened 2019-08-04
Fix reading N dimension vector bug

guoxiaoyu543:
Recently I have been using the HighFive interface to read a HDF5 format file with a 3D dataset, and the code is very simple:
```
int main(int argc_temp, char *argv_temp[])
{
    HighFive::File test_h5("test.h5", HighFive::File::Read);

    vector<vector<vector<double> > > test_vector;
    HighFive::DataSet test_dataset = test_h5.getDataSet("TestData3D");

    test_dataset.read(test_vector);

    return 0;
}
```

However, the compilation failed with the following message:
```
PAYH_TO_TEST_DIR/src/lib/HighFive/include/highfive/bits/../bits/H5Converter_misc.hpp:119:48: error: no matching function for call to ‘error: no matching function for call to ‘single_buffer_to_vectors(std::vector<double>::iterator&, std::vector<double>::iterator&, const std::vector<long unsigned int>&, long unsigned int, std::vector<std::vector<double> >&)’
         begin_buffer = single_buffer_to_vectors(begin_buffer, end_buffer, dims,
                                        ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
                                                 current_dim + 1, *it);
                                                 ~~~~~~~~~~~~~~~~~
```

So firstly I tried reading a 2D dataset
```
int main(int argc_temp, char *argv_temp[])
{
    HighFive::File test_h5("test.h5", HighFive::File::Read);

    vector<vector<double> > test_vector;
    HighFive::DataSet test_dataset = test_h5.getDataSet("TestData2D");

    test_dataset.read(test_vector);

    return 0;
}
```
And the compilation succeeded.

 I have also tried 4D dataset and gotten compilation error. I tried changing the compiler from g++ 7.4.0 to intel 17.0.5 and the compilation didn't succeed, either.

I began to search if anyone had the same problem with me and I found #96 . It seems that no solutions have been provided. Therefore, I thought maybe there was someting wrong with the 3D+ dataset reading (using vector) capability for the HighFive interface.

So I started to debug and it seems that the deduction for the template parameter 'T' fails for the function `single_buffer_to_vectors`, and I changed the code. 

After the modification, everything works fine for me(I tested the code using 1D, 2D, 3D and 4D vector.)


tristan0x:
Thank you for your contribution @guoxiaoyu543 
Could you please rebase your branch please?

ferdonline:
Thanks @guoxiaoyu543!
Besides rebasing, please add the 3D and 4D tests to the suite

guoxiaoyu543:
> @guoxiaoyu543
> I have worked on the tests suite, so it would be nice to have access to your branch. Otherwise we create a new PR

Thank you, and I have sent you an invitation link. Unfortuantely I only have time working on this code during weekends, so it's really nice of you to help with it.

ferdonline:
Thanks @guoxiaoyu543 for the access. To cleanup all "merge master" and since we didn't reviewed code, I force-pushed, but I'm not a fan of doing it

CMake/portability/CheckCXXStandardSupport.cmake
```diff
@@ -22,3 +22,14 @@ endif()
 check_cxx_compiler_flag("${CMAKE_CXX11_STANDARD_COMPILE_OPTION}" COMPILER_SUPPORTS_CXX11)
 check_cxx_compiler_flag("${CMAKE_CXX14_STANDARD_COMPILE_OPTION}" COMPILER_SUPPORTS_CXX14)
 check_cxx_compiler_flag("${CMAKE_CXX17_STANDARD_COMPILE_OPTION}" COMPILER_SUPPORTS_CXX17)
+
+function(select_recent_cxx_standard_available)
```
tristan0x: IMHO I find too intrusive to modify the parent environment. The `message(FATAL_ERROR ...)` in top-level CMakeLists.txt is enough.

tests/unit/CMakeLists.txt
```diff
@@ -35,7 +35,7 @@ if(HIGHFIVE_PARALLEL_HDF5)
   target_link_libraries(tests_parallel_bin
     ${Boost_UNIT_TEST_FRAMEWORK_LIBRARIES} ${HDF5_C_LIBRARIES}
     ${MPI_C_LIBRARIES})
-  add_test(NAME tests_parallel COMMAND ${TEST_MPI_EXEC_PREFIX_DEFAULT} -n 8
+  add_test(NAME tests_parallel COMMAND ${TEST_MPI_EXEC_PREFIX_DEFAULT}
```
tristan0x: without `-n 8`, is the test still executed in // ?

include/highfive/bits/H5Converter_misc.hpp
```diff
@@ -96,28 +94,25 @@ single_buffer_to_vectors(typename std::vector<T>::iterator begin_buffer,
                          const std::vector<size_t>& dims,
                          const size_t current_dim,
                          std::vector<T>& vec_single_dim) {
-    const size_t n_elems = dims[current_dim];
-    typename std::vector<T>::iterator end_copy_iter =
-        std::min(begin_buffer + static_cast<long>(n_elems), end_buffer);
+    const long n_elems = static_cast<long>(dims[current_dim]);
```
tristan0x: Remove duplicate type information:

```suggestion
    const auto n_elems = static_cast<long>(dims[current_dim]);
```

CMake/portability/CheckCXXStandardSupport.cmake
ferdonline: It only sets when user requests it with `select_recent_cxx_standard_available` and only if CMAKE_CXX_STANDARD is not set yet. Parent scope is the caller scope so the user gets what he asked, no side-effects AFAIK. Do you see other problems? suggestions?

tests/unit/CMakeLists.txt
ferdonline: srun and mpiexec take by default the size of the allocation. If you prefer to hardcode the value then I'd suggest 2, since 8 cores is still not very common

CMake/portability/CheckCXXStandardSupport.cmake
tristan0x: I am fine with that then.

tests/unit/CMakeLists.txt
tristan0x: I am good

PR #198, opened 2019-09-23
Added missing inline to H5Easy (fixes #194)

### LINKED ISSUE #194: Multiple Definition in H5Easy
Hey there,

I just found a problem in the H5Easy Eigen extension: A linker error occurs in g++ 9.1.0 when including H5Easy.hpp from multiple modlules:

>multiple definition of `H5Easy::detail::eigen::shape(HighFive::File const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, HighFive::DataSet const&, int)';

Following the error leads to **include/highfive/bits/H5Easy_Eigen.hpp:41**. For resolving the issue without introducing source files I propose inlining the aforementioned function:

`
**inline** std::vector<Eigen::Index> shape(const File& file,
                                const std::string& path,
                                const DataSet& dataset,
                                int RowsAtCompileTime)
`

tdegeus:


ferdonline:
Thanks @tdegeus

PR #199, opened 2019-09-23
Making optional Boost dependency non-default (fixes #196)

### LINKED ISSUE #196: cmake fails without boost
Is boost optional or no?

tdegeus:
Making optional Boost dependency non-default in CMake test/installation.

Fixes #196

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@ if(NOT COMPILER_SUPPORTS_CXX11)
 endif()
 
 option(HIGHFIVE_UNIT_TESTS "enable unit tests" TRUE)
-option(USE_BOOST "enable Boost Support" TRUE)
+option(USE_BOOST "enable Boost Support" FALSE)
 option(USE_EIGEN "enable Eigen testing" FALSE)
 option(USE_XTENSOR "enable xtensor testing" FALSE)
 option(HIGHFIVE_EXAMPLES "Compile examples" TRUE)
--- a/.travis.yml
+++ b/.travis.yml
@@ -130,6 +130,7 @@ script:
   - mkdir -p build && pushd build
   - >
     cmake --warn-uninitialized
+    -DUSE_BOOST=TRUE
     -DUSE_EIGEN:BOOL=ON
     -DEigen3_DIR=/usr/lib/cmake/eigen3
     -DUSE_XTENSOR:BOOL=${USE_XTENSOR}
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ HighFive does not require an additional library and supports both HDF5 thread sa
 
 
 ### Design
+
 - Simple C++-ish minimalist interface
 - No other dependency than libhdf5
 - Zero overhead
@@ -20,9 +21,9 @@ HighFive does not require an additional library and supports both HDF5 thread sa
 
 
 ### Dependencies
-- libhdf5
-- (optional) boost >= 1.41
 
+- libhdf5
+- boost >= 1.41 (optional, enable CMake option `USE_BOOST`)
 
 ### CMake integration
 
@@ -162,6 +163,7 @@ make test
 
 
 ### Contributors
+
 - Adrien Devresse <adrien.devresse@epfl.ch> - Blue Brain Project
 - Ali Can Demiralp <demiralpali@gmail.com> -
 - Fernando Pereira <fernando.pereira@epfl.ch> - Blue Brain Project
@@ -171,4 +173,5 @@ make test
 - Tom de Geus <tom@geus.me> - EPFL
 
 ### License
+
 Boost Software License 1.0


CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ if(NOT COMPILER_SUPPORTS_CXX11)
 endif()
 
 option(HIGHFIVE_UNIT_TESTS "enable unit tests" TRUE)
-option(USE_BOOST "enable Boost Support" TRUE)
+option(USE_BOOST "enable Boost Support" FALSE)
```
contre: Having USE_BOOT false and HIGHFIVE_UNIT_TESTS true does not make sense as to have the unit tests you need boost. Should that also be changed?

ferdonline:
Indeed, having an inconsistent set of defaults is not good.
Since tests are important for development, and when used in reality this is typically a submodule, then it makes sense to me we set both tests and boost to FALSE in case this is the top-level project.

tdegeus:
@ferdonline I agree with your comment. In addition I think that users who are simply installing the library should also not be faced with the seeming need to install Boost. So I would argue to set both `USE_BOOST` and `HIGHFIVE_UNIT_TESTS` off by default. Those who want to test can simply enable the options, and in the CI, naturally, they will be enabled. 

The only thing is that maybe it make sense to switch `USE_BOOST` to `TRUE` when `HIGHFIVE_UNIT_TESTS=TRUE`.

ferdonline:
My comment was in the direction of conditionally disabling both :)

tdegeus:
@ferdonline Does that mean that the current version is in line with your ideas?

ferdonline:
@tdegeus I believe it would be a much better default to keep them active when developing, i.e. when it is the main cmake project. When Highfive is incorporated in another project (since it's header only) then indeed tests and boost can be disabled by default. As mentioned, a way I see of checking it is with CMAKE_PROJECT_NAME.

tdegeus:
@ferdonline I understand your point. However, when one would like to simply install the library it would be great to be able to:

```cpp
mkdir build
cd build 
cmake ..
make install
```

without extra, project specific, options. I believe the unneeded Boost dependency can be confusing to new users (see #196) and can thus potentially render the library less attractive.

ferdonline:
I gave a little thought about this. The problem is for new users who don't know cmake well to look for configuration entries. If we just disable boost, then the same problem arises the other way around, they won't know how to enable it. As happened with #196 it is simple to explain the reason and solution.
As so, I believe it is best to warn the user of what's going on and default to enabled to encourage tests and boost. 

Resuming, to my mind. USE_BOOST should default to True. If it fails to find it, print a nice Error , like "Boost not found which is required for efficient multi-dimension and tests.... To disable support please use `cmake.. -DUSE_BOOST=OFF`.
Building of tests: Given the user was warned,  `-DUSE_BOOST=OFF` can automatically imply `HIGHFIVE_UNIT_TESTS=OFF`. Additionally, I believe these tests shall be always disabled when HighFive is a subproject.

tdegeus:
@ferdonline As you wish. I will update the PR soon.

tdegeus:
Sure. I could have it done within 10 days (though hopefully earlier).

tdegeus:
@ferdonline I updated the PR.

Or at leasts... I pushed changes to the branch of my fork. But I think that GitHub is malfunctioning as the PR was not updated. Maybe we wait a bit to see if things get updated.

tdegeus:
@ferdonline @alkino 

Note that the PR contains an urgent bug-fix for a missing include in `bits/H5Dataspace_misc.hpp`.I urge you to either treat this PR with priority. Otherwise I guess I should open a second PR.

tdegeus:
@alkino I don't think one should have to do that for basic use. I would strongly argue to keep all traces from the CMakeLists

CMakeLists.txt
```diff
@@ -20,88 +19,90 @@ include(CheckCXXStandardSupport)
...
-option(USE_EIGEN "enable Eigen testing" FALSE)
-option(USE_XTENSOR "enable xtensor testing" FALSE)
+option(HIGHFIVE_UNIT_TESTS "Enable unit tests" TRUE)
+option(USE_BOOST "Enable Boost Support" TRUE)
+option(USE_EIGEN "Enable Eigen testing" FALSE)
+option(USE_XTENSOR "Enable xtensor testing" FALSE)
 option(HIGHFIVE_EXAMPLES "Compile examples" TRUE)
 option(HIGHFIVE_PARALLEL_HDF5 "Enable Parallel HDF5 support" FALSE)
 
+if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
```
ferdonline: -> `if (NOT...`
Btw, can't we put the next if here with an `OR`?

tdegeus: Done

CMakeLists.txt
```diff
@@ -20,88 +19,87 @@ include(CheckCXXStandardSupport)
...
-option(USE_XTENSOR "enable xtensor testing" FALSE)
+option(HIGHFIVE_UNIT_TESTS "Enable unit tests" TRUE)
+option(USE_BOOST "Enable Boost Support" TRUE)
+option(USE_EIGEN "Enable Eigen testing" FALSE)
+option(USE_XTENSOR "Enable xtensor testing" FALSE)
 option(HIGHFIVE_EXAMPLES "Compile examples" TRUE)
 option(HIGHFIVE_PARALLEL_HDF5 "Enable Parallel HDF5 support" FALSE)
 
+# Disable test if Boost will not be loaded, or if HighFive is a sub-project
+if (NOT USE_BOOST OR CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
```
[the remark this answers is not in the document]
tdegeus: Sorry 😞

tdegeus:
@ferdonline `include/highfive/bits/H5Dataspace_misc.hpp` was missing `#include <numeric>` to do `std::accumulate(...)`. Compilers on different platform were throwing on this while compiling the examples.

alkino:
@tdegeus can you merge this, or do we need to do it?

tdegeus:
@alkino I cannot merge.

PR #204, opened 2019-09-27
H5Easy: Support for n-d vector

tdegeus:
H5Easy read/write support for n-d vector. 

Supersedes #187 . 

Reading of n-d vector should work after merging #191 (relevant test should be uncommented then).


tdegeus:
@adevress Are there existing functions for the following support functions that I introduced: `H5Easy::detail::vector::checkShape`, `H5Easy::detail::vector::getShape`, `H5Easy::detail::vector::fillShape` ?

tdegeus:
@adevress Never mind, I simply tried... By removing also this the implementation is I think much more clean than before, and more generic too.

ferdonline:
@tdegeus could we uncomment the tests?

tdegeus:
@ferdonline Done! Looks like everything is good to go.

PR #220, opened 2019-10-14
Compile doxygen documentation with cmake

alkino:
Here is the first version of doxygen documentation.
This doc need to be better because it is not so nice.
You can see the output at: http://bluebrain.github.io/HighFive/


doc/CMakeLists.txt
```diff
@@ -16,4 +16,12 @@ install(
...
 )
-endif(HIGH_FIVE_DOCUMENTATION)
\ No newline at end of file
+
+find_package(Doxygen)
+if(Doxygen_FOUND)
+    configure_file(${CMAKE_CURRENT_SOURCE_DIR}/Doxyfile ${CMAKE_CURRENT_BINARY_DIR}/Doxyfile @ONLY)
+    add_custom_target(doxygen ALL COMMAND Doxygen::doxygen ${CMAKE_CURRENT_BINARY_DIR}/Doxyfile)
+else()
+    message(STATUS "Doxygen documentation will not be build because Doxygen has not been found.")
```
tristan0x:
```suggestion
    message(WARNING "Doxygen documentation will not be built because Doxygen has not been found.")
```

tristan0x: Plus I am not a big fan of building the documentation in target `all` in case it fails.
I would either add a CMake option to enable/disable Doxygen build or remove it from `all` target (I prefer the second solution).

alkino: To have a documentation that build is a good thing imho, I like that travis said no if you break it.

tristan0x: > To have a documentation that build is a good thing imho, I like that travis said no if you break it.

I agree. I suggest to execute `make doxygen` in Travis

tristan0x:
> This doc need to be better because it is not so nice.

Could you please be more explicit about the improvements you have in mind?

alkino:
> > This doc need to be better because it is not so nice.
> 
> Could you please be more explicit about the improvements you have in mind?

We should add an index which explain H5Easy and some examples.
We should rephrase some of the brief.

tristan0x:
> > > This doc need to be better because it is not so nice.
> > 
> > 
> > Could you please be more explicit about the improvements you have in mind?
> 
> We should add an index which explain H5Easy and some examples.
> We should rephrase some of the brief.
> And put more links between helps.
> 
> But not in this commit.

Good idea. Can you create a ticket about that?

doc/CMakeLists.txt
alkino: For travis the patch is easy but let's wait that the windows travis changes are merged.

doc/CMakeLists.txt
```diff
@@ -1,19 +1,7 @@
-if(HIGH_FIVE_DOCUMENTATION)
```
tristan0x: Get also rid of the .sh and .css files in the `doc/` sub-directory

doc/CMakeLists.txt
```diff
@@ -1,27 +1,7 @@
...
-)
-install(
-	DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/src
-    DESTINATION ${HTML_INSTALL_DIR}
-)
-
 find_package(Doxygen)
 if(Doxygen_FOUND)
     configure_file(${CMAKE_CURRENT_SOURCE_DIR}/Doxyfile ${CMAKE_CURRENT_BINARY_DIR}/Doxyfile @ONLY)
-    add_custom_target(doxygen ALL COMMAND Doxygen::doxygen ${CMAKE_CURRENT_BINARY_DIR}/Doxyfile)
```
alkino: i should name the target **doc** instead of **doxygen**, no?

**make doc** vs **make doxygen**

tristan0x: I would go for `html`.
See https://www.gnu.org/prep/standards/html_node/Standard-Targets.html

alkino: I never see this convention respected anywhere.

tristan0x: > I never see this convention respected anywhere.

I don't understand your point?

tristan0x: In the last meeting about coding conventions in BBP (I was in vacations), it was suggested to have a make target called `doc`:
See https://bbpteam.epfl.ch/project/spaces/display/QA/Workshop+Coding+Guidelines

alkino: I mean that **make html** sounds weird for me, and I can't remember a project using that to build the doc.

alkino: I mean that **make html** sounds weird for me, and I can't remember a project using that to build the doc.

tristan0x: The fact that there are projects that don't follow the convention doesn't make it less valid.
I can name a few that follow it: sphinx Makefiles and all GNU projects.

Like I said above, the target may be named `doc` to meet BBP convention.

alkino:
It break because of a bug of conda since yesterday...
https://github.com/conda/conda/issues/9337

doc/CMakeLists.txt
alkino: Ok I put doc.

alkino:
Can we merge this one once the travis is good?

doc/CMakeLists.txt
alkino: The `make doc` should be done in a new patchset

doc/CMakeLists.txt
```diff
@@ -1,19 +1,7 @@
...
-    DESTINATION ${HTML_INSTALL_DIR}
-)
-install(
-	DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/src
-    DESTINATION ${HTML_INSTALL_DIR}
-)
-endif(HIGH_FIVE_DOCUMENTATION)
\ No newline at end of file
+find_package(Doxygen)
+if(Doxygen_FOUND)
```
tdegeus: I would prefer having an option to compile the doxygen docs. If I just want to install the library on several machines I want to have a quick option to do so.

PR #221, opened 2019-10-14
Getting type info and object kind

### LINKED ISSUE #178: Get Type of dataset
I have a file containing many vector<double> and vector<uint64_t> which I would like to read and store inside map<string, vector<double>> or map<string, vector<uint64_t>>. So the code I'd like to have is something like this:
// iterate over all the datasets in a group
std::string type = H5::getType(file, dataSetName);
if(type == "double"){
  myMap[dataSetName] = H5Easy::load(file, dataSetName);
} else if { // repeat for uint64_t and all other types }

The type of a dataset is stored inside the HDF5 file, but is there also such a function getType(...) available? Otherwise I can also add an attribute type = "double", ... to every dataset, but this information seems redundant.

### LINKED ISSUE #209: How to know is the data path exists?
I would like to know if is possible to check/verify if some data exists in some HDF5 path before try to read it using HighFive.

Pseudo-code:

```
path = path/to/data

bool status = HDF5_path_exists(path);

if(status == true) {
    // High Five :: read(path)
}
```

**Is this possible?**

ferdonline:
This PR implements some inspection mechanisms, namely for 
- the type of objects in a "Node"
- The data type of a 

### Type of objects
hdf5, since version 1.8 deprecates treating link types and object types together and makes a clear distinction of the two concepts. Therefore a "name" can be described in two ways:
- What kind of link it is (hard, soft, external)
- What kind of object it addresses (a group, dataset, user type...)

As so, two API functions were created for Nodes:
 - `getLinkType(string)`
 - `getObjectType(string)`

**Example**
```
BOOST_CHECK(file.getLinkType("group1") == LinkType::Hard);
BOOST_CHECK(file.getObjectType("group1") == ObjectType::Group);
```
As aux private API, a `getType()` was added to Object to help obtaining the object type from any hdf5 object addressed with an hid

### Data Type
To obtain data types, only a few functions were added to `H5DataType`, namely:
- `DataTypeClass getClass();` // fundamental type
- `size_t getSize() const;`  // byte size
- `std::string string() const;`  // string representation

**Example**
```
auto dt = g.getDataSet("ds").getDataType();
BOOST_CHECK(dt.getClass() == DataTypeClass::Float);
BOOST_CHECK(dt.getSize() == 8);
BOOST_CHECK(dt.string() == "Float64");
```
Fixes #178 
Fixes #209

### DIFF
--- a/include/highfive/H5Object.hpp
+++ b/include/highfive/H5Object.hpp
@@ -10,6 +10,8 @@
 #define H5OBJECT_HPP
 
 #include <H5Ipublic.h>
+#include <H5Opublic.h>
+#include <ctime>
 
 namespace HighFive {
 
@@ -19,6 +21,18 @@ class NodeTraits;
 
 template <typename Derivate>
 class AnnotateTraits;
+
+class ObjectInfo;
+
+enum class ObjectType {
+    Invalid = -1,
+    File,
+    Group,
+    UserDataType,
+    DataSpace,
+    Dataset,
+    Attribute
+};
 
 
 class Object {
@@ -39,6 +53,17 @@ class Object {
     ///
     hid_t getId() const;
 
+    ///
+    /// \brief getInfo
+    /// \return Obtains several infos about the object
+    ///
+    ObjectInfo getInfo() const;
+
+    ///
+    /// \brief Gets the fundamental type of the object (dataset, group,...)
+    ///
+    ObjectType getType() const;
+
   protected:
     // empty constructor
     Object();
@@ -51,12 +76,33 @@ class Object {
     hid_t _hid;
 
   private:
+    // Init with an low-level object id
+    Object(hid_t);
+
     template <typename Derivate>
     friend class NodeTraits;
     template <typename Derivate>
     friend class AnnotateTraits;
 };
-}
+
+
+class ObjectInfo  {
+  public:
+    haddr_t getAddress() const noexcept;
+
+    size_t referenceCount() const noexcept;
+
+    std::time_t creationTime() const noexcept;
+
+    std::time_t modificationTime() const noexcept;
+
+  protected:
+    H5O_info_t raw_info;
+
+    friend class Object;
+};
+
+}  // namespace HighFive
 
 #include "bits/H5Object_misc.hpp"
 
--- a/include/highfive/bits/H5Node_traits.hpp
+++ b/include/highfive/bits/H5Node_traits.hpp
@@ -19,6 +19,9 @@ class Group;
 class DataSpace;
 class DataType;
 
+enum class LinkType;
+
+
 template <typename Derivate>
 class NodeTraits {
   public:
@@ -114,12 +117,31 @@ class NodeTraits {
     /// false
     bool exist(const std::string& node_name) const;
 
+    /// \brief Returns the kind of link of the given name (soft, hard...)
+    LinkType getLinkType(const std::string& node_name) const;
+
+    /// \brief A shorthand to get the kind of object pointed to (group, dataset, type...)
+    inline ObjectType getObjectType(const std::string& node_name) const;
+
+
   private:
     typedef Derivate derivate_type;
 
     // A wrapper over the low-level H5Lexist
     bool _exist(const std::string& node_name) const;
+
+    // Opens an arbitrary object to obtain info
+    Object _open(const std::string& node_name,
+                 const DataSetAccessProps& accessProps=DataSetAccessProps()) const;
 };
+
+
+enum class LinkType {
+    Hard,
+    Soft,
+    External
+};
+
 
 }  // namespace HighFive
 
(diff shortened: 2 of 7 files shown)


include/highfive/bits/H5Node_traits.hpp
```diff
@@ -114,13 +117,32 @@ class NodeTraits {
     /// false
     bool exist(const std::string& node_name) const;
 
+    /// \brief Returns the kind of link of the given name (soft, hard...)
```
alkino: Document \param please

include/highfive/bits/H5Node_traits.hpp
```diff
@@ -114,13 +117,32 @@ class NodeTraits {
...
+    LinkType getLinkType(const std::string& node_name) const;
+
+    /// \brief A shorthand to get the kind of object pointed to (group, dataset, type...)
```
alkino: idem

include/highfive/bits/H5Node_traits.hpp
```diff
@@ -114,13 +117,32 @@ class NodeTraits {
...
+    inline ObjectType getObjectType(const std::string& node_name) const;
+
+
   private:
     typedef Derivate derivate_type;
 
     // A wrapper over the low-level H5Lexist
     bool _exist(const std::string& node_name) const;
+
+    // Opens an arbitrary object to obtain info
```
alkino: idem

ferdonline: Here it was not documentation, just a developer comment (dual-slash)

include/highfive/H5Object.hpp
```diff
@@ -51,12 +79,40 @@ class Object {
     hid_t _hid;
 
   private:
+    // Init with an low-level object id
+    Object(hid_t);
```
tristan0x:
```suggestion
    explicit Object(hid_t);
```
To avoid explicit conversion

include/highfive/H5Object.hpp
```diff
@@ -51,12 +79,40 @@ class Object {
...
-}
+
+
+///
+/// \brief A class for accessing hdf5 objects info
+///
+class ObjectInfo  {
+  public:
+    /// \brief Retrieve the address of the object (within its file)
+    haddr_t getAddress() const noexcept;
```
tristan0x: `getAddress` is not consistent with the other member variables that are not prefixed by `get`.
The naming convention in HighFive is in favor of prefixing accessor with `get`, so I would rename the other member functions instead.

include/highfive/bits/H5Node_traits.hpp
```diff
@@ -110,17 +113,41 @@ class NodeTraits {
     ///
     /// \brief check a dataset or group exists in the current node / group
     /// \param dataset/group name to check
-    /// \return true if a dataset/group with the associated name exist, or
-    /// false
+    /// \return true if a dataset/group with the associated name exist, or false
```
tristan0x:
```suggestion
    /// \return true if a dataset/group with the associated name exists, or false
```

include/highfive/bits/H5Node_traits_misc.hpp
```diff
@@ -198,6 +201,55 @@ inline bool NodeTraits<Derivate>::exist(const std::string& group_path) const {
...
+static inline LinkType _convert_link_type(const H5L_type_t& ltype) {
+    switch (ltype) {
+        case H5L_TYPE_HARD:
+            return LinkType::Hard;
+        case H5L_TYPE_SOFT:
+            return LinkType::Hard;
+        case H5L_TYPE_EXTERNAL:
+            return LinkType::Hard;
+        default:
+            throw std::invalid_argument("Invalid Link type ");
```
tristan0x: Error handling in this PR seems inconsistent: sometimes the `default` cases of the `switch` statements return _invalid value_, sometimes it throws.
It is my understanding that HighFive relies on exception for error handling, so should this PR.

ferdonline: This is an internal function that eventually doesn't even make sense to throw.. we dont want to use exceptions for flow control, right? On the other hand I also don't like enums with the "Invalid" entry.
What's your suggestion @tristan0x ?

include/highfive/H5Object.hpp
ferdonline: Yes to consistency! thanks!

include/highfive/bits/H5Node_traits_misc.hpp
tristan0x: If a function is not supposed to handle all cases, do not specify a `default` case, or put an `assert(true)` to be explicit about the lack of code.

Modern compilers will emit a warning if you forget to update a `switch(my_enum)` statement after adding a value to the enum.

tdegeus:
Nice! Thanks for this useful addition!

PR #183, opened 2019-07-05
Travis: add windows build

tristan0x:
Fixes #146

PR #223, opened 2019-10-16
Workaround a conda issue

alkino:


ferdonline:
Can we use this way instead? Looks way cleaner
https://github.com/conda/conda-prefix-replacement/blob/master/.travis.yml#L9-L14

ferdonline:
Please make these changes in #183 It started to fail there so it's just one PR to fix all

PR #225, opened 2019-10-16
Rewrite headers so that all public headers can be included

### LINKED ISSUE #103: Compilation error when only including H5Group.hpp
I am using the master branch of HighFive and compiling with gcc 5.4.0.

I create a file main.cpp with these contents:
```C++
// Compiler error:
#include <highfive/H5Group.hpp>

// Works:
//#include <highfive/H5File.hpp>

int main()
{
	HighFive::Group group;
}
```

I then compile with this command:
```Shell
g++ -std=c++11 -o program -I/home/jkarlsso/Work/HighFive/include -I/usr/include/hdf5/serial main.cpp -lhdf5_serial
```

I then get the following error:
```Shell
In file included from /home/jkarlsso/Work/HighFive/include/highfive/bits/H5Node_traits.hpp:99:0,
                 from /home/jkarlsso/Work/HighFive/include/highfive/H5Group.hpp:14,
                 from main.cpp:2:
/home/jkarlsso/Work/HighFive/include/highfive/bits/H5Node_traits_misc.hpp: In member function ‘HighFive::Group HighFive::NodeTraits<Derivate>::createGroup(const string&)’:
/home/jkarlsso/Work/HighFive/include/highfive/bits/H5Node_traits_misc.hpp:72:77: error: return type ‘class HighFive::Group’ is incomplete
 inline Group NodeTraits<Derivate>::createGroup(const std::string& group_name) {
                                                                             ^
/home/jkarlsso/Work/HighFive/include/highfive/bits/H5Node_traits_misc.hpp:74:15: error: invalid use of incomplete type ‘class HighFive::Group’
     if ((group._hid = H5Gcreate2(static_cast<Derivate*>(this)->getId(),
               ^
In file included from /home/jkarlsso/Work/HighFive/include/highfive/H5Group.hpp:13:0,
                 from main.cpp:2:
/home/jkarlsso/Work/HighFive/include/highfive/bits/H5Annotate_traits.hpp:18:7: note: forward declaration of ‘class HighFive::Group’
 class Group;
       ^
In file included from /home/jkarlsso/Work/HighFive/include/highfive/bits/H5Node_traits.hpp:99:0,
                 from /home/jkarlsso/Work/HighFive/include/highfive/H5Group.hpp:14,
                 from main.cpp:2:
/home/jkarlsso/Work/HighFive/include/highfive/bits/H5Node_traits_misc.hpp: In member function ‘HighFive::Group HighFive::NodeTraits<Derivate>::getGroup(const string&) const’:
/home/jkarlsso/Work/HighFive/include/highfive/bits/H5Node_traits_misc.hpp:85:63: error: return type ‘class HighFive::Group’ is incomplete
 NodeTraits<Derivate>::getGroup(const std::string& group_name) const {
                                                               ^
/home/jkarlsso/Work/HighFive/include/highfive/bits/H5Node_traits_misc.hpp:87:15: error: invalid use of incomplete type ‘class HighFive::Group’
     if ((group._hid = H5Gopen2(static_cast<const Derivate*>(this)->getId(),
               ^
In file included from /home/jkarlsso/Work/HighFive/include/highfive/H5Group.hpp:13:0,
                 from main.cpp:2:
/home/jkarlsso/Work/HighFive/include/highfive/bits/H5Annotate_traits.hpp:18:7: note: forward declaration of ‘class HighFive::Group’
 class Group;
       ^
```

If I instead include "H5File.hpp" it compiles fine.

alkino:
Currently some header files cannot be included alone - See #103 

The first approach is to simply move some of the files which are not supposed to be directly imported to "bits" - Nicolas worked it out

- H5Group.hpp and H5Selection.hpp are move to bits and so should never be included by user again.
- User should user factories to instance them. Btw H5Selection ctor were private.
- *_misc.hpp are include by their parents and should never be include
directly so remove parents include in all of them.

However this approach has the drawback of requiring a lot of files to go into misc and somehow defeating the purpose of the splitting between declaration header and implementation.

Therefore a second approach is proposed, where all includes were redone with a methodology that avoids include loops. The principles are:
 
- Trait headers shall not include their implementation counterpart. Since the implementation depends on some classes defined in the including headers that would create loops
- Select a main client for each trait class, which is the only header including the trait implementation
- Except for the main client, every file requiring a trait shall include the trait declaration header and at the end, the trait main client header.

Tests were added (Thanks Nicolas) to be sure that we can include any of the headers alone.

Fixes #103


ferdonline:
Is this to fix #103 ?

alkino:
Yes and beyond

ferdonline:
I don't know if I agree with your approach. Even though the user shall not instantiate them directly they are still public API and the user might want to "look" at it.
Isn't it enough to add a few `#includes` to these headers?

alkino:
Maybe Attribute should be move to bits because it should be create by a factory too.

ferdonline:
Another possibility, couldn't the trait objects be completely independent if we template them?
For instance, maybe we could pass an extra template to Node to specify the Group class

alkino:
I don't think this is more clean to do it like you suggest.

tristan0x:
> Another possibility, couldn't the trait objects be completely independent if we template them?
> For instance, maybe we could pass an extra template to Node to specify the Group class

It is also tempting to use C++ templates to decouple dependent components, but it usually complicates the API for a wrong reason and compilation issues become less readable.

ferdonline:
My point was that it is nicer to have public classes defined in the main directory. I would like to avoid the case that with more dependencies soon we have only a couple of *.hpp the main dir.
How about keeping the files in the main dir and prefix them with an underscore? For me it would seem evident they shouldn't be imported.

alkino:
The question is what is a public class? What about Node_Traits?
File inherit of it and a lot of public functions are in the bits directory (createGroup, createDataset, exist, ...).

ferdonline:
> What about Node_Traits?

It is a trait, the user will never have a direct instance of that class.

I found logical the way @adevress structured in the first place, and moving three of these files to bits defeats to my mind a bit the purpose of this organization. Maybe someone has a better idea?

ferdonline:
@tristan0x When you can let us know if you agree with this fix

.travis.yml
```diff
@@ -139,10 +139,10 @@ script:
   - mkdir -p build && pushd build
   - >
     cmake --warn-uninitialized --debug-output
+    -DHIGHFIVE_SINGLE_INCLUDE_TESTS=ON
```
tristan0x:
```suggestion
    -DHIGHFIVE_TEST_SINGLE_INCLUDES:BOOL=ON
```

include/highfive/H5DataSet.hpp
```diff
@@ -12,18 +12,19 @@
 #include <vector>
 
 #include "H5Object.hpp"
+#include "H5DataType.hpp"
+#include "H5DataSpace.hpp"
```
tristan0x: Can these headers be included in a lexicographic order?

include/highfive/H5Object.hpp
```diff
@@ -9,6 +9,8 @@
 #ifndef H5OBJECT_HPP
 #define H5OBJECT_HPP
 
+#include "H5Exception.hpp"
```
tristan0x: Any reason why `H5Exception.hpp` isn't part of the include section below. Besides standard includes should be a in [separated sections](https://github.com/BlueBrain/hpc-coding-conventions/tree/master/cpp/formatting#sort-includes-in-different-sections).

include/highfive/H5PropertyList.hpp
```diff
@@ -10,12 +10,15 @@
 #define H5PROPERTY_LIST_HPP
 
 #include "H5Object.hpp"
+#include "H5Exception.hpp"
```
tristan0x: Use lexicographic order if possible

tests/unit/CMakeLists.txt
```diff
@@ -40,3 +40,15 @@ if(HIGHFIVE_PARALLEL_HDF5)
 
 endif()
 
+
+option(HIGHFIVE_SINGLE_INCLUDE_TESTS "Enable testing single includes" FALSE)
```
tristan0x:
```suggestion
option(HIGHFIVE_TEST_SINGLE_INCLUDES "Enable testing of single includes" FALSE)
```

tests/unit/CMakeLists.txt
```diff
@@ -40,3 +40,15 @@ if(HIGHFIVE_PARALLEL_HDF5)
...
+
+if(HIGHFIVE_SINGLE_INCLUDE_TESTS)
```
tristan0x:
```suggestion
if(HIGHFIVE_TEST_SINGLE_INCLUDES)
```

include/highfive/H5Exception.hpp
```diff
@@ -14,6 +14,8 @@
 
 #include "bits/H5Utils.hpp"
 
+#include <H5Ipublic.h>  // for hid_t
```
tristan0x: Why is the include relative here? IMO it should be in between " " ?

ferdonline: this is a header from the hdf5 lib... I will change the order

PR #229, opened 2019-10-17
Manage HighFive dependecies in a clean way

alkino:


alkino:
I was using a cmake 3.15 feature...

CMake/PackageConfig.cmake
```diff
@@ -1,13 +1,22 @@
...
 target_include_directories(HighFive INTERFACE
   "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>"
   "$<INSTALL_INTERFACE:include>")
 target_include_directories(HighFive SYSTEM INTERFACE ${HDF5_INCLUDE_DIRS})
+target_link_libraries(HighFive INTERFACE ${HDF5_LIBRARIES})
+target_compile_definitions(HighFive INTERFACE ${HDF5_DEFINITIONS})
+
+# MPI
+if(HIGHFIVE_PARALLEL_HDF5)
+  target_link_libraries(HighFive INTERFACE MPI::MPI_C)
```
ferdonline: Use also the back-compat format ${MPI_C_LIBRARIES} and ${MPI_C_HEADER_DIR}

alkino: Ok, but I use `MPI_C_INCLUDE_PATH` because `MPI_C_HEADER_DIR` is only available since 3.10.

CMake/PackageConfig.cmake
```diff
@@ -1,9 +1,23 @@
...
   "$<INSTALL_INTERFACE:include>")
 target_include_directories(HighFive SYSTEM INTERFACE ${HDF5_INCLUDE_DIRS})
+target_link_libraries(HighFive INTERFACE ${HDF5_LIBRARIES})
+target_compile_definitions(HighFive INTERFACE ${HDF5_DEFINITIONS})
+
+# MPI
+if(HIGHFIVE_PARALLEL_HDF5)
+  target_include_directories(HighFive SYSTEM INTERFACE ${MPI_C_INCLUDE_PATH})
+  target_link_libraries(HighFive INTERFACE ${MPI_C_LIBRARIES})
+  if(CMAKE_VERSION VERSION_GREATER 3.13)
```
ferdonline: Why this `if`? Can't MPI_C_LINK_FLAGS be always there?

alkino: `target_link_options` is only available in recent version of cmake. And put this argument in `target_link_libraries` only work for gcc-like compiler.

ferdonline: Lets try with this
```
set_target_properties(HighFive PROPERTIES LINK_FLAGS ${MPI_C_LINK_FLAGS})
```
It should do what is intended, even if not specifically INTERFACE. The default is normally PUBLIC which in this case is equivalent.

alkino: Oups did it, but forget to push

ferdonline: Oopps... we really need an IF
```
CMake Error at CMake/PackageConfig.cmake:15 (set_target_properties):
  INTERFACE_LIBRARY targets may only have whitelisted properties.  The
  property "LINK_FLAGS" is not allowed.

bash-5.0$ cmake --version
cmake version 3.14.4
```

alkino:
Done ;)

CMake/PackageConfig.cmake
ferdonline: And no else?

alkino: Like before, if someone complains we will think about making something work on all platform for cmake from 3.0 to 3.16+.
But maybe we can improve cmake version

ferdonline: # In CMake 3.13+, LINK_FLAGS was converted to LINK_OPTIONS
See https://www.gitmemory.com/issue/LLNL/blt/266/503660028

alkino: But no.
`VERSION_GREATER_EQUAL` does not exist on old cmake version

PR #233, opened 2019-10-20
exist: always return true for /

### LINKED ISSUE #232: Spurious failure of file.exists("/");
Hi there,

I've run into a peculiar issue.

```C++
    auto getOrCreateGroup(std::string path)
    {
        if (!hi5_.file_.exist(path))
            hi5_.file_.createGroup(path);
        return hi5_.file_.getGroup(path);
    }
```

then calling 
```C++
getOrCreate("/");
```

is fine on debian with mpich, but isn't on fedora with openmpi (not sure how relevant these things are)

in the second case, it attempts to create a group for root "/" which already exists so it fails.

alkino:
Should fix #232

### DIFF
--- a/include/highfive/bits/H5Node_traits_misc.hpp
+++ b/include/highfive/bits/H5Node_traits_misc.hpp
@@ -175,17 +175,19 @@ inline std::vector<std::string> NodeTraits<Derivate>::listObjectNames() const {
 
 template <typename Derivate>
 inline bool NodeTraits<Derivate>::_exist(const std::string& node_name) const {
-    // The root path always exists, but H5Lexists return 0 or 1
-    // depending of the version of HDF5, so always return true for it
-    if (node_name == "/")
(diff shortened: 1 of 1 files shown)


include/highfive/bits/H5Node_traits_misc.hpp
```diff
@@ -175,6 +175,11 @@ inline std::vector<std::string> NodeTraits<Derivate>::listObjectNames() const {
 
 template <typename Derivate>
 inline bool NodeTraits<Derivate>::_exist(const std::string& node_name) const {
+    // The root path always exists, but H5Lexists return 0 or 1
```
ferdonline: We can't assume H5Lexists("/") will succeed because exceptions can happen.
It's easier if we put this `if` in the routine below (below the call to this `_exist()`, so that eventual exceptions are correctly raised.

PR #236, opened 2019-10-21
Add dependency on hpc-conding-conventions

alkino:


ferdonline:
I am not super found of automatic formatting hooks... For that to work we need to use always the same style and a formatter tool always at the same version. And still sometimes it can still produce ugly code.
I would say we adjust the style based on BBP conventions to make a minimum change, then check  for ugly code...

If you want an automatic way, then I I'd propose it to be applied only to the "diff" code.

tristan0x:
According to [BBP coding conventions](https://github.com/BlueBrain/hpc-coding-conventions/tree/master/cpp), you may use Clang Format 7 and the following [configuration file](https://github.com/BlueBrain/hpc-coding-conventions/blob/master/cpp/.clang-format)

tristan0x:
The `deps/git-cmake-format` is deprecated. You may remove it as well as everything that references it.
You can use [BBP Coding Conventions](https://github.com/BlueBrain/hpc-coding-conventions/tree/master/cpp) as a submodule instead to take benefits of:
* the `FindCmakeFormat`
* the `check-clang-format` make target that will check that code is properly formatted.
* Stay up to date as Clang Format supported version and configuration file will evolve over time.

Thank you @alkino for this initiative, and sorry for jumping in the conversation only now.

alkino:
No more travis, no more automatic, only integration

ferdonline:
I don't think we should reformat everything with hoc-coding-conventions. Each project can have its own set of rules, and minimal changes to the style have huge diffs, which is what just happened.
I updated the local .clang-format with things from the hpc-coding-conventions that I find important, but avoided touch more than that.
There is still the problem of manually nice formatted blocks. That's why I'm slightly in favor of using `clang-format/clang-format-diff.py`

PR #237, opened 2019-10-24
Build of documentation is now opt-in

alkino:


tdegeus:
Awesome!

alkino:
We should publish implicit rules of this project.

ferdonline:
Note that doc/CMakeLists already uses a variable to control documentation which is
`HIGH_FIVE_DOCUMENTATION`
 let's fix that.

alkino:
I remove this.

ferdonline:
If I may give my suggestion, I think we should

alkino:
We merge this and I publish the new version?

ferdonline:
Hum... so now we have a double condition, have this flag to TRUE and later call `make docs` :/
In tristan comment we have 
> I would either add a CMake option to enable/disable Doxygen build or remove it from all target (I prefer the second solution).

Which I agree, so be either one thing or the other

What's the problem of not having the option and let it issue the warning?

alkino:
@tdegeus ?

tdegeus:
Personally, I really think that building docs should be something that is opt-in. Ideally speaking I would even say that the end-user has to never worry about building docs as they would be published online. But in the mean-time it feels more natural to keep optional (advanced) dependencies to a minimum. For Boost I could see the point as you'd want user to test upon installation, but for the docs I do think the situation is different. 

On the other hand, it is true that currently one just sees a warning, which one can simply ignore.

ferdonline:
Since it was removed from the `ALL` target AND the dependency is optional, I don't see a good reason for this flag. Building docs must already be manually triggered. It is a warning, but that's about it. I prefer the warning other than having to do two things to have docs.

PR #239, opened 2019-10-29
AnnotateTraits: add deleteAttribute function

### LINKED ISSUE #213: Create a deleteAttribute in Annotate_Traits
Is there a way how to delete or rewrite an attribute using HighFive?

alkino:
Fix #213

### DIFF
--- a/tests/unit/tests_high_five_base.cpp
+++ b/tests/unit/tests_high_five_base.cpp
@@ -49,7 +49,7 @@ BOOST_AUTO_TEST_CASE(HighFiveBasic) {
 
     // check if the dataset exist
     bool dataset_exist = file.exist(DATASET_NAME + "_double");
-    BOOST_CHECK_EQUAL(dataset_exist, false);
+    BOOST_CHECK(!dataset_exist);
 
     // Create a dataset with double precision floating points
     DataSet dataset_double = file.createDataSet(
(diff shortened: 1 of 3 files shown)


tests/unit/tests_high_five_base.cpp
```diff
@@ -651,25 +651,21 @@ void readWriteAttributeVectorTest() {
         Group g = file.createGroup("dummy_group");
 
         // check that no attributes are there
-        std::size_t n = g.getNumberAttributes();
-        BOOST_CHECK_EQUAL(n, 0);
+        BOOST_CHECK_EQUAL(g.getNumberAttributes(), 0);
```
ferdonline: BOOST_CHECK_EQUAL, on error, this works pretty bad when trying to output a meaningful message.
I'd suggest to leave as it was (maybe changing the type to auto)

ferdonline:
Maybe there was some problem merging, maybe a fast-forward, because we see some commits that shouldnt be here.

alkino:
better like that @ferdonline ?

ferdonline:
I am just wondering... maybe we should enforce good behavior, by checking (per the docs)
"This function should not be used when attribute identifiers are open on loc_id"

alkino:
I cannot see a function to know if a attribute identifiers  is open.

ferdonline:
https://support.hdfgroup.org/HDF5/doc1.8/RM/RM_H5I.html#Identify-GetRef

alkino:
I think of it, but we don't know if internally they use some refs.

alkino:
Can we merge it like this? @ferdonline

PR #244, opened 2019-11-04
Avoid conversion warnings being treated as errors

### LINKED ISSUE #243: Build of 2.1 Failing on Linux due to warnings
Hi all,

When building HighFive 2.1, it seems to fail on Linux due to warnings (in the examples): https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=85546&view=logs

Just wanted to give you a heads-up. 
I guess for the recipe, we could disable erroring on warnings. (https://github.com/conda-forge/highfive-feedstock)

Cheers!

ferdonline:
Conversion warnings are too common to be treated as errors (since we use -Werror). Such warnings are even found within boost libraries, so there is no way around.
This patch simply turns them non-errors again. The warnings are still raised.
Also, avoiding passing unknown options to ICC.
Fixes #243


ferdonline:
Nice @alkino. Please approve

PR #246, opened 2019-11-05
Since v2.0 CXX11 is mandatory by CMake

alkino:


include/highfive/H5PropertyList.hpp
```diff
@@ -45,10 +45,8 @@ class PropertyList {
   public:
     ~PropertyList();
 
-#ifdef H5_USE_CXX11
```
alkino: It was buggy, `H5_USE_CXX11` is always **defined** to `0` or to `1`, but **defined**.

ferdonline:
We should at some point also make a “modernization” of the code for c++11.
Maybe before merging check if some constructs can be simplified throughout  the code

alkino:
I'm not sure, by reading the code often I don't see offending c++03 or older code. We should do it when we see it.

PR #247, opened 2019-11-06
Make ElementSet work in N-dim

alkino:


tests/unit/tests_high_five_base.cpp
```diff
@@ -810,6 +810,37 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(selectionArraySimple, T, dataset_test_types) {
...
+
+    {
+        int value;
+        set.select(ElementSet{{1,1}}).read(value);
+        BOOST_CHECK_EQUAL(value, 5);
+    }
+
+    {
+        int value[2];
+        set.select(ElementSet{0,0,2,2}).read(value);
```
ferdonline: I think we should avoid the user from doing this. For 2D the indexes should be 2D and so on. 
If there any special case we need this internally, then this constructor should be protected.

alkino: I don't need it, it's only that I want to write an example and so see that it's not working.

ferdonline: Yes, nothing against the test itself. It's related to the API. We should throw if the dimensions don't match because that's most probably a user mistake. So in this case we should expect an exception.

alkino:
I cannot do better. I want to compile-check that all the inner vector have the same size. I tried with `template <size_t N> std::array<size_t, N>`
But we lost initializer_list

include/highfive/bits/H5Slice_traits.hpp
```diff
@@ -23,7 +23,28 @@ class SliceTraits;
 
 class ElementSet {
   public:
+    ///
```
ferdonline: I think we should template ElementSet to the number of dimensions and use `std::array`
`ElementSet(std::initializer_list<std::array<std::size_t, NDIM>> list);`

ferdonline:
I fyou think the sollution with std::array takes too long to implement we can do that later. This solution is anyway Ok

PR #254, opened 2019-12-02
Missing semicolon

marianheil:
Without the semicolon building HighFive fails with:
```
HighFive/tests/unit/tests_high_five_parallel.cpp:40:1: error: expected initializer before ‘template’
   40 | template <typename T>
```

Tested with `gcc 9.2.1` and `boost 1.67.0.2`.

### DIFF
--- a/tests/unit/tests_high_five_parallel.cpp
+++ b/tests/unit/tests_high_five_parallel.cpp
@@ -34,7 +34,7 @@ struct MpiFixture {
     ~MpiFixture() { MPI_Finalize(); }
 };
 
-BOOST_GLOBAL_FIXTURE(MpiFixture)
+BOOST_GLOBAL_FIXTURE(MpiFixture);
 
 
 template <typename T>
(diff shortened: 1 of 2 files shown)


marianheil:
Judging from the failed CI it looks like boost removed the semicolon from the `BOOST_GLOBAL_FIXTURE` macro, at some point between version 1.58 and 1.67.

alkino:
Thank you for the patch.

Since, 1.59: https://github.com/boostorg/test/commit/3f7216db3db2e11a768d8d0c8bb18632f106c466#diff-5791b6c006a0cb19d6f44c365a1c4678

What can we do, try to disable the warning only for this line in a portable way?
Put a semicolon depending of the boost version?

alkino:
I find it ugly

tests/unit/tests_high_five_parallel.cpp
```diff
@@ -35,7 +35,9 @@ struct MpiFixture {
 };
 
 BOOST_GLOBAL_FIXTURE(MpiFixture)
-
+#if BOOST_VERSION >= 105900
```
ferdonline: Thanks @alkino, in the end I think we need to do something like this.
I would maybe create a `#define BOOST_GLOBAL_FIXTURE_END` containing the semicolon (or empty) (in tests_high_five.hpp?) and use it here.

alkino:
Let's merge it.

PR #257, opened 2019-12-05
Switching to modern CMake

tdegeus:
**Edited based on the comment on the namespace below.**

I started playing around with modern CMake. I think it could be very interesting here as it gives a lot of room for flexible automatic dependencies and customisation by the end-user. I have prepared a first version that for now only installs the library and provides CMake support for the end-user. I've left out testing, examples, ... because at the moment I think it is good to first discuss. 

The point here is that a target `HighFive` is exported that contains the headers of the library itself, and includes targets for the required dependency (`HighFive::HDF5`) and optional dependencies (`HighFive::Boost`, `HighFive::Eigen`, and `HighFive::xtensor`). At compile time, these targets need to be defined, but their content allows for flexibility. The user can define them, or, when they are not defined, they are filed in `HighFiveConfig.cmake` depending on availability of the optional dependencies. If the libraries are found the defines (`H5_USE_BOOST`, ...) are also set to the relevant target (`HighFive::Boost`, ...).

> Enabling tests and examples will hereafter not be hard, the target `HighFive` in the main `CMakeLists.txt` contains all the goodies and install/test time, without exporting any values/decisions that are set at install/test time to the package. Leaving the end-user with full flexibility.

For testing, I suggest a conda environment. Then install (in the activated environment) by
```bash
cd /path/to/HighFive
mkdir build
cd build
cmake .. -DCMAKE_INSTALL_PREFIX:PATH="${CONDA_PREFIX}"
```

One can then run an example:
```cpp
#include <highfive/H5Easy.hpp>

int main()
{
    std::vector<size_t> a = {1, 2, 3};
    H5Easy::File file("test.h5", H5Easy::File::Overwrite);
    H5Easy::dump(file, "/a", a);
    return 0;
}
```

with the following `CMakeLists.txt`
```cmake
cmake_minimum_required(VERSION 3.1)
project(test)
add_executable(test main.cpp)
target_link_libraries(test HighFive)
```

Notice that depending on availability Boost, Eigen3, and xtensor will be auto-activated.


tdegeus:
(The CI naturally fails, as I excluded tests and examples for now)

ferdonline:
Hi Tom. Thanks for your effort. Improvement of CMake is always welcomed.
My only requirement is to not break compatibility. Starting from scratch might sound great but we know that CMake is complex and we might not be able to anticipate differences and break integration with other projects. For instance I see that the targets now have a namespace. That is a breaking change I don't think we need.

Feel free to play around, but I'm more in favor of using this as a "reference" to take ideas from and apply them little by little to the current build. 
If things start working nicely we can keep it in a branch and start using it from other projects but only when we feel confident enough (maybe on a major release) it would be considered to be merged.

tdegeus:
@ferdonline I think that modernising CMake does not have to break the current functionality: having `HighFive::HighFive` or `HighFive` is a matter of style. I have the impression that `HighFive::HighFive` could be the preferred convention, but I will change it to `HighFive`.

tdegeus:
@ferdonline Beyond the namespace maybe you can explain your comments a bit further? Maybe you can list some examples of functionality that should be kept (except of course testing and compile examples, that should obviously be integrated).

SylvainCorlay:
I support the initiative of @tdegeus to switch to a modern cmake.

PR #273, opened 2020-01-22
Fix shortcut syntax with plain C arrays

ferdonline:
Extracted from the Initial work towards fixed-len strings


ferdonline:
@alkino @tristan0x Any remarks here?

PR #277, opened 2020-01-28
Support for Fixed-length strings

### LINKED ISSUE #94: Support for fixed-length strings
Hi,

I am trying to read a dataset with this type:

```C++
DATATYPE H5T_STRING {
  STRSIZE 50;
  STRPAD H5T_STR_NULLPAD;
  CSET H5T_CSET_ASCII;
  CTYPE H5T_C_S1;
}
```
Glancing at `AtomicType<std::string>::AtomicType()` it seems that only variable-length strings are expected.

Is this something that can be supported in the future?

### LINKED ISSUE #275: const / non-const data matching different write() methods
Having a buffer const or non const would match different write() methods, e.g.:
 - `const int*` -> picks `write(const T*)`  (as expected)
 - `int*` picks `write(const& T)` unexpected, but probably it's a better match than the other overload which has const

Changing `write(const T*)` -> `write(T* const)` will however not match buffers that are const.

The feature was so weak that removing the `write(const T*)` overload altogether would work, since `write(const T&)` can still be matched. However there is dimension checking which is what the last overload is meant to not do.

This will probably be solved with renaming the method to allow the user to be explicit if he wants his buffer to be treated as dimensionless, without checks. Suggestion: `write_raw()`
Ideas @tristan0x?

ferdonline:
## This PR
* Adds core support and API for fixed length strings
* Change `write(const T*)` to `write_raw(const T*, const DataType& dtype = DataType())`.
 The change of the name is intended to fix #275, while the introduction of the optional dtype allows users to raw write data structures whose type is ambiguous (e.g. char[] is fixed-string or int8 array?)

## Rationale
The type `char[N]` is now considered a valid Atomic type. The user can therefore
```c++
// create  a dataset for fixed length strings as 
ds = file.createDataSet<char[10]>("ds1", dataspace);
// write raw data
char my_strings[][10] = {"abc", "def"};
ds.write(my_strings);
```

While the previous API is great for performance, a high level API was also introduced, compatible with the short-circuit syntax:
```c++
// Have FixedLenStringArray initialised from std::string containers
FixedLenStringArray<10> arr{"abcde", "010101"};
// create DS and write in single step
file.createDataSet("ds3", arr);
```
Data is internally stored as `vector<array<char,N>>`, as returned by Hdf5 directly. As so operator returns char*, but the user may request an `std::string` using `getString()`

### Extras
* Simplifies internal converters / Improves const-correctness
* Introduces `BufferInfo` in ReadWrite_misc to help detecting a buffer type with the help of the DataSet type. 
* Spins-off Eigen Converter routines t its own source

fixes #94 
fixes #275

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,6 @@ if(NOT COMPILER_SUPPORTS_CXX11)
   message(FATAL_ERROR "HighFive version >= 2.0 requires c++ standard >= c++11")
 endif()
 
-
 # Compat within Highfive 2.x series
 set(USE_BOOST ON CACHE BOOL "Enable Boost Support")
 set(USE_EIGEN OFF CACHE BOOL "Enable Eigen testing")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,8 @@ if(NOT COMPILER_SUPPORTS_CXX11)
   message(FATAL_ERROR "HighFive version >= 2.0 requires c++ standard >= c++11")
 endif()
 
+add_definitions("-ferror-limit=1")
+
 # Compat within Highfive 2.x series
 set(USE_BOOST ON CACHE BOOL "Enable Boost Support")
 set(USE_EIGEN OFF CACHE BOOL "Enable Eigen testing")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,6 @@ if(NOT COMPILER_SUPPORTS_CXX11)
   message(FATAL_ERROR "HighFive version >= 2.0 requires c++ standard >= c++11")
 endif()
 
-add_definitions("-ferror-limit=1")
 
 # Compat within Highfive 2.x series
 set(USE_BOOST ON CACHE BOOL "Enable Boost Support")
--- a/include/highfive/H5DataType.hpp
+++ b/include/highfive/H5DataType.hpp
@@ -132,15 +132,24 @@ class FixedLenStringArray {
     std::string getString(std::size_t index) const;
 
     // Container interface
-    inline const char* operator[](std::size_t i) const { return datavec[i].data(); }
+    inline const char* operator[](std::size_t i) const noexcept { return datavec[i].data(); }
     inline const char* at(std::size_t i) const { return datavec.at(i).data(); }
     inline bool empty() const noexcept { return datavec.empty(); }
     inline std::size_t size() const noexcept { return datavec.size(); }
-    inline void resize(std::size_t i) { datavec.resize(i); }
+    inline void resize(std::size_t n) { datavec.resize(n); }
     inline const char* front() const { return datavec.front().data(); }
     inline const char* back() const { return datavec.back().data(); }
     inline char* data() noexcept { return datavec[0].data(); }
     inline const char* data() const noexcept { return datavec[0].data(); }
+    // Use the underlying iterator
+    typedef typename std::vector<std::array<char, N>>::iterator iterator;
+    typedef typename std::vector<std::array<char, N>>::const_iterator const_iterator;
+    inline iterator begin() noexcept { return datavec.begin(); }
+    inline iterator end() noexcept { return datavec.end(); }
+    inline const_iterator begin() const noexcept { return datavec.begin(); }
+    inline const_iterator cbegin() const noexcept { return datavec.cbegin(); }
+    inline const_iterator end() const noexcept { return datavec.end(); }
+    inline const_iterator cend() const noexcept { return datavec.cend(); }
 
   private:
     typedef typename std::vector<std::array<char, N>> vector_t;
--- a/include/highfive/H5DataType.hpp
+++ b/include/highfive/H5DataType.hpp
@@ -92,7 +92,58 @@ class AtomicType : public DataType {
 
     typedef T basic_type;
 };
-}
+
+
+///
+/// \brief A structure representing a set of fixed-length strings
+///
+/// Although fixed-len arrays can be created 'raw' without the need for
+/// this structure, to retrieve results efficiently it must be used.
+///
+template <std::size_t N>
+class FixedLenStringArray {
+  public:
+    inline FixedLenStringArray() {};
+
+    ///
+    /// \brief Create a FixedStringArray from a raw contiguous buffer
+    ///
+    FixedLenStringArray(const char array[][N], std::size_t length);
+
+    ///
+    /// \brief Create a FixedStringArray from a sequence of strings.
+    ///
+    /// Such conversion involves a copy, original vector is not modified
+    ///
+    FixedLenStringArray(const std::string* iter_begin, const std::string* iter_end);
+
+    FixedLenStringArray(const std::vector<std::string> & vec);
+
+    FixedLenStringArray(const std::initializer_list<std::string> &);
+
+    ///
+    /// \brief Append an std::string to buffer structure
+    ///
+    void push_back(const std::string&);
+
+    ///
+    /// \brief Retrieve a string from the structure as std::string
+    ///
+    std::string operator[](std::size_t index) const;
+
+
+    // Container API
+    inline std::size_t size() const { return datavec.size(); }
+    inline const char* data() const { return datavec[0].data(); }
+    void resize(const std::size_t& new_size) { datavec.resize(new_size); }
+
+  private:
+    typedef std::vector<std::array<char, N>> vector_t;
+    vector_t datavec;
+};
+
+
+}  // namespace HighFive
 
 #include "bits/H5DataType_misc.hpp"
 
--- a/include/highfive/H5DataType.hpp
+++ b/include/highfive/H5DataType.hpp
@@ -126,6 +126,8 @@ class FixedLenStringArray {
     ///
     void push_back(const std::string&);
 
+    void push_back(const std::array<char, N>&);
+
     ///
     /// \brief Retrieve a string from the structure as std::string
     ///
@@ -152,7 +154,44 @@ class FixedLenStringArray {
     inline const_iterator cend() const noexcept { return datavec.cend(); }
 
   private:
-    typedef typename std::vector<std::array<char, N>> vector_t;
+    using vector_t = typename std::vector<std::array<char, N>>;
+
+  public:
+    using iterator = typename vector_t::iterator;
+    using const_iterator = typename vector_t::const_iterator;
+    using reverse_iterator = typename vector_t::reverse_iterator;
+    using const_reverse_iterator = typename vector_t::const_reverse_iterator;
+    using value_type = typename vector_t::value_type;
+
+    iterator begin() noexcept {
+        return datavec.begin();
+    }
+    iterator end() noexcept {
+        return datavec.end();
+    }
+
+    const_iterator begin() const noexcept {
+        return datavec.begin();
+    }
+    const_iterator end() const noexcept {
+        return datavec.end();
+    }
+
+    reverse_iterator rbegin() noexcept {
+        return datavec.rbegin();
+    }
+    reverse_iterator rend() noexcept {
+        return datavec.rend();
+    }
+
+    const_reverse_iterator rbegin() const noexcept {
+        return datavec.rbegin();
+    }
+    const_reverse_iterator rend() const noexcept {
+        return datavec.rend();
+    }
+
+  private:
     vector_t datavec;
 };
 
--- a/include/highfive/H5DataType.hpp
+++ b/include/highfive/H5DataType.hpp
@@ -66,6 +66,8 @@ class DataType : public Object {
     friend class Attribute;
     friend class File;
     friend class DataSet;
+
+    DataType(hid_t type_hid);
 };
 
 ///
--- a/include/highfive/H5DataType.hpp
+++ b/include/highfive/H5DataType.hpp
@@ -134,7 +134,7 @@ class FixedLenStringArray {
 
     // Container API
     inline std::size_t size() const { return datavec.size(); }
-    inline const char* data() const { return datavec[0].data(); }
+    inline char* data() const { return const_cast<char*>(datavec[0].data()); }
     void resize(const std::size_t& new_size) { datavec.resize(new_size); }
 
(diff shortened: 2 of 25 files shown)


ferdonline:
Used features not c++11... will try to fix them

include/highfive/H5DataType.hpp
```diff
@@ -80,7 +92,58 @@ class AtomicType : public DataType {
...
+
+    ///
+    /// \brief Append an std::string to buffer structure
+    ///
+    void push_back(const std::string&);
+
+    ///
+    /// \brief Retrieve a string from the structure as std::string
+    ///
+    std::string operator[](std::size_t index) const;
```
tristan0x: The copy in `operator[]` makes me cry here.

With C++17, we could use `std::string_view`.
But GSL provides `string_span` that we could probably leverage already.

include/highfive/bits/H5DataType_misc.hpp
```diff
@@ -151,9 +180,75 @@ inline AtomicType<std::complex<double> >::AtomicType()
...
+inline void FixedLenStringArray<N>::push_back(const std::string& src) {
+    datavec.emplace_back();
+    const size_t length = std::min(N - 1 , src.length());
+    std::memcpy(datavec.back().data(), src.c_str(), length);
+    datavec.back()[length] = 0;
+}
+
+template <std::size_t N>
+inline std::string FixedLenStringArray<N>::operator[](std::size_t i) const {
+    return std::string(datavec[i].data());
```
tristan0x: The copy in `operator[]` makes me want to cry ;)

With C++17, we could use `std::string_view`.
But the GSL already provides `string_span` that we could probably leverage.

src/examples/read_write_fixedlen_string.cpp
```diff
@@ -0,0 +1,53 @@
...
+        FixedLenStringArray<10> arr{"0000000", "1111111"};
+        auto ds = file.createDataSet("ds3", arr);
+
+        FixedLenStringArray<4> array_back;
+        ds.read(array_back);
+        std::cout << "First item is '" << array_back[0] << "'" << std::endl
+                  << "Second item is '" << array_back[1] << "'" << std::endl;
+
+
+    } catch (Exception& err) {
```
tristan0x:
```suggestion
    } catch (const Exception& err) {
```

src/examples/read_write_fixedlen_string.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include <string>
+#include <vector>
+
+#include <highfive/H5File.hpp>
+#include <highfive/H5DataSet.hpp>
+#include <highfive/H5DataSpace.hpp>
+
+using namespace HighFive;
+
+const std::string FILE_NAME("create_dataset_string_example.h5");
```
tristan0x:
```suggestion
static const std::string FILE_NAME("create_dataset_string_example.h5");
```

src/examples/read_write_fixedlen_string.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+    try {
+        // Create a new file using the default property lists.
+        File file(FILE_NAME, File::ReadWrite | File::Create | File::Truncate);
+        const char strings_fixed[][16] = {"abcabcabcabcabc", "123123123123123"};
+
+        // create a dataset ready to contains strings of the size of the vector
+        file.createDataSet<char[10]>("ds1", DataSpace(2)).write(strings_fixed);
+
+        // Without specific type info this will crate an int8 dataset
```
tristan0x:
```suggestion
        // Without specific type info this will create an int8 dataset
```

src/examples/read_write_fixedlen_string.cpp
```diff
@@ -0,0 +1,53 @@
...
+        file.createDataSet("ds2", strings_fixed);
+
+        // Now test the new interface type
+        //FixedLenStringArray<10> arr(std::vector<std::string>{"0000000", "1111111"});
+        FixedLenStringArray<10> arr{"0000000", "1111111"};
```
tristan0x: This is really awesome!

ferdonline:
passing 🎉

include/highfive/bits/H5DataType_misc.hpp
tristan0x: I created a separate issue about that, see #277

CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ if(NOT COMPILER_SUPPORTS_CXX11)
   message(FATAL_ERROR "HighFive version >= 2.0 requires c++ standard >= c++11")
 endif()
 
+
```
tristan0x: Is it on purpose?

include/highfive/H5DataType.hpp
```diff
@@ -80,7 +92,58 @@ class AtomicType : public DataType {
...
+    /// Such conversion involves a copy, original vector is not modified
+    ///
+    FixedLenStringArray(const std::string* iter_begin, const std::string* iter_end);
+
+    explicit FixedLenStringArray(const std::vector<std::string> & vec);
+
+    FixedLenStringArray(const std::initializer_list<std::string> &);
+
+    ///
+    /// \brief Append an std::string to buffer structure
```
tristan0x:
```suggestion
    /// \brief Append a std::string to the buffer structure
```

?

include/highfive/H5DataType.hpp
```diff
@@ -80,7 +92,58 @@ class AtomicType : public DataType {
...
+
+    ///
+    /// \brief Retrieve a string from the structure as std::string
+    ///
+    std::string operator[](std::size_t index) const;
+
+
+    // Container API
+    inline std::size_t size() const noexcept { return datavec.size(); }
+    inline char* data() const { return const_cast<char*>(datavec[0].data()); }
```
tristan0x:
```suggestion
    inline char* data() const noexcept { return const_cast<char*>(datavec[0].data()); }
```

Also, IMO `data()` member function is not const because it returns a pointer that allow modification of datavec.
I would implement the const and non-const version of `data`, like `std::vector`:

```c++
value_type* data() noexcept;
const value_type* data() const noexcept;
```

include/highfive/H5DataType.hpp
```diff
@@ -80,7 +92,58 @@ class AtomicType : public DataType {
...
+    void resize(std::size_t new_size) { datavec.resize(new_size); }
```
tristan0x: In case you want this class to mimic std::vector, there is also `void std::vector::resize(size_type n, const value_type& val)`. But we shouldn't add it if unused/tested in the project.

include/highfive/H5DataType.hpp
```diff
@@ -80,7 +92,58 @@ class AtomicType : public DataType {
...
+
+
+///
+/// \brief A structure representing a set of fixed-length strings
+///
+/// Although fixed-len arrays can be created 'raw' without the need for
+/// this structure, to retrieve results efficiently it must be used.
+///
+template <std::size_t N>
+class FixedLenStringArray {
```
tristan0x: IMO the `begin`, `end`, `cbegin`, `cend`, `rbegin`, and `rend`, `at`, `front`, `back`, `empty`, and `size`, member methods would be useful for the user.

src/examples/read_write_fixedlen_string.cpp
```diff
@@ -0,0 +1,53 @@
...
+        const char strings_fixed[][16] = {"abcabcabcabcabc", "123123123123123"};
+
+        // create a dataset ready to contains strings of the size of the vector
+        file.createDataSet<char[10]>("ds1", DataSpace(2)).write(strings_fixed);
+
+        // Without specific type info this will create an int8 dataset
+        file.createDataSet("ds2", strings_fixed);
+
+        // Now test the new interface type
+        //FixedLenStringArray<10> arr(std::vector<std::string>{"0000000", "1111111"});
```
tristan0x: Dead code here!

tests/unit/tests_high_five_base.cpp
```diff
@@ -1043,7 +1078,7 @@ void readWriteShuffleDeflateTest() {
 }
 
 BOOST_AUTO_TEST_CASE_TEMPLATE(ReadWriteShuffleDeflate, T, numerical_test_types) {
-    readWriteShuffleDeflateTest<T>();
+    //readWriteShuffleDeflateTest<T>();
```
tristan0x: dead code here!

CMakeLists.txt
ferdonline: not really. addressing your review I will take care of minimizing changes

include/highfive/H5DataType.hpp
ferdonline: my english tells me to use `an` since we pronounce 'esse' in s-t-d 😃 but will ask Jim

include/highfive/H5DataType.hpp
ferdonline: I dont think that's required API for this class.

include/highfive/H5DataType.hpp
ferdonline: I thought about inheriting instead of composing... still not sure the best option here

include/highfive/bits/H5Slice_traits_misc.hpp
```diff
@@ -88,36 +90,32 @@ SliceTraits<Derivate>::select(const std::vector<size_t>& offset,
...
     std::vector<size_t> dims = space.getDimensions();
     std::vector<hsize_t> counts(dims.size());
     std::copy(dims.begin(), dims.end(), counts.begin());
     counts[dims.size() - 1] = 1;
     std::vector<hsize_t> offsets(dims.size(), 0);
 
     H5Sselect_none(space.getId());
-    for (std::vector<size_t>::const_iterator i = columns.begin();
-         i != columns.end(); ++i) {
+    for (std::vector<size_t>::const_iterator i = columns.begin(); i != columns.end();
```
alkino: for range based loop here

include/highfive/H5DataType.hpp
tristan0x: I usually prefer encapsulation over inheritance when it comes to standard containers.
I prefer to hide complexity, and I think their destructors are not virtual.

include/highfive/H5DataType.hpp
```diff
@@ -115,30 +115,35 @@ class FixedLenStringArray {
...
     ///
     /// \brief Retrieve a string from the structure as std::string
     ///
-    std::string operator[](std::size_t index) const;
-
+    std::string getString(std::size_t index) const;
 
-    // Container API
+    // Container interface
+    inline const char* operator[](std::size_t i) const { return datavec[i].data(); }
```
tristan0x:
```suggestion
    inline const char* operator[](std::size_t i) const noexcept { return datavec[i].data(); }
```

include/highfive/H5DataType.hpp
```diff
@@ -115,30 +115,35 @@ class FixedLenStringArray {
...
+    inline const char* at(std::size_t i) const { return datavec.at(i).data(); }
+    inline bool empty() const noexcept { return datavec.empty(); }
     inline std::size_t size() const noexcept { return datavec.size(); }
-    inline char* data() const { return const_cast<char*>(datavec[0].data()); }
-    void resize(std::size_t new_size) { datavec.resize(new_size); }
+    inline void resize(std::size_t i) { datavec.resize(i); }
```
tristan0x: Nitpicking, but `i` refers more to an index than a size.
```suggestion
    inline void resize(std::size_t n) { datavec.resize(n); }
```

src/examples/read_write_fixedlen_string.cpp
```diff
@@ -34,10 +34,10 @@ int main(void) {
         file.createDataSet("ds2", strings_fixed);
 
         // Now test the new interface type
-        //FixedLenStringArray<10> arr(std::vector<std::string>{"0000000", "1111111"});
-        FixedLenStringArray<10> arr{"0000000", "1111111"};
+        FixedLenStringArray<10> arr{"0000000", "1111111"};  // also accepts std::vector
```
tristan0x: I would at least ensure that the code compiles when passing a vector. If it is done elsewhere, please ignore this comment.

include/highfive/H5DataType.hpp
```diff
@@ -115,30 +115,35 @@ class FixedLenStringArray {
...
+    inline const char* front() const { return datavec.front().data(); }
+    inline const char* back() const { return datavec.back().data(); }
+    inline char* data() noexcept { return datavec[0].data(); }
+    inline const char* data() const noexcept { return datavec[0].data(); }
```
tristan0x: Thank you for these new methods. Unless you think it is not useful at all, I would add the begin() and end() (const and non const) methods to allow usage of the STL methods like copy, transform, back_inserter, ...

ferdonline: The problem I see with begin() and end() is that we need to define a custom iterator because we cant return `array<char, N>` as `char[N]`. Or you think it's ok we return
array<char, N>* ?

src/examples/read_write_fixedlen_string.cpp
ferdonline: Yep, the unit tests have that ctor :)

include/highfive/H5DataType.hpp
tristan0x: I will write snippets of code using FixedLenStringArray and STL algorithm that should make sense to see what is the best type we should use for iterator::value_type
We can do that in a separate change to not block this PR.

include/highfive/bits/H5Slice_traits_misc.hpp
ferdonline: we still to make a global c++11 modernization... I only "fixed" the line splitting

include/highfive/H5DataType.hpp
ferdonline: I had implemented the iterator methods in a simple way. Unless we want value_type of the iterator to be char*, maybe we were good.

ferdonline:
@tristan0x @alkino could you please review in order to merge or discard this? Its a pity now we have to solve conflicts because this stayed on hold 😞

PR #284, opened 2020-02-10
Unlink

tvandera:
Hi,

I've added `NodeTraits::unlink` that calls `H5Ldelete`. Suggestions welcome.

Cheers,
Tom


tvandera:
I'll do the extra test. T

PR #293, opened 2020-02-20
H5Easy: Simplifying xtensor implementation

tdegeus:


include/highfive/h5easy_bits/H5Easy_xtensor.hpp
```diff
@@ -75,6 +67,22 @@ struct load_impl
...
 
+// front-end
+template <class T>
+inline DataSet dump(File& file,
+                    const std::string& path,
+                    T&& data,
+                    DumpMode mode)
+{
+    if (!file.exist(path)) {
+        return detail::xtensor::dump_impl(file, path, data);
```
ferdonline: @alexsavulescu data is universal ref, why dont we std::forward(data)? This looks dangerous (copies may be made!)

ferdonline: Actually, what was the prob with const-ref?

tdegeus: We could indeed use `std::forward`, though for the moment there is not reason to prefer `T&& data` over `const T& data`. So I can change it to the latter if you prefer.

alexsavulescu: Yes, we should go with const T& data.

tdegeus: As you wish: https://github.com/BlueBrain/HighFive/pull/299

Issue #294, opened 2020-02-20
Towards an H5Easy opencv support (incl. rough prototype for H5Easy::load)


spacescientist:
Incidentally, since I was so happy about #286 and since we also mentioned a possible opencv overload, I've given a try and somehow managed to make HighFive read a dataset from an HDF file and store it as an opencv `cv::Mat_<double>`.

It's not nice at all as it stands: it would still need to be generalised beyond `double` and it's moreover incomplete (since there is only `load` for now; I haven't touched `dump`), but it seems to work in the simple setting I used. I'm surprised `cv::Mat_` understood `rows` and `cols`, but it apparently did.

##### H5Easy_opencv.hpp (prototype)
```cpp
#ifndef H5EASY_BITS_OPENCV_HPP
#define H5EASY_BITS_OPENCV_HPP

#include "../H5Easy.hpp"
#include "H5Easy_misc.hpp"
#include "H5Easy_scalar.hpp"  // to get the basic "load_impl"

////#ifdef H5_USE_OPENCV

namespace H5Easy {

namespace detail {

namespace opencv {

// return the shape of the opencv-object as "std::vector<size_t>"
template <class T>
inline std::vector<size_t> shape(const T& data)
{
    return std::vector<size_t>(data.shape().cbegin(), data.shape().cend());
}


// load opencv-object from DataSet
template <class T>
struct load_impl
{
    static T run(const File& file, const std::string& path)
    {
        DataSet dataset = file.getDataSet(path);
        std::vector<size_t> dims = dataset.getDimensions();
        T data(dims[0], dims[1]);
        dataset.read(reinterpret_cast<double*>(data.data));
        //                             ^ ! Shouldn't be explicitly double of course TODO.
        return data;
    }
};

}  // namespace opencv

// front-end
template <class T>
struct load_impl<cv::Mat_<T>>
{
    static cv::Mat_<T> run(const File& file, const std::string& path)
    {
        return detail::opencv::load_impl<cv::Mat_<T>>::run(file, path);
    }
};


}  // namespace detail

}  // namespace H5Easy

////////#endif  // H5_USE_OPENCV

#endif  // H5EASY_BITS_OPENCV_HPP
```

##### H5Easy.hpp
```cpp
// I simply added at the end of H5Easy.hpp:
#include "h5easy_bits/H5Easy_opencv.hpp"
```

And modified the MWE to:
##### main.cpp
```cpp
#include <iostream>
#include <Eigen/Eigen>
[...]

PR #295, opened 2020-02-21
Adding some H5Easy support for OpenCV (rough prototype)

spacescientist:
See #294. 

##### MWE using `cv::Mat_<double>`
```cpp
#include <iostream>
#include <opencv2/core.hpp>
#include "highfive/H5Easy.hpp"

int main()
{
        H5Easy::File input_hdf("../in.hdf", H5Easy::File::ReadOnly);

        // Works!
        auto data_cv = H5Easy::load<cv::Mat_<double>>(input_hdf, "/data");

        std::cout << data_cv << std::endl;
        std::cout << data_cv.rows << 'x' << data_cv.cols << std::endl;
        std::cout << data_cv.row(2) << std::endl;

        return 0;
}
```

Cheers.

### DIFF
--- a/include/highfive/h5easy_bits/H5Easy_opencv.hpp
+++ b/include/highfive/h5easy_bits/H5Easy_opencv.hpp
@@ -1,0 +1,66 @@
+/*
+ *  Copyright (c), 2017, Adrien Devresse <adrien.devresse@epfl.ch>
+ *
+ *  Distributed under the Boost Software License, Version 1.0.
+ *    (See accompanying file LICENSE_1_0.txt or copy at
+ *          http://www.boost.org/LICENSE_1_0.txt)
+ *
+ */
+#ifndef H5EASY_BITS_OPENCV_HPP
+#define H5EASY_BITS_OPENCV_HPP
+
+#include "../H5Easy.hpp"
+#include "H5Easy_misc.hpp"
+#include "H5Easy_scalar.hpp"  // to get the basic "load_impl"
+
+////#ifdef H5_USE_OPENCV
+
+namespace H5Easy {
+
+namespace detail {
+
+namespace opencv {
+
+// return the shape of the opencv-object as "std::vector<size_t>"
+template <class T>
+inline std::vector<size_t> shape(const T& data)
+{
+    return std::vector<size_t>(data.shape().cbegin(), data.shape().cend());
+}
+
+
+// load opencv-object from DataSet
+template <class T>
+struct load_impl
+{
+    static T run(const File& file, const std::string& path)
+    {
+        DataSet dataset = file.getDataSet(path);
+        std::vector<size_t> dims = dataset.getDimensions();
+        T data(dims[0], dims[1]);
+        dataset.read(reinterpret_cast<double*>(data.data));
+        //                             ^ ! Shouldn't be explicitly double of course TODO.
+        return data;
+    }
+};
+
+}  // namespace opencv
+
+// front-end
+template <class T>
+struct load_impl<cv::Mat_<T>>
+{
+    static cv::Mat_<T> run(const File& file, const std::string& path)
+    {
+        return detail::opencv::load_impl<cv::Mat_<T>>::run(file, path);
+    }
+};
+
+
+}  // namespace detail
+
+}  // namespace H5Easy
+
+////////#endif  // H5_USE_OPENCV
+
+#endif  // H5EASY_BITS_OPENCV_HPP
--- a/include/highfive/h5easy_bits/H5Easy_opencv.hpp
+++ b/include/highfive/h5easy_bits/H5Easy_opencv.hpp
@@ -28,31 +28,32 @@ inline std::vector<size_t> shape(const T& data)
     return std::vector<size_t>(data.shape().cbegin(), data.shape().cend());
 }
 
-
-// load opencv-object from DataSet
-template <class T>
+template<typename _Tp, template<typename> class T>
 struct load_impl
 {
-    static T run(const File& file, const std::string& path)
+    static T<_Tp> run(const File& file, const std::string& path)
     {
         DataSet dataset = file.getDataSet(path);
         std::vector<size_t> dims = dataset.getDimensions();
-        T data(dims[0], dims[1]);
-        dataset.read(reinterpret_cast<double*>(data.data));
-        //                             ^ ! Shouldn't be explicitly double of course TODO.
+        T<_Tp> data(dims[0], dims[1]);
+        dataset.read(reinterpret_cast<_Tp*>(data.data)); // idem v, choose
+        //dataset.read(data.template ptr<_Tp>()); // idem ^, choose
         return data;
     }
 };
 
+
+
+
 }  // namespace opencv
 
 // front-end
-template <class T>
-struct load_impl<cv::Mat_<T>>
+template <class _Tp>
+struct load_impl<cv::Mat_<_Tp>>
 {
-    static cv::Mat_<T> run(const File& file, const std::string& path)
+    static cv::Mat_<_Tp> run(const File& file, const std::string& path)
     {
-        return detail::opencv::load_impl<cv::Mat_<T>>::run(file, path);
+        return detail::opencv::load_impl<_Tp, cv::Mat_>::run(file, path);
     }
 };
 
--- a/include/highfive/H5Easy.hpp
+++ b/include/highfive/H5Easy.hpp
@@ -207,5 +207,6 @@ inline T load(const File& file, const std::string& path);
 #include "h5easy_bits/H5Easy_vector.hpp"
 #include "h5easy_bits/H5Easy_Eigen.hpp"
 #include "h5easy_bits/H5Easy_xtensor.hpp"
+#include "h5easy_bits/H5Easy_opencv.hpp"
 
 #endif  // H5EASY_HPP


include/highfive/h5easy_bits/H5Easy_opencv.hpp
```diff
@@ -0,0 +1,66 @@
...
+ *
+ */
+#ifndef H5EASY_BITS_OPENCV_HPP
+#define H5EASY_BITS_OPENCV_HPP
+
+#include "../H5Easy.hpp"
+#include "H5Easy_misc.hpp"
+#include "H5Easy_scalar.hpp"  // to get the basic "load_impl"
+
+////#ifdef H5_USE_OPENCV
```
tdegeus: This can go

include/highfive/h5easy_bits/H5Easy_opencv.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+// load opencv-object from DataSet
+template <class T>
+struct load_impl
+{
+    static T run(const File& file, const std::string& path)
+    {
+        DataSet dataset = file.getDataSet(path);
+        std::vector<size_t> dims = dataset.getDimensions();
+        T data(dims[0], dims[1]);
```
tdegeus: I'm a bit confused here... `T` is a `double`, `int`, ... is it not? Should you not rather construct on OpenCV object that owns its own memory? Then you can call on the next line `dataset.read(data.data());`

include/highfive/h5easy_bits/H5Easy_opencv.hpp
```diff
@@ -0,0 +1,66 @@
...
+        dataset.read(reinterpret_cast<double*>(data.data));
```
tdegeus: Obsolete because of previous comment, but you could have used `T`.

include/highfive/h5easy_bits/H5Easy_opencv.hpp
```diff
@@ -0,0 +1,66 @@
...
+        return detail::opencv::load_impl<cv::Mat_<T>>::run(file, path);
+    }
+};
+
+
+}  // namespace detail
+
+}  // namespace H5Easy
+
+////////#endif  // H5_USE_OPENCV
```
tdegeus: This can obviously also go.

spacescientist:
About CMake:
```cmake
# example
cmake_minimum_required(VERSION 3.0 FATAL_ERROR)
project(Prog)
set(CMAKE_CXX_STANDARD 11)

set(programme_exec_id executable)

find_package(OpenCV REQUIRED)

add_executable(${programme_exec_id}
	main.cpp
)

target_include_directories(
	${programme_exec_id}
	PUBLIC $<TARGET_PROPERTY:HighFive,INTERFACE_INCLUDE_DIRECTORIES>
	${OpenCV_INCLUDE_DIRS}
)

target_link_libraries(
	${programme_exec_id}
	HighFive
	${OpenCV_LIBS}
)

set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC -Wall -Wextra -pedantic")

set(CMAKE_BUILD_TYPE distribution)
set(CMAKE_CXX_FLAGS_DISTRIBUTION "-O2")
```

include/highfive/h5easy_bits/H5Easy_opencv.hpp
spacescientist: Sorry about that. In fact, my ugly-hack approach was probably *much* more naive than you think. 

It's been nothing more than looking at the code for xtensor, eigen and vector in h5easy_bits, and modifying bits here and there in ways that seemed promising. I didn't actually seriously studied how the internals work ('not at all' would be a fair statement). I just wanted to get a feel for it, and didn't expect it to converge. Somehow after pseudo-random trial and error, taking how this made the compiler complain as a feedback, the programme compiled and gave the expected result. To be honest, it felt a bit like the infinite monkey theorem — except that it took a surpringly small number of attempts, and only featured a single monkey. Well, and also that it's not quite Shakespeare either. That's also why I initially opened an issue rather than going for a pull request.

Now. It felt as if `T` was the matrix, which is why I eventually gave up and explicitly put `double` in the `reinterpret_cast<>` below. More precisely, replacing `double` by `T` in H5Easy_opencv.hpp gives the following error at compile time:
```
undefined reference to `HighFive::AtomicType<cv::Mat_<double> >::AtomicType()'
```

By the way, since you might be more familiar with `Eigen`, note that
```cpp
// if we have
Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> eigmat;
// and
cv::Mat_<T> cvmat;
```
an equivalent of `eigmat.data()` is `cvmat.ptr<T>()` or `reinterpret_cast<T*>(cvmat.data)`.



To answer your second question, it does seem to own its own memory if I'm not mistaken:

```cpp
#include <iostream>
#include <opencv2/core.hpp>
#include "highfive/H5Easy.hpp"

int main()
{
        cv::Mat_<double> test;
        {
        	H5Easy::File input_hdf("../in.hdf", H5Easy::File::ReadOnly);
        	test = H5Easy::load<cv::Mat_<double>>(input_hdf, "/data");
        }
        // expected output; no segfault:
        std::cout << test << std::endl;

        return 0;
}
```

tdegeus: Ah sorry, I misread. Indeed it seems that you allocate an opencv object that owns its own memory. In fact, why do you need the `dataset.read(reinterpret_cast<double*>(data.data));`. Why does `dataset.read(data.data());` or `dataset.read(data.begin());` not work?

spacescientist: The main issue I think is the fact that we technically have a T<_Tp>. If you're ok with the style of the following code, then it actually works with the exact same main.cpp file, and no longer explicitly involves `double`.
```cpp
// load opencv-object from DataSet
template<typename _Tp, template<typename> class T>
struct load_impl
{
    static T<_Tp> run(const File& file, const std::string& path)
    {
        DataSet dataset = file.getDataSet(path);
        std::vector<size_t> dims = dataset.getDimensions();
        T<_Tp> data(dims[0], dims[1]);
        dataset.read(reinterpret_cast<_Tp*>(data.data)); // Done.
        return data;
    }
};

}  // namespace opencv

// front-end
template <class _Tp>
struct load_impl<cv::Mat_<_Tp>>
{
    static cv::Mat_<_Tp> run(const File& file, const std::string& path)
    {
        return detail::opencv::load_impl<_Tp, cv::Mat_>::run(file, path);
    }
};
```
I believe it's much better, but please tell me what you think.

Note: Yesterday I did try to make things work with `dataset.read(data.data());` for quite some time, but I didn't know what to do with this error message, so I changed the approach:
```makefile
In file included from path/to/highfive/H5Easy.hpp:253,
                 from path/to/projectmain.cpp:5:
path/to/highfive/h5easy_bits/H5Easy_opencv.hpp: In instantiation of ‘static T H5Easy::detail::opencv::load_impl<T>::run(const HighFive::File&, const string&) [with T = cv::Mat_<double>; std::__cxx11::string = std::__cxx11::basic_string<char>]’:
path/to/highfive/h5easy_bits/H5Easy_opencv.hpp:88:59:   required from ‘static cv::Mat_<_Tp> H5Easy::detail::load_impl<cv::Mat_<_Tp> >::run(const HighFive::File&, const string&) [with T = double; std::__cxx11::string = std::__cxx11::basic_string<char>]’
path/to/highfive/h5easy_bits/H5Easy_scalar.hpp:166:37:   required from ‘T H5Easy::load(const HighFive::File&, const string&) [with T = cv::Mat_<double>; std::__cxx11::string = std::__cxx11::basic_string<char>]’
path/to/projectmain.cpp:12:81:   required from here
path/to/highfive/h5easy_bits/H5Easy_opencv.hpp:74:9: error: expression cannot be used as a function
         dataset.read(data.data());
```
While .begin() complains this way:
```makefile
path/to/highfive/h5easy_bits/H5Easy_opencv.hpp:76:15: error: invalid cast from type ‘cv::MatIterator_<double>’ to type ‘double*’
  dataset.read(reinterpret_cast<double*>(data.begin()));
```

spacescientist: Maybe even better if we use that, it appears that we can use:
```cpp
dataset.read(data.ptr()); // idem v
```
instead of 
```cpp
dataset.read(reinterpret_cast<_Tp*>(data.data)); // idem ^
```

tdegeus: Can you use `.data()` or `.begin` if you don't template? If so, we should just improve our template.

spacescientist: Sorry I was wrong with `data.ptr()`. Things still work with `reinterpret_cast`, though.
 
What I actually expected is that I would have to use `data.ptr<_Tp>()`, so I was surprised to get errors until I remove <_Tp>. Except that when I do remove `<_Tp>`, then it's wrong. 

There's should be a rather simple way to fix this but I don't know what yet. Here's the error message:
```makefile
path/to/library/highfive/h5easy_bits/H5Easy_opencv.hpp:117:34: error: expected primary-expression before ‘>’ token
         dataset.read(data.ptr<_Tp>());
                                  ^
path/to/library/highfive/h5easy_bits/H5Easy_opencv.hpp:117:36: error: expected primary-expression before ‘)’ token
         dataset.read(data.ptr<_Tp>());
                                    ^
```

spacescientist: And about your `data()` and `begin` question: these are the error messages are what I got *before* I went for `template<typename _Tp, template<typename> class T>`.

Sorry if that wasn't clear; I was happy with the template solution, so I put it first.

tdegeus: I think that you should stick to your `T` and then use `typename std::decay_t<T>::value_type` to get the inner type of the entries (`double` in your earlier example).

tdegeus:
Superseded by #343 . Thanks for the initial work @spacescientist

PR #297, opened 2020-02-24
Add support for enums

### LINKED ISSUE #110: Add support for Enums
It would be helpful to have first class support for Enums in HighFive. Currently I have to use a workaround like this to add them:

```c++
namespace HighFive
{
class UserType : public DataType
{
public:
    UserType(hid_t hid) { _hid = hid; }
};
}
```

```c++
enum Sorting
{
        none = 0,
        by_gid = 1,
        by_time = 2
};
```

```c++
auto enumValue = Sorting::none;
H5Tenum_insert(sortingEnum.getId(), "none", &enumValue);
enumValue = Sorting::by_gid;
H5Tenum_insert(sortingEnum.getId(), "by_gid", &enumValue);
enumValue = Sorting::by_time;
H5Tenum_insert(sortingEnum.getId(), "by_time", &enumValue);

auto sortingAttr =
            group.createAttribute(std::string("sorting"),
                                  HighFive::DataSpace::From(sortingEnum),
                                  sortingEnum);

H5Awrite(sortingAttr.getId(), sortingEnum.getId(), &enumValue);
```

alkino:
Fix #110

### DIFF
--- a/include/highfive/H5DataType.hpp
+++ b/include/highfive/H5DataType.hpp
@@ -168,12 +168,12 @@ private:
 /// };
 ///
 /// EnumType<Position> create_enum_position() {
-///     return EnumType<Position>({{"FIRST", Position::FIRST},
-///                                {"SECOND", Position::SECOND}});
+///     return {{"FIRST", Position::FIRST},
+///             {"SECOND", Position::SECOND}};
 /// }
 ///
 /// // You have to register the type inside HighFive
-/// HIGHFIVE_REGISTER_ENUM(Position, create_enum_position)
+/// HIGHFIVE_REGISTER_TYPE(Position, create_enum_position)
 ///
 /// void write_first(H5::File& file) {
 ///     auto dataset = file.createDataSet("/foo", Position::FIRST);
@@ -199,7 +199,6 @@ public:
         create();
     }
 
-    template<typename U>
     EnumType(std::initializer_list<member_def> t_members)
         : EnumType(std::vector<member_def>({t_members})) {}
 
--- a/include/highfive/H5DataType.hpp
+++ b/include/highfive/H5DataType.hpp
@@ -136,8 +136,8 @@ private:
     void create(size_t size = 0);
 };
 
-template <typename T>
-class EnumType : public DataType {
+template<typename T>
+class EnumType: public DataType {
 public:
     struct member_def {
         member_def(std::string t_name, T t_value)
@@ -149,20 +149,19 @@ public:
 
     EnumType(const EnumType& other) = default;
 
-    EnumType(std::vector<member_def> t_members, DataType t_underlying_type)
-        : members(std::move(t_members))
-        , underlying_type(t_underlying_type) {
+    EnumType(std::vector<member_def> t_members)
+        : members(std::move(t_members)) {
         create();
     }
 
-    EnumType(const std::initializer_list<member_def>& t_members, DataType t_underlying_type)
-        : EnumType(std::vector<member_def>({t_members}), t_underlying_type) {}
+    template<typename U>
+    EnumType(const std::initializer_list<member_def>& t_members)
+        : EnumType(std::vector<member_def>({t_members})) {}
 
      void commit(const Object& object, const std::string& name) const;
 
 private:
     std::vector<member_def> members;
-    DataType underlying_type;
 
     void create();
 };
--- a/include/highfive/H5DataType.hpp
+++ b/include/highfive/H5DataType.hpp
@@ -136,9 +136,37 @@ private:
     void create(size_t size = 0);
 };
 
+///
+/// \brief Create a enum HDF5 datatype
+///
+/// \code{.cpp}
+/// enum class Position {
+///     FIRST = 1,
+///     SECOND = 2,
+/// };
+///
+/// EnumType<Position> create_enum_position() {
+///     return EnumType<Position>({{"FIRST", Position::FIRST},
+///                                {"SECOND", Position::SECOND}});
+/// }
+///
+/// template<>
+/// DataType create_datatype<Position>() {
+///     return create_enum_position();
+/// }
+///
+/// void write_first(H5::File& file) {
+///     auto en = create_enum_position();
+///     en.commit(file, "Position");
+///     auto dataset = file.createDataset("/foo", DataSpace(1), en);
+///     dataset.write(Position::FIRST);
+/// }
+/// \endcode
 template<typename T>
 class EnumType: public DataType {
 public:
+    ///
+    /// \brief Use for defining a member of enum type
     struct member_def {
         member_def(std::string t_name, T t_value)
             : name(std::move(t_name))
@@ -158,7 +186,10 @@ public:
     EnumType(const std::initializer_list<member_def>& t_members)
         : EnumType(std::vector<member_def>({t_members})) {}
 
-     void commit(const Object& object, const std::string& name) const;
+    /// \brief Commit datatype into the given Object
+    /// \param object Location to commit object into
+    /// \param name Name to give the datatype
+    void commit(const Object& object, const std::string& name) const;
 
 private:
     std::vector<member_def> members;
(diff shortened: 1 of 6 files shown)


alkino:
Only first try, lot of things to fix

include/highfive/H5DataType.hpp
```diff
@@ -136,6 +136,37 @@ class CompoundType : public DataType {
     void create(size_t size = 0);
 };
 
+template <typename T>
+class EnumType : public DataType {
+public:
+    struct member_def {
```
matz-e:
```suggestion
    struct Member {
```
? Reads easier…

ohm314: yeah but it would break from the naming used for `CompoundType`

include/highfive/H5DataType.hpp
```diff
@@ -136,6 +136,37 @@ class CompoundType : public DataType {
...
+        member_def(std::string t_name, T t_value)
+            : name(std::move(t_name))
+            , value(std::move(t_value)) {}
+        std::string name;
+        T value;
+    };
+
+    EnumType(const EnumType& other) = default;
+
+    EnumType(std::vector<member_def> t_members, DataType t_underlying_type)
```
ohm314: I think this constructor doesn't make sense in the context of enum?

include/highfive/H5DataType.hpp
matz-e: fine

include/highfive/H5DataType.hpp
```diff
@@ -136,6 +136,37 @@ class CompoundType : public DataType {
     void create(size_t size = 0);
 };
 
+template <typename T>
```
ohm314: in this case `T` should be the `AtomicType<T>` which you can use both for the underlying type and the value type. Or maybe since you need to pass to the constructor the underlying type as an instantiated type, you can remove the template parameter here. Either way, the type T and the template parameter of the underlying type should be the same. This should be restricted.

include/highfive/H5DataType.hpp
```diff
@@ -136,6 +136,37 @@ class CompoundType : public DataType {
...
+
+    EnumType(const EnumType& other) = default;
+
+    EnumType(std::vector<member_def> t_members, DataType t_underlying_type)
+        : members(std::move(t_members))
+        , underlying_type(t_underlying_type) {
+        create();
+    }
+
+    EnumType(const std::initializer_list<member_def>& t_members, DataType t_underlying_type)
```
ohm314: can you use more specifically here `AtomicType<T>` instead of the more general `DataType` ?

include/highfive/H5DataType.hpp
alkino: It's only to be the same than compound type.
Any thoughts @ferdonline ?

include/highfive/H5DataType.hpp
alkino: Fixed

include/highfive/H5DataType.hpp
alkino: Fixed

include/highfive/H5DataType.hpp
```diff
@@ -136,6 +136,67 @@ class CompoundType : public DataType {
...
+        member_def(std::string t_name, T t_value)
+            : name(std::move(t_name))
+            , value(std::move(t_value)) {}
+        std::string name;
+        T value;
+    };
+
+    EnumType(const EnumType& other) = default;
+
+    EnumType(std::vector<member_def> t_members)
```
ferdonline: Idem here. Copy and move is slow

include/highfive/H5DataType.hpp
```diff
@@ -158,6 +158,67 @@ class CompoundType : public DataType {
...
+        : members(std::move(t_members)) {
+        create();
+    }
+
+    template<typename U>
+    EnumType(const std::initializer_list<member_def>& t_members)
```
ferdonline: I've recently read that initialiser lists (and btw, also iterators and function objs) are better passed by value: https://stackoverflow.com/questions/17803475/why-is-stdinitializer-list-often-passed-by-value. I think the difference is negligible though

include/highfive/H5DataType.hpp
```diff
@@ -158,6 +158,67 @@ class CompoundType : public DataType {
...
+///     dataset.write(Position::FIRST);
+/// }
+/// \endcode
+template<typename T>
+class EnumType: public DataType {
+public:
+    ///
+    /// \brief Use for defining a member of enum type
+    struct member_def {
+        member_def(std::string t_name, T t_value)
```
ferdonline: Prefer const-ref argument which gets copied. Cheaper than copy argument and then move. If you want to be faster implement also a move constructor `std::string&&` which you can move

include/highfive/H5DataType.hpp
```diff
@@ -158,6 +158,67 @@ class CompoundType : public DataType {
...
+///
+/// \brief Create a enum HDF5 datatype
+///
+/// \code{.cpp}
+/// enum class Position {
+///     FIRST = 1,
+///     SECOND = 2,
+/// };
+///
+/// EnumType<Position> create_enum_position() {
```
ferdonline: We have to explain better the code, unfortunately it's not so self-descriptive, specially the specialization.
And btw, I assume this specialization must be done within the HighFive namespace. I would stress that in the comments.
I would really try to reduce the amount of code, maybe drop the function, chain calls... e.g.:
`file.createDataset("/foo", DataSpace(1), en).write(Position::FIRST);`
Can we use the shortcut syntax?
`file.createDataset("/foo", en, Position::FIRST);`

tests/unit/tests_high_five_base.cpp
```diff
@@ -1319,6 +1319,96 @@ BOOST_AUTO_TEST_CASE(HighFiveCompounds) {
...
+namespace HighFive {
+    EnumType<Position> create_enum_position() {
+        return EnumType<Position>({{"FIRST", Position::FIRST},
+                                   {"SECOND", Position::SECOND},
+                                   {"THIRD", Position::THIRD},
+                                   {"LAST", Position::LAST}});
+    }
+
+    template<>
+    DataType create_datatype<Position>() {
```
ferdonline: I know this might not be possible, but it would be awesome if we didnt have to add specializations to the HighFive namespace... I'm afraid this is hard for the common user...
maybe let's think a bit more about this before the release...

include/highfive/H5DataType.hpp
ferdonline: Maybe I dont see your point @ohm314, to me enums members can have a very similar API than compound, except that there is a single type.

tests/unit/tests_high_five_base.cpp
alkino: I agree with you!

ferdonline: What about providing the user a macro?

alkino: ```CREATE_ENUM(POSITION, FIRST, SECOND)```

ferdonline: HIGHFIVE_REGISTER_TYPE (?)

alkino: Not possible due to comma inside your argument macro

ferdonline: Please update the example with the new syntax

tests/unit/tests_high_five_base.cpp
```diff
@@ -1319,6 +1319,96 @@ BOOST_AUTO_TEST_CASE(HighFiveCompounds) {
...
+};
+
+// This is only for boost test
+std::ostream& operator<<(std::ostream& ost, const Direction& dir) {
+    ost << static_cast<int>(dir);
+    return ost;
+}
+
+namespace HighFive {
+    EnumType<Position> create_enum_position() {
```
ferdonline: We have an implicit constructor, no?
```suggestion
    EnumType<Position> create_enum_position() {
     return  {{"FIRST", Position::FIRST},
              {"SECOND", Position::SECOND},
              {"THIRD", Position::THIRD},
              {"LAST", Position::LAST}});
```

PR #306, opened 2020-03-03
HDF5 object references

ohm314:
## This PR

Adds support for HDF5 [references](https://confluence.hdfgroup.org/display/HDF5/References). To keep things simple we support for now only object references (no region references, which are also not supported in parallel HDF5).
We introduce a new `Reference` class that represents an HDF5 object reference when writing to or reading from a file.

## Rationale

Unlike soft and hard links references can be stored and read as data in HDF5 files as Datasets. They can be used to create more complex data relationships beyond vectors and matrices.

## Use

Here is a minimal example to create and read an HDF5 reference with the new `Reference` class:

```c++
// write references
{
    File file("example.h5", File::ReadWrite | File::Create | File::Truncate);
    // Create some data and write it into a DataSet
    auto data = std::vector<double>({1.0, 2.0});
    DataSet dataset = file.createDataSet("dset", data);

    // Create a Reference vector with one element and write it into a new DataSet
    auto references = std::vector<Reference>({{file, dataset}});
    DataSet ref_ds = file.createDataSet("rds", references);
}
// read references
{
    File file("example.h5", File::ReadOnly);
    // get the reference dataset
    DataSet ref_ds = file.getDataSet("rds");
    // create a vector to read the references into
    auto refs = std::vector<Reference>();
    // read the references
    ref_ds.read(refs);
    auto data_ds = refs[0].dereference<DataSet>(file);
    std::vector<double> rdata;
    data_ds.read(rdata);
}
```

### DIFF
--- a/include/highfive/bits/H5Reference_misc.hpp
+++ b/include/highfive/bits/H5Reference_misc.hpp
@@ -34,21 +34,38 @@ inline void Reference::create_ref(hobj_ref_t* refptr) const {
 
 inline ObjectType Reference::getType(const Object& location) const {
     hid_t res;
+#if (H5Rdereference_vers == 2)
+    if ((res = H5Rdereference(location.getId(), H5P_DEFAULT, H5R_OBJECT, &href)) < 0) {
+        HDF5ErrMapper::ToException<ReferenceException>("Unable to dereference.");
+    }
+#else
     if ((res = H5Rdereference(location.getId(), H5R_OBJECT, &href)) < 0) {
         HDF5ErrMapper::ToException<ReferenceException>("Unable to dereference.");
     }
+#endif
     return Object(res).getType();
 }
 
 template <typename T>
 inline T Reference::dereference(const Object& location) const {
-    static_assert(std::is_base_of<Object, T>::value != 0,
-                  "Can only return a subtype of HighFive::Object");
+    static_assert(std::is_same<DataSet, T>::value || std::is_same<Group, T>::value,
+                  "We can only (de)reference HighFive::Group or HighFive:DataSet");
     hid_t res;
+#if (H5Rdereference_vers == 2)
+    if ((res = H5Rdereference(location.getId(), H5P_DEFAULT, H5R_OBJECT, &href)) < 0) {
+        HDF5ErrMapper::ToException<ReferenceException>("Unable to dereference.");
+    }
+#else
     if ((res = H5Rdereference(location.getId(), H5R_OBJECT, &href)) < 0) {
         HDF5ErrMapper::ToException<ReferenceException>("Unable to dereference.");
     }
-    return T(Object(res));
+#endif
+    Object obj(res);
+    if (obj.getType() != T::type) {
+        HDF5ErrMapper::ToException<ReferenceException>(
+            "Trying to dereference the wrong type");
+    }
+    return T(obj);
 }
 
 }  // namespace HighFive
--- a/include/highfive/bits/H5Reference_misc.hpp
+++ b/include/highfive/bits/H5Reference_misc.hpp
@@ -33,23 +33,22 @@ inline void Reference::create_ref(hobj_ref_t* refptr) const {
 }
 
 inline ObjectType Reference::getType(const Object& location) const {
-    hid_t res;
-#if (H5Rdereference_vers == 2)
-    if ((res = H5Rdereference(location.getId(), H5P_DEFAULT, H5R_OBJECT, &href)) < 0) {
-        HDF5ErrMapper::ToException<ReferenceException>("Unable to dereference.");
-    }
-#else
-    if ((res = H5Rdereference(location.getId(), H5R_OBJECT, &href)) < 0) {
-        HDF5ErrMapper::ToException<ReferenceException>("Unable to dereference.");
-    }
-#endif
-    return Object(res).getType();
+    return get_ref(location).getType();
 }
 
 template <typename T>
 inline T Reference::dereference(const Object& location) const {
     static_assert(std::is_same<DataSet, T>::value || std::is_same<Group, T>::value,
                   "We can only (de)reference HighFive::Group or HighFive:DataSet");
+    auto obj = get_ref(location) ;
+    if (obj.getType() != T::type) {
+        HDF5ErrMapper::ToException<ReferenceException>(
+            "Trying to dereference the wrong type");
+    }
+    return T(obj);
+}
+
+Object Reference::get_ref(const Object& location) const {
     hid_t res;
 #if (H5Rdereference_vers == 2)
     if ((res = H5Rdereference(location.getId(), H5P_DEFAULT, H5R_OBJECT, &href)) < 0) {
@@ -60,12 +59,7 @@ inline T Reference::dereference(const Object& location) const {
         HDF5ErrMapper::ToException<ReferenceException>("Unable to dereference.");
     }
 #endif
-    Object obj(res);
-    if (obj.getType() != T::type) {
-        HDF5ErrMapper::ToException<ReferenceException>(
-            "Trying to dereference the wrong type");
-    }
-    return T(obj);
+    return Object(res);
 }
 
 }  // namespace HighFive
--- a/include/highfive/bits/H5Reference_misc.hpp
+++ b/include/highfive/bits/H5Reference_misc.hpp
@@ -1,0 +1,87 @@
+/*
+ *  Copyright (c), 2020, EPFL - Blue Brain Project
+ *
+ *  Distributed under the Boost Software License, Version 1.0.
+ *    (See accompanying file LICENSE_1_0.txt or copy at
+ *          http://www.boost.org/LICENSE_1_0.txt)
+ *
+ */
+
+#ifndef H5REFERENCE_MISC_HPP
+#define H5REFERENCE_MISC_HPP
+
+#include <string>
+
+namespace HighFive {
+
+inline Reference::Reference(const Object& parent, const Object& o)
+    : parent_id(parent.getId()) {
+
+    const size_t maxLength = 255;
+    char buffer[maxLength + 1];
+    if (H5Iget_name(o.getId(), buffer, maxLength) <= 0) {
+        HDF5ErrMapper::ToException<DataTypeException>("Invalid object or location");
+    }
+    obj_name = std::string(buffer);
+}
+
+
+void Reference::create_ref(hobj_ref_t* refptr) const {
+
+    H5Rcreate(refptr, parent_id, obj_name.c_str(), H5R_OBJECT, -1);
+}
+
+template <typename T>
+T Reference::dereference(const Object& loc) {
+    static_assert(std::is_base_of<Object, T>::value != 0,
+        "Can only return a subtype of HighFive::Object");
+    hid_t res = H5Rdereference(loc.getId(), H5R_OBJECT, &href);
+    auto obj = Object(res);
+    T h5_obj(obj);
+    return h5_obj;
+}
+
+namespace details {
+
+// apply conversion for vectors nested vectors
+template <>
+struct data_converter<std::vector<Reference>> {
+    inline data_converter(const std::vector<Reference>&, const DataSpace& space)
+        : _dims(space.getDimensions()) {
+        if (!is_1D(_dims)) {
+            throw DataSpaceException("Only 1D std::array supported currently.");
+        }
+    }
+
+    inline typename type_of_array<hobj_ref_t>::type*
+    transform_read(std::vector<Reference>&) {
+        _vec_align.resize(compute_total_size(_dims));
+        return _vec_align.data();
+    }
+
+    inline typename type_of_array<hobj_ref_t>::type* transform_write(const std::vector<Reference>& vec) {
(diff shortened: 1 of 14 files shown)


tests/unit/tests_high_five_base.cpp
```diff
@@ -1531,6 +1532,83 @@ BOOST_AUTO_TEST_CASE(HighFiveFixedLenStringArrayStructure) {
...
+        std::vector<double> rdata;
+        data_ds.read(rdata);
+        for (size_t i = 0; i < rdata.size(); ++i) {
+            BOOST_CHECK_EQUAL(rdata[i], vec1[i]);
+        }
+
+        auto group = refs[1].dereference<Group>(file);
+        DataSet data_ds2 = group.getDataSet(DATASET2_NAME);
+        std::vector<double> rdata2;
+        data_ds.read(rdata2);
```
matz-e: Shouldn't this be `data_ds2`?

I'd be tempted to say that each part of this test should be in its own scope.

include/highfive/bits/H5Reference_misc.hpp
```diff
@@ -0,0 +1,99 @@
...
+    }
+    auto obj = Object(res);
+    T h5_obj(obj);
+    return h5_obj;
+}
+
+namespace details {
+
+template <>
+struct data_converter<std::vector<Reference>> {
```
ferdonline: Hopefully by inheriting from `template<Container> data_converter` you can drop a lot of code here. (rebased in the latest master?)

tests/unit/tests_high_five_base.cpp
```diff
@@ -1531,6 +1532,83 @@ BOOST_AUTO_TEST_CASE(HighFiveFixedLenStringArrayStructure) {
...
+        dataset2.write(vec2);
+
+        // create group and dataset to hold reference
+        Group refgroup = file.createGroup(REFGROUP_NAME);
+
+        // Create the data space for the dataset.
+        std::vector<size_t> ref_dims{2};
+        DataSpace ref_dspace(ref_dims);
+
+        DataSet ref_ds = refgroup.createDataSet(REFDATASET_NAME, ref_dspace, AtomicType<Reference>());
```
ferdonline: IMO the standard way to create an empty dataset is: (which is probably supported)
`createDataSet<Reference>(REFDATASET_NAME, ref_dspace);`

tests/unit/tests_high_five_base.cpp
```diff
@@ -1531,6 +1532,83 @@ BOOST_AUTO_TEST_CASE(HighFiveFixedLenStringArrayStructure) {
...
+        File file(FILE_NAME, File::ReadWrite | File::Create | File::Truncate);
+
+        // create group
+        Group g1 = file.createGroup(GROUP_NAME);
+
+        // Create the data space for the dataset.
+        std::vector<size_t> dims{4};
+        DataSpace dataspace(dims);
+
+        DataSet dataset1 = g1.createDataSet(
```
ferdonline: maybe simplify the example with the shortcut syntax (without manual dataspaces)
```suggestion
        DataSet dataset1 = g1.createDataSet(DATASET1_NAME, vec1);
```

tests/unit/tests_high_five_base.cpp
```diff
@@ -1531,6 +1532,83 @@ BOOST_AUTO_TEST_CASE(HighFiveFixedLenStringArrayStructure) {
...
+        auto references = std::vector<Reference>({{g1, dataset1},{file, g1}});
+
+        ref_ds.write(references);
+    }
+    // read it back
+    {
+        File file(FILE_NAME, File::ReadOnly);
+        Group refgroup = file.getGroup(REFGROUP_NAME);
+
+        DataSet refdataset = refgroup.getDataSet(
```
ferdonline: We have a clang-format (sorry, short line length hurt my eyes :p)

tests/unit/tests_high_five_base.cpp
```diff
@@ -1531,6 +1532,83 @@ BOOST_AUTO_TEST_CASE(HighFiveFixedLenStringArrayStructure) {
...
+
+        // Create the data space for the dataset.
+        std::vector<size_t> ref_dims{2};
+        DataSpace ref_dspace(ref_dims);
+
+        DataSet ref_ds = refgroup.createDataSet(REFDATASET_NAME, ref_dspace, AtomicType<Reference>());
+
+        auto references = std::vector<Reference>({{g1, dataset1},{file, g1}});
+
+        ref_ds.write(references);
```
ferdonline: Actually, we should be able to use the shortcut syntax. Add a test to prove so please.
`refgroup.createDataSet(REFDATASET_NAME, references)`

include/highfive/bits/H5Reference_misc.hpp
```diff
@@ -0,0 +1,99 @@
...
+    }
+    return Object(res).getType();
+}
+
+template <typename T>
+T Reference::dereference(const Object& location) const {
+    static_assert(std::is_base_of<Object, T>::value != 0,
+        "Can only return a subtype of HighFive::Object");
+    hid_t res;
+    if ((res = H5Rdereference(location.getId(), H5R_OBJECT, &href)) < 0) {
```
ferdonline: Maybe we should add an extra check, whether the reference target type is T

tests/unit/tests_high_five_base.cpp
```diff
@@ -1531,6 +1532,83 @@ BOOST_AUTO_TEST_CASE(HighFiveFixedLenStringArrayStructure) {
...
+        }
+
+        auto group = refs[1].dereference<Group>(file);
+        DataSet data_ds2 = group.getDataSet(DATASET2_NAME);
+        std::vector<double> rdata2;
+        data_ds2.read(rdata2);
+        for (size_t i = 0; i < rdata2.size(); ++i) {
+            BOOST_CHECK_EQUAL(rdata2[i], vec2[i]);
+        }
+    }
```
ferdonline: And add here if a CHECK if assertion is raised in case we ask to dereference wrong type (group <-> dataset)

tests/unit/tests_high_five_base.cpp
```diff
@@ -1531,6 +1532,83 @@ BOOST_AUTO_TEST_CASE(HighFiveFixedLenStringArrayStructure) {
...
+        auto refs = std::vector<Reference>();
+        refdataset.read(refs);
+        auto data_ds = refs[0].dereference<DataSet>(file);
+        std::vector<double> rdata;
+        data_ds.read(rdata);
+        for (size_t i = 0; i < rdata.size(); ++i) {
+            BOOST_CHECK_EQUAL(rdata[i], vec1[i]);
+        }
+
+        auto group = refs[1].dereference<Group>(file);
```
ferdonline: If not too complex, file should default to be the same where the the reference dataset is on.

include/highfive/bits/H5Reference_misc.hpp
```diff
@@ -0,0 +1,99 @@
...
+        HDF5ErrMapper::ToException<ReferenceException>("Unable to dereference.");
+    }
+    return Object(res).getType();
+}
+
+template <typename T>
+T Reference::dereference(const Object& location) const {
+    static_assert(std::is_base_of<Object, T>::value != 0,
```
ferdonline: Isn't this too broad? e.g. property lists are objects, but can we can have references?

include/highfive/bits/H5Reference_misc.hpp
```diff
@@ -0,0 +1,99 @@
...
+        "Can only return a subtype of HighFive::Object");
+    hid_t res;
+    if ((res = H5Rdereference(location.getId(), H5R_OBJECT, &href)) < 0) {
+        HDF5ErrMapper::ToException<ReferenceException>("Unable to dereference.");
+    }
+    auto obj = Object(res);
+    T h5_obj(obj);
+    return h5_obj;
```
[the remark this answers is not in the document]
ohm314: Well I can do `T(Object(res))` because I can only do explicit conversion (which is also for the better I suppose?)

tests/unit/tests_high_five_base.cpp
ohm314: these are two different references, though...? I can pass any object as location, but file is here the most appropriate location.

include/highfive/bits/H5Reference_misc.hpp
ohm314: I was wondering if it's worth to do the extra overhead, but I can add it.

include/highfive/bits/H5Reference_misc.hpp
ohm314: It is too broad, you are correct. I can do an assert to check that T is `DataSet` or `Group` I just find this so ugly. maybe I can find a nicer way to do this?

alkino: `static_assert` is here for that

tests/unit/tests_high_five_base.cpp
ohm314: HDF5 1.12 introduces [H5R_OPEN_OBJECT](https://portal.hdfgroup.org/display/HDF5/H5R_OPEN_OBJECT), which is exactly what you want, but without that you sort of need a file. Can we defer this for now?

include/highfive/bits/H5Reference_misc.hpp
```diff
@@ -34,21 +34,38 @@ inline void Reference::create_ref(hobj_ref_t* refptr) const {
...
+    if ((res = H5Rdereference(location.getId(), H5P_DEFAULT, H5R_OBJECT, &href)) < 0) {
+        HDF5ErrMapper::ToException<ReferenceException>("Unable to dereference.");
+    }
+#else
     if ((res = H5Rdereference(location.getId(), H5R_OBJECT, &href)) < 0) {
         HDF5ErrMapper::ToException<ReferenceException>("Unable to dereference.");
     }
-    return T(Object(res));
+#endif
+    Object obj(res);
```
ferdonline: would it be too hard to enable T(res)? I guess we would need to make Group and Dataset friendly with us, but it seems alright with me and we avoid this double conversion

include/highfive/bits/H5Reference_misc.hpp
```diff
@@ -0,0 +1,73 @@
...
+inline void Reference::create_ref(hobj_ref_t* refptr) const {
+    if (H5Rcreate(refptr, parent_id, obj_name.c_str(), H5R_OBJECT, -1) < 0) {
+        HDF5ErrMapper::ToException<ReferenceException>(
+            std::string("Unable to create the reference for \"") + obj_name + "\":");
+    }
+}
+
+inline ObjectType Reference::getType(const Object& location) const {
+    hid_t res;
+#if (H5Rdereference_vers == 2)
```
ferdonline: This and next functions share a lot of code, it would be nice to dedup a bit

tests/unit/tests_high_five_base.cpp
```diff
@@ -1531,20 +1503,90 @@ BOOST_AUTO_TEST_CASE(HighFiveFixedLenStringArrayStructure) {
...
+
+        // Create the data space for the dataset.
+        std::vector<size_t> dims{4};
+        DataSpace dataspace(dims);
+
+        DataSet dataset1 = g1.createDataSet(DATASET1_NAME, vec1);
+        DataSet dataset2 = g1.createDataSet(DATASET2_NAME, vec2);
+
+        // write some data
+        dataset1.write(vec1);
```
ferdonline: data has been already written. this is duplicating the action

tests/unit/tests_high_five_base.cpp
```diff
@@ -1531,20 +1503,90 @@ BOOST_AUTO_TEST_CASE(HighFiveFixedLenStringArrayStructure) {
...
+
+        // Create the data space for the dataset.
+        std::vector<size_t> ref_dims{2};
+        DataSpace ref_dspace(ref_dims);
+
+        DataSet ref_ds = refgroup.createDataSet<Reference>(REFDATASET_NAME, ref_dspace);
+
+        auto references = std::vector<Reference>({{g1, dataset1}, {file, g1}});
+
+        ref_ds.write(references);
```
ferdonline:
```suggestion
        auto references = std::vector<Reference>({{g1, dataset1}, {file, g1}});
        refgroup.createDataSet(REFDATASET_NAME, references);
```

ferdonline:
Omar this is great to see this addition to HighFive and the very clean API. I'm very pleased as well with the neat implementation! I made just a couple of final picky remarks

tests/unit/tests_high_five_base.cpp
ferdonline: Sure! Either way if it was there in the original API it shouldnt be a surprise for users.

include/highfive/bits/H5Reference_misc.hpp
ohm314: I still need to construct first an object to get the type, so it seems redundant to me? otherwise I would agree with you that I've been doing this a bit roundabout.. maybe in this case it's ok (taking into account my dedup changes :) )

alkino:
Impressive work, thanks for that!

ohm314:
Thank you @ferdonline and @alkino for your reviews and help!

Issue #224, opened 2019-10-16
Exception Triggered


nyckmaia:
I did a simple "copy and paste" test with Qt 5 and HighFive library and it works good!
I tested the H5Easy C++ example...

So, now I would like to use this library in my Qt C++ project.

In the middle of my code, I typed:
```
H5Easy::File file("example.h5", H5Easy::File::Overwrite);
std::vector<double> B  = { 1., 2., 3. };
H5Easy::dump(file, "/path/to/B", B, H5Easy::DumpMode::Overwrite);
```
**All compiles Ok...no errors.**

It creates the `example.h5` file... it creates the `B` vector...all ok.
But when the debugger hits the `H5::dump()` **I got this error message:**

![error](https://user-images.githubusercontent.com/4555404/66926270-1ab33b80-f004-11e9-8c55-cf101c309f09.jpg)

The Qt console points to this file:
```
highfive\bits\H5Exception_misc.hpp:63: 
warning: Debugger encountered an exception: Exception at 0x7ffc2c9e9129, code: 0xe06d7363: C++ exception, flags=0x1 (execution cannot be continued) (first chance)
```

**My questions:**
1- What can I do **to solve** this exception?
2- Is there a way to know if the H5::dump() **wrote all data with no errors?** Like a `return bool`?

**My system is:**

- OS: Windows 10 x64
- Compiler: MSVC2019 64bit
- IDE: Qt Creator IDE 4.10

ferdonline:
Hi @nyckmaia 
If you start it from a command line, do you see any additional error messages?

alkino:
In fact it already return the created `DataSet` or `throw` on error.
Do you need more informations?

nyckmaia:
I'm trying to catch the throw exception:
```
try {
    // Write density vector to a HDF5 file:
    H5Easy::dump(file, "/my/path/to/density", _density, H5Easy::DumpMode::Overwrite);
} catch (HighFive::Exception e) {
    // Print error on console:
    qDebug() << e.what();
}
```
But when the debugger hits the `H5Easy::dump()` I got the expection GUI window like my first post here (image above).

Questions:
1 - So, how can I **hold the exception using HighFive?**
2- Could you please give a **short example** about the return of `H5Easy::dump()`?
I tried something like:

```
H5Easy::DataSet = H5Easy::dump(file, "/my/path/to/density", _density, H5Easy::DumpMode::Overwrite);
```
But I got a compilation error. **Could you help me?**

[...]

PR #308, opened 2020-03-04
Dont rely on exception in node.exist

### LINKED ISSUE #289: MSVC: Exception Trigger
One last mention about this #224 bug:

I know that MSVC compiler can be picky sometimes, but here is just a little help about how/where can you find this bug, ok?

When I run my application in:
 * Release mode: I got **no** exceptions at all.
 * Debug mode **without breakpoint** in any HighFive API lines: I got **no** exceptions too. All working good!

The only case that I got this exception above is when I ran my application in debug mode and **set some breakpoint** before same HighFive API command and then **I hit F10** to execute the app line by line.

In this case I got this exception below:

image

Can be here executing this:
```
bool exists = outputFile.exist(group_path);
```

And can be here too:
```
H5Easy::DataSet ds = H5Easy::dump(outputFile, full_path, my_vector, H5Easy::DumpMode::Overwrite);
```
Obs.: The exactly same source-code **runs perfect in Linux (g++).**

**My Setup:**
 * Windows 10 x64
 * Qt Creator IDE 5.11.1
 * Qt version 5.14.1
 * Compiler: MSVC 2019

ferdonline:
The implementation of `Node.exist()` required catching an exception that we generated ourselves. Despite being a simple principle, it kind of used exceptions for flow control and In some compiler toolchains (QT) this would create unwanted error messages (#224, #289)

## This PR
Makes the low-level `_exist` function conditionally treat the error as a normal false, not throwing exception. This design even simplified the implementation of the public `exist()` method.

Fixes #289

### DIFF
--- a/include/highfive/bits/H5Node_traits_misc.hpp
+++ b/include/highfive/bits/H5Node_traits_misc.hpp
@@ -181,11 +181,12 @@ inline std::vector<std::string> NodeTraits<Derivate>::listObjectNames() const {
 }
 
 template <typename Derivate>
-inline bool NodeTraits<Derivate>::_exist(const std::string& node_name) const {
-    htri_t val = H5Lexists(static_cast<const Derivate*>(this)->getId(),
-                           node_name.c_str(), H5P_DEFAULT);
-
-    if (val < 0) {
+inline bool NodeTraits<Derivate>::_exist(const std::string& node_name,
+                                         bool raise_errors) const {
+    SilenceHDF5 silencer{raise_errors};
+    const auto val = H5Lexists(static_cast<const Derivate*>(this)->getId(),
+                               node_name.c_str(), H5P_DEFAULT);
+    if (raise_errors && val < 0) {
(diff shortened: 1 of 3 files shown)


include/highfive/bits/H5Node_traits_misc.hpp
```diff
@@ -205,12 +206,8 @@ inline bool NodeTraits<Derivate>::exist(const std::string& group_path) const {
...
     if (group_path.find('/') != std::string::npos) {
         _exist("/");  // Shall not throw under normal circumstances
-        try {
-            SilenceHDF5 silencer;
-            return _exist(group_path);
-        } catch (const GroupException&) {
-            return false;
-        }
+        // Unless "/" (already checked), verify path exists (not thowing errors)
+        return (group_path == "/")? true : _exist(group_path, false);
```
alexsavulescu: White space between ")" and "?" :)

ferdonline:
@alexsavulescu sorry, github always dimisses even on small changes. I dont intend to change more

PR #311, opened 2020-03-04
update Object::getInfo to work with hdf5 1.12.0 as well

### LINKED ISSUE #309: Update HighFive to run with HDF5 1.12.0
We just updated to the latest HDF5 1.12.0 release which broke some things in HIghFive. 

I will look into it and create a pull request if I manage to get it running.

Mightrider:
### This PR
Fixes HighFive with the latest release of HDF5 1.12.0. 
It leverages hdf5 compat API to keep supporting existing functions and marks them  **Deprecated**. In the next major release we can remove this API and only use the new hdf5 structure/calls.
The case here was getAddress() which relies on a H5O_info_t and H5Oget_info, both had their API changed.

Fixes #309


ferdonline:
Great contribution! Nevertheless dropping the old API completely without first deprecating it might be too harsh.
Since, for some time, people won't use the new function, we could keep compatibility (using the old function renamed to `H5Oget_info1()`) and mark getAddress deprecated()

mgeplf:
> Nevertheless dropping the old API completely without first deprecating it might be too harsh.

Is there a way to emulate the old API with the new `HDF5 1.12.0`?  Brew has already moved to it, so `HighFive` won't compile.  This is currently preventing me from deploying `libsonata`.

emmenlau:
We build hdf5 with `-DHDF5_ENABLE_DEPRECATED_SYMBOLS=ON` which makes HighFive mostly work for us. The only other required change is included in this PR. We did not need to change the default API version AFAIK, at least not for hdf5-1.12.0.

Can you try to add `-DHDF5_ENABLE_DEPRECATED_SYMBOLS=ON` to the brew build?

mgeplf:
> Can you try to add -DHDF5_ENABLE_DEPRECATED_SYMBOLS=ON to the brew build?

You mean in the official brew repo?

If not, I can probably build it myself w/ the `DEPRECATED_SYMBOLS`, but that partly defeats the point of having a package manager do it for me.

For reference, it's also hit the `spack` repos:
https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/hdf5/package.py#L26

emmenlau:
> > Can you try to add -DHDF5_ENABLE_DEPRECATED_SYMBOLS=ON to the brew build?
> 
> You mean in the official brew repo?
> 
> If not, I can probably build it myself w/ the `DEPRECATED_SYMBOLS`, but that partly defeats the point of having a package manager do it for me.

At first I would in any case suggest to try it out locally, to see if it even solves the issue for you. You may still need further changes in HighFive, depending on your use case. Basically, what I'm saying is that hdf5-1.12 is not yet well tested or integrated.

In the long run, I see two options:
  * Update HighFive to not require deprecated symbols anymore, and/or
  * Kindly ask brew to change their default build.

I'm not a HighFive maintainer, but I assume that possibility (1) can take quite some time unless someone volunteers to do it (or it becomes a high priority for the core team).

mgeplf:
I can work around the problem, possibly even for deployment.  However, `libsonata` is one of `BlueBrain`'s software distributions, and I'm not sure everyone will be able to work around it easily.

> In the long run, I see two options:
> Update HighFive to not require deprecated symbols anymore, and/or
> Kindly ask brew to change their default build.

I'm not sure how viable the second choice is, especially since other package managers will have to implement include the same build flags (and already there is a lot of fragmentation in the HDF5 world on that).

Likely the first choice will have to be made, which circles back to the original question: Is it possible to emulate the old `HighFive` API with the HDF5 1.12.0? If the answer is no, I'm not sure what can be done to maintain backwards compatibility, and thus the deprecation will have to happen for anyone using the newer HDF5.  Another option might be to put the `Object::getInfo()` behind a flag or ifdef - so if you need it, you can enable it, but then you won't be able to use the newer 1.12.

emmenlau:
> > In the long run, I see two options:
> > Update HighFive to not require deprecated symbols anymore, and/or
> > Kindly ask brew to change their default build.
> 
> I'm not sure how viable the second choice is, especially since other package managers will have to implement include the same build flags (and already there is a lot of fragmentation in the HDF5 world on that).
> 
> Likely the first choice will have to be made, which circles back to the original question: Is it possible to emulate the old `HighFive` API with the HDF5 1.12.0? If the answer is no, I'm not sure what can be done to maintain backwards compatibility, and thus the deprecation will have to happen for anyone using the newer HDF5. Another option might be to put the `Object::getInfo()` behind a flag or ifdef - so if you need it, you can enable it, but then you won't be able to use the newer 1.12.

I think we're just coming from two completely different sides. I'm not an HighFive maintainer and this PR is not about BlueBrain's stance on hdf5-1.12. This PR is solely the effort to improve hdf5-1.12 compatibility. I'd be happy if you can test it and report back if it works for you! And I'm also happy if it brings you a step closer to your goals. But other concerns should be raised in other issues, would that make sense? So while I can very much relate to your sentiments, I think its not viable to discuss them in this PR.

mgeplf:
> I'd be happy if you can test it and report back if it works for you! 

It does.  However, I don't use the `Object::getInfo()` or the new `getToken`

My understanding is that it's not being merged because 'Nevertheless dropping the old API completely without first deprecating it might be too harsh.', so, I asked whether it's feasible to maintain backwards compat with the newer version of hdf5 - hoping to get feedback on what the way forward is here.  This is why I think it's important to figure it out in this PR, because it directly impacts whether this PR can/will be merged, or whether it needs work - as suggested by the 'too harsh' comment.

emmenlau:
Maybe I'm confused now, but I'm under the impression that this is mixing two things. The comment about `dropping the old API` does in my eyes not refer to deprecated symbols in hdf5. Its about the changed return type of `Object::getInfo()`. The two things are unrelated:
  * HighFive is currently coded against an older API of hdf5, thats why you need `HDF5_ENABLE_DEPRECATED_SYMBOLS=ON` in hdf5. This PR here does not change that.
  * hdf5 1.12 does no longer provide `raw_info.addr` for `Object::getInfo()`, independent of enabling the deprecated symbols or not.
Therefore to get *any* chance to use hdf5-1.12 with HighFive, you need this PR. But this PR does not fix the requirement for deprecated symbols, and its not planned or wanted to include the fix for deprecated symbols in this PR.

Was there some confusion or am I now on the wrong track?

mgeplf:
> Therefore to get any chance to use hdf5-1.12 with HighFive, you need this PR. But this PR does not fix the requirement for deprecated symbols, and its not planned or wanted to include the fix for deprecated symbols in this PR.

That's basically the essence of my question: for 1.12 compat, we need this patch, or something like it.  The question is whether the we need to emulate the behavior of HighFive's `Object::getInfo()`, or whether we're forced to deprecate this no matter what.

Anyways, I'll go downstairs and ask some of the maintainers.

emmenlau:
> > Therefore to get any chance to use hdf5-1.12 with HighFive, you need this PR. But this PR does not fix the requirement for deprecated symbols, and its not planned or wanted to include the fix for deprecated symbols in this PR.
> 
> That's basically the essence of my question: for 1.12 compat, we need this patch, or something like it. The question is whether the we need to emulate the behavior of HighFive's `Object::getInfo()`, or whether we're forced to deprecate this no matter what.

I may be confused, but I'm under a different impression. Could you help clarify if the major part of your questions where revolving around deprecated symbols or not?

mgeplf:
Ok, I went and talked to @ferdonline.  The ideal solution is having `ObjectInfo::getAddress()` continue to exist, and return `haddr_t`, with all versions of `hdf5`.  Now it's a question of checking that this is possible w/ the new hdf5 API, and implementing it.

emmenlau:
> I may be confused, but I'm under a different impression. Could you help clarify if the major part of your questions where revolving around deprecated symbols or not?

So is your question about deprecated symbols now resolved? Because I'd like to close that part of the discussion and shift the focus back on the `haddr_t`.

mgeplf:
> So is your question about deprecated symbols now resolved?
Yeah -> we'll hopefully be able to implement the ideal solution.

emmenlau:
> Yeah -> we'll hopefully be able to implement the ideal solution.

Ok great! Let me know if we can contribute something to preserving the API of `haddr_t` with hdf5-1.12.

We currently did not use the token that is shipped with hdf5-1.12, so I can't say in what way its better or worse. What we did find is that the token contains a lot more information, which may be relevant for the new virtual layer in hdf5-1.12.

I'm very curious what you find out!

ferdonline:
I quickly put up a fix. However I cant push to this repo

emmenlau:
I've added you with write access. Can you try again?

ferdonline:
Yep it works, thanks! :) 
Could I edit the PR description?

emmenlau:
You're a maintainer of this fork now, does that enable you to edit the PR? I've not worked with Github permissions before, let me know if you have a better understanding of whats required.

And please feel free to edit the PR to your liking!

ferdonline:
@mgeplf I tested with 1.12 from brew and it works... probably we can update the CI to use it too

include/highfive/H5Object.hpp
```diff
@@ -104,7 +91,7 @@ class Object {
 class ObjectInfo  {
   public:
     /// \brief Retrieve the address of the object (within its file)
-    haddr_t getAddress() const noexcept;
+    H5_DEPRECATED haddr_t getAddress() const noexcept;
```
mgeplf: Could this be enhanced with what is replacing it?  ie: `H5_DEPRECATED("Return type will change to size_t")` or whatever the plan is?

mgeplf:
@ferdonline 
> I tested with 1.12 from brew and it works... probably we can update the CI to use it too

Yeah, I'd add that to the CI if possible.  Can we also add a test that exercises the `getAddress()` functionality while we're here?

ferdonline:
I added CI and deprecation messages and noticed they are visible during compilation.
I would like to provide some alternative API but I'm lacking bandwidth. We can't just return the raw Token. Not encapsulating is bad design and the user cant do much with it.
For the next release we should definitely provide some support.

ferdonline:
@ohm314 could you please reevaluate this PR? My lant change is regarding CI so we have tests with libhdf5 1.12

.travis.yml
```diff
@@ -109,10 +117,16 @@ env:
...
+  - if [[ "$BREW_AUTOUPDATE" && "$BREW_AUTOUPDATE" != 0 ]]; then
+      unset HOMEBREW_NO_AUTO_UPDATE;
+    fi
+
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
       MINCONDA_OS=$MINCONDA_LINUX;
     elif [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
-      HOMEBREW_NO_AUTO_UPDATE=1 brew install boost hdf5 eigen ninja;
+      brew install boost hdf5 eigen ninja;
+      if [ -z "$HOMEBREW_NO_AUTO_UPDATE" ]; then brew upgrade hdf5; fi;
```
matz-e: This seems brittle at best… does this rely on Travis caching an older `hdf5` version with the older XCode?

include/highfive/bits/H5_definitions.hpp
```diff
@@ -0,0 +1,46 @@
...
+#define H5_DEPRECATED(msg) __attribute__((deprecated(#msg)))
+#elif defined(_MSC_VER)
+#define H5_DEPRECATED(msg) __declspec(deprecated(#msg))
+#else
+#pragma message("WARNING: Compiler doesnt support deprecation")
+#define H5_DEPRECATED
+#endif
+
+
+// Fw declarations
```
matz-e:
```suggestion
// Forward declarations
```
It's not worth saving those few characters here ;)

.travis.yml
ohm314: I'm curious about that too? Could you explain @ferdonline

.travis.yml
```diff
@@ -109,10 +117,16 @@ env:
     - MINCONDA_OSX="MacOSX-x86_64"
 
 install:
+  - export HOMEBREW_NO_AUTO_UPDATE=1
```
ohm314: why do you have this complicated logic with first setting a veriable, then checking the parameter and unsetting it and then having the logic below.. i think this can be simplified?

.travis.yml
ferdonline: I was actually proud this is slightly more stable... now, by default I set `HOMEBREW_NO_AUTO_UPDATE=1` in the top, so all installs provide the same version
However, atm to use hdf5 1.12 we need to force the upgrade. I wish there was a better solution, but nope. In either casethis only applies to the second MacOS test, wont affect any other setup

.travis.yml
ferdonline: This is setting a safer default -> dont automatically update versions

ferdonline:
Than for the review @ohm314 and @matz-e, got motivation to find a cleaner solution! could you recheck?

ferdonline:
please retest

PR #312, opened 2020-03-09
Add HighFive target to export case and rework a little.

matz-e:
Libraries building on HighFive and linking against the "HighFive" target
publicly failed to configure properly, as the target "HighFive" was not
explicitly exported.

Further, reconstructing the target "HighFive" with dynamic dependency
finding lost the IMPORTED keyword, leading to the same issue. Copy
properties from a newly constructed target into the exported one.


matz-e:
@ferdonline works with CMake 3.1.0, installed via Späck on Ubuntu.

PR #317, opened 2020-03-12
Add more inclusive integration tests.

matz-e:
To prevent #316 in the future.


tests/test_library/include/simpleton.hpp
```diff
@@ -1,7 +1,20 @@
 #ifndef H5_TEST_SIMPLETON_HPP
 #define H5_TEST_SIMPLETON_HPP
 
+#include <highfive/H5Attribute.hpp>
```
ferdonline: We already have tests that #include headers one by one, I don't see as necessary all these imports here. Plus, most headers will import recursively a lot of others. 
Maybe in the future we create a HighFive.hpp master header.

tests/test_library/src/otherton.cpp
```diff
@@ -0,0 +1,9 @@
+#include <stdexcept>
+
+#include "simpleton.hpp"
+
+void noop(const HighFive::Object& obj) {
+    if (!obj.isValid()) {
```
ferdonline: For completeness we could create a Boost array (ensure that the brought dependencies linking works)

tests/test_library/include/simpleton.hpp
matz-e: I'd rather do this explicitly and dumb… trying to be as fool-proof as possible.

tests/test_library/src/otherton.cpp
matz-e: Should be a follow up, maybe?

tests/test_library/include/simpleton.hpp
ferdonline: I meant the tests using `tests_import_public_headers.cpp` which cover all the cases here, but it's not a blocker either way.

matz-e: I'd be up for a master header, the point here is to catch any missing `inline`s that show up when all headers are included in two different compilation units. Thus it may be a little different from the public import test…

tests/test_library/src/otherton.cpp
ohm314: @matz-e maybe there was a little confusion, why not rename this test case to be a little more explicit in what it is testing?

matz-e: @ohm314 good point, although I'm a little stuck on a better name. `test_dependent_library` maybe? I can also add a note for all the headers…

ohm314: yeap, that would be not bad I think!

PR #318, opened 2020-03-12
Compat with hdf5 1.12 : ObjectInfo

ferdonline:
### This PR
Leverages hdf5 compat API to keep supporting existing functions. Functions relying on such calls are marked `Deprecated`. In the next major release we can remove this API and only use the new hdf5 structure/calls.

The case here was `getAddress()` which relies on a `H5O_info_t` and `H5Oget_info`, which broke compat.


ferdonline:
pushed to original PR #311

PR #324, opened 2020-04-10
Simplifying imports, new policy

ferdonline:
## Problem

HighFive has suffered from include hell since some base classes require the definition of some of their own subclasses. In a world on header-only, this translated in order-dependent, or unsatisfiable include loops.

## This PR

Changes the general include policy. Instead of attempting to bring implementation always with the declaration, only headers with user-constructible classes shall bring their implementation. All other cases we leave the pure definition so they can be freely included anywhere we need it.
This greatly simplifies the include dependencies.

As an exception to this rule there's `H5Object.hpp` which is a fundamental header, to avoid including both headers everywhere.

## Implications

A large set of classes are not user-constructible and they can only be created by higher-order classes e.g. create a dataset from a Node (i.e. File or Group). If the user cannot create them directly they will forcibly include the factory class, which is required to bring all necessary implementations, directly or indirectly. If the user includes, e.g. by mistake, the header without implementation, being header only, not calling methods doesn't incur any error.

Therefore, for the user this change is *mostly* invisible. The only edge situation, as raised in comments, would be in case the user has a source file which only operates on a given entity, e.g. `Group` without constructing it. 
To avoid this from being a problem, as a new public policy, **users shall always include the top-level H5File.hpp** which is a sufficient include for the core.

### Extras

- Use `H5_declarations.hpp` everywhere where the fw declarations were required.
- Inherit, as protected, the Object constructors. This allows, among others, to deduplicate constructor code and drop default constructors in places it is arguably non-sense:
  - `DataSet`, `DataType` and `Group`
- Add move constructor to `Object` (which is noexcept!), used by "pilfering" constructors in `DataSet` and `Group`


include/highfive/H5Group.hpp
```diff
@@ -10,35 +10,30 @@
...
 
-    friend class ::HighFive::Reference;
+    friend class File;
+    friend class Reference;
+    template <typename Derivate> friend class ::HighFive::NodeTraits;
 };
-}
 
-#include "bits/H5Node_traits_misc.hpp"
-#include "bits/H5Group_misc.hpp"
```
matz-e, line 41: So, I'm interested in this. Thank you for your work! I like the premise of a simplified include structure…

If I were to have a class/namespace/function that only ever takes a `H5Group` and does operations with that group, or extracts `Datasets` from it, will this work? Wouldn't I have to include the implementation somewhere? From my naïve thinking, just including `H5Group.hpp` should allow me to compile a simple library operating only on said group…

include/highfive/bits/H5Reference_misc.hpp
```diff
@@ -45,7 +46,7 @@ inline T Reference::dereference(const Object& location) const {
         HDF5ErrMapper::ToException<ReferenceException>(
             "Trying to dereference the wrong type");
     }
-    return T(obj);
+    return std::move(obj);
```
ohm314, line 49: I feel like I keep going back to the same question over and over.. but why the move specifically here? I thought this is a case for copy-illision?

include/highfive/H5Reference.hpp
```diff
@@ -64,8 +55,8 @@ class Reference {
 
   protected:
     /// \brief Create a Reference from a low-level HDF5 object reference
-    explicit Reference(const hobj_ref_t h5_ref)
-        : href(h5_ref){};
+    inline explicit Reference(const hobj_ref_t h5_ref)
```
ohm314, line 58: Since this constructor is defined inside the class declaration, is it not automatically `inline` ?

include/highfive/H5File.hpp
```diff
@@ -72,11 +72,16 @@ class File : public Object,
...
 }  // namespace HighFive
 
+
+// H5File is the main user constructible -> bring in implementation headers
+#include "bits/H5Annotate_traits_misc.hpp"
 #include "bits/H5File_misc.hpp"
-#include "H5Attribute.hpp"  // for AnnotateTraits
-#include "H5Group.hpp"      // for Node
+#include "bits/H5Node_traits_misc.hpp"
+#include "bits/H5Object_misc.hpp"
```
ohm314, line 83: So I understand correctly that `H5Object_misc.hpp` is included here because it will be needed in user code?

include/highfive/H5DataSpace.hpp
```diff
@@ -150,14 +149,15 @@ class DataSpace : public Object {
...
-    explicit DataSpace();
+    DataSpace() = default;
 
     friend class Attribute;
     friend class File;
     friend class DataSet;
 };
 }
 
+// We include bits right away since DataSpace can be constructed stand-alone
```
ohm314, line 160: so why do we also include bits of `H5Object` ? I would think that class is not contructed stand-alone? Unless I misunderstand the meeting of _stand-alone_ here.

ferdonline: Ok I didn’t apply the simplification rules to classes whose implementation is self contained, like H5object, and therefore not prone to be problematic. Otherwise we would have to track everywhere where h5object.hpp is used and include additionally h5object_misc. I’ll add a note

include/highfive/H5Group.hpp
ferdonline: > If I were to have a class/namespace/function that only ever takes a H5Group and does operations with that group

Internally we know that anything manipulating group must include the implementation.

For the user, in most cases it is a no-problem since the user must have created the group from a File, and then (s)he has all the required header
In the rare case someone wants to do a library or something operating in a group, then he must include `H5File.hpp`. Actually **we probably should update the docs** stating this all new developments should prefer doing the single include of h5File.hpp

include/highfive/bits/H5Reference_misc.hpp
ferdonline: In this case we can't have copy-illision, we have different classes (or am I missing something?)
What I did was to transform a constructor by coping the base class to a constructor which moves it (since this PR implements Object move ctor).

include/highfive/H5Reference.hpp
ferdonline: I thought that user-defined constructors follow same rules as functions, i.e. not inline. I did a quick search in cppreference and did not find anything on it, but Id be happy to know the fine print.

include/highfive/H5File.hpp
ferdonline: Apart from the fact that `H5File.hpp` was promoted to the main header, implementation of headers which were simplified must be brought where those things are used.
Now, you have a point here since H5Object, as you mentioned below, already bring bits.
I'll another pass of checking these cases

include/highfive/H5DataSpace.hpp
ferdonline: True, stand-alone was not the best wording. It's user constructible, and that's a reason strong enough.
Btw, H5Object also brings implementation since it's a fundamental block, I think it's the only exception.

include/highfive/H5Group.hpp
matz-e: OK, if this going to be the official policy, I'm fine with that. Thank you for responding!

ferdonline:
@tristan0x @alkino do you agree with this change?

include/highfive/bits/H5Reference_misc.hpp
ohm314: After discussion all clear. thank you!

include/highfive/H5Reference.hpp
ohm314: Just for completeness: 
https://en.cppreference.com/w/cpp/language/inline (second paragraph)
`inline` is not necessary here because it's implicit (but it's not a mistake).

PR #327, opened 2020-04-16
SFINAE for eigen

eudoxos:
This is something like proof of concept (it does not compile) for SFINAE so that load/dump can be easily tailored for specific types incl. CPRT (Eigen).

Generic functions (for scalars, strings) are limited via the `is_easy`, specially treated is vector with `is_vector`. This is good for dumping.

For loading, I am too dumb to make load_impl available with enable_if only as needed, maybe someone could figure that out better than me.

### DIFF
--- a/include/highfive/H5Easy.hpp
+++ b/include/highfive/H5Easy.hpp
@@ -73,26 +73,29 @@ inline size_t getSize(const File& file, const std::string& path);
 /// \return the shape of the DataSet
 inline std::vector<size_t> getShape(const File& file, const std::string& path);
 
-///
-/// \brief Write an Eigen matrix to a new DataSet in an open HDF5 file.
-///
-/// \param file Writeable opened file
-/// \param path Path of the DataSet
-/// \param data eigen matrix to write
-/// \param mode Write mode
-///
-/// \return the newly created DataSet
-///
-#ifdef H5_USE_EIGEN
-template <class T, int Rows, int Cols, int Options, int MaxRows, int MaxCols>
-inline DataSet dump(File& file,
-                    const std::string& path,
-                    const Eigen::Matrix<T, Rows, Cols, Options, MaxRows, MaxCols>& data,
-                    DumpMode mode = DumpMode::Create);
-#endif
+template<typename T>
+struct is_easy{
+	static const bool value=std::is_arithmetic<T>::value
+        || std::is_same<char *, typename std::decay<T>::type>::value
+        || std::is_same<const char *, typename std::decay<T>::type>::value
+        || std::is_same<std::string, typename std::decay<T>::type>::value
+	;
+};
+
+
+template<class T>
+struct is_vector {
+    static const bool value = false;
+};
+
+template<class T>
+struct is_vector<std::vector<T>> {
+    const static bool value = true;
+};
+
 
 ///
-/// \brief Write an Eigen::Ref object to a new DataSet in an open HDF5 file.
+/// \brief Write any (dense) Eigen object to a new DataSet in an open HDF5 file.
 ///
 /// \param file Writeable opened file
 /// \param path Path of the DataSet
@@ -105,27 +108,10 @@ inline DataSet dump(File& file,
 template <class T>
 inline DataSet dump(File& file,
                     const std::string& path,
-                    const Eigen::Ref<T>& data,
+                    const Eigen::DenseBase<T>& data,
                     DumpMode mode = DumpMode::Create);
 #endif
 
-///
-/// \brief Write an Eigen::Map object to a new DataSet in an open HDF5 file.
-///
-/// \param file Writeable opened file
-/// \param path Path of the DataSet
-/// \param data eigen matrix to write
-/// \param mode Write mode
-///
-/// \return the newly created DataSet
-///
-#ifdef H5_USE_EIGEN
-template <class T>
-inline DataSet dump(File& file,
-                    const std::string& path,
-                    const Eigen::Map<T>& data,
-                    DumpMode mode = DumpMode::Create);
-#endif
 
 ///
 /// \brief Write "xt::xarray<T>" to a new DataSet in an open HDF5 file.
@@ -163,6 +149,7 @@ inline DataSet dump(File& file,
                     DumpMode mode = DumpMode::Create);
 #endif
 
+
 ///
 /// \brief Write "std::vector<T>" to a new DataSet in an open HDF5 file.
 ///
@@ -173,10 +160,10 @@ inline DataSet dump(File& file,
 ///
 /// \return the newly created DataSet
 ///
-template <class T>
+template <class T, typename std::enable_if<is_vector<T>::value,int>::type=0>
 inline DataSet dump(File& file,
                     const std::string& path,
-                    const std::vector<T>& data,
+                    const T& data,
                     DumpMode mode = DumpMode::Create);
 
 ///
@@ -189,7 +176,7 @@ inline DataSet dump(File& file,
 ///
 /// \return The newly created DataSet
 ///
-template <class T>
+template <class T, typename std::enable_if<is_easy<T>::value,int>::type=0>
 inline DataSet dump(File& file,
                     const std::string& path,
                     const T& data,
@@ -205,7 +192,7 @@ inline DataSet dump(File& file,
 ///
 /// \return The newly created DataSet
 ///
-template <class T>
+template <class T, typename std::enable_if<is_easy<T>::value,int>::type=0>
 inline DataSet dump(File& file,
                     const std::string& path,
                     const T& data,
@@ -220,6 +207,7 @@ inline DataSet dump(File& file,
 ///
 /// \return the read data
 ///
+// template <class T, typename std::enable_if<is_easy<T>::value,int>::type=0>
 template <class T>
 inline T load(const File& file,
               const std::string& path,
@@ -233,8 +221,12 @@ inline T load(const File& file,
 ///
 /// \return the read data
 ///
+// template <class T, typename std::enable_if<is_easy<T>::value,int>::type=0>
 template <class T>
 inline T load(const File& file, const std::string& path);
+
+
+
 
 }  // namespace H5Easy
 
--- a/include/highfive/H5Easy.hpp
+++ b/include/highfive/H5Easy.hpp
@@ -49,10 +49,10 @@ using HighFive::File;
 
 ///
 /// \brief Write mode for DataSets
-enum class DumpMode
-{
-  Create, /*!< Dump only if DataSet does not exist, otherwise throw. */
-  Overwrite /*!< If DataSet already exists overwrite it if data has the same shape, otherwise throw. */
+enum class DumpMode {
+    Create,   /*!< Dump only if DataSet does not exist, otherwise throw. */
+    Overwrite /*!< If DataSet already exists overwrite it if data has the same shape,
+                 otherwise throw. */
 };
 
 ///
@@ -72,7 +72,6 @@ inline size_t getSize(const File& file, const std::string& path);
 ///
 /// \return the shape of the DataSet
 inline std::vector<size_t> getShape(const File& file, const std::string& path);
-
 
 ///
 /// \brief Write scalar/string to a new DataSet in an open HDF5 file.
@@ -101,10 +100,8 @@ inline DataSet dump(File& file,
 /// \return The newly created DataSet
 ///
 template <class T>
-inline DataSet dump(File& file,
-                    const std::string& path,
-                    const T& data,
-                    const std::vector<size_t>& idx);
+inline DataSet
+dump(File& file, const std::string& path, const T& data, const std::vector<size_t>& idx);
 
 ///
 /// \brief Load entry "(i,j)" from a rank-two DataSet in an open HDF5 file to a scalar.
@@ -116,9 +113,7 @@ inline DataSet dump(File& file,
 /// \return the read data
 ///
 template <class T>
-inline T load(const File& file,
-              const std::string& path,
-              const std::vector<size_t>& idx);
+inline T load(const File& file, const std::string& path, const std::vector<size_t>& idx);
 
 ///
 /// \brief Load a DataSet in an open HDF5 file to an object (templated).
@@ -133,10 +128,10 @@ inline T load(const File& file, const std::string& path);
 
 }  // namespace H5Easy
 
+#include "h5easy_bits/H5Easy_Eigen.hpp"
 #include "h5easy_bits/H5Easy_misc.hpp"
 #include "h5easy_bits/H5Easy_scalar.hpp"
 #include "h5easy_bits/H5Easy_vector.hpp"
-#include "h5easy_bits/H5Easy_Eigen.hpp"
 #include "h5easy_bits/H5Easy_xtensor.hpp"
 
 #endif  // H5EASY_HPP
--- a/include/highfive/H5Easy.hpp
+++ b/include/highfive/H5Easy.hpp
@@ -73,99 +73,7 @@ inline size_t getSize(const File& file, const std::string& path);
 /// \return the shape of the DataSet
 inline std::vector<size_t> getShape(const File& file, const std::string& path);
 
-template<typename T>
-struct is_easy{
-	static const bool value=std::is_arithmetic<T>::value
-        || std::is_same<char *, typename std::decay<T>::type>::value
-        || std::is_same<const char *, typename std::decay<T>::type>::value
-        || std::is_same<std::string, typename std::decay<T>::type>::value
-	;
-};
-
-
-template<class T>
-struct is_vector {
-    static const bool value = false;
-};
-
-template<class T>
-struct is_vector<std::vector<T>> {
-    const static bool value = true;
-};
-
-
-///
-/// \brief Write any (dense) Eigen object to a new DataSet in an open HDF5 file.
-///
-/// \param file Writeable opened file
-/// \param path Path of the DataSet
-/// \param data eigen matrix to write
-/// \param mode Write mode
-///
-/// \return the newly created DataSet
-///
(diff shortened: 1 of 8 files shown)


tdegeus:
Thanks. Handling Eigen more robustly is better indeed. A quick feedback/question is that it would be good to have a grip on which overloads we still miss. For one I think that nested vectors are missing. Maybe some the core team can also comment @ferdonline @alkino

eudoxos:
What is the target standard? c++11, c++14, c++17, c++20? Makes such things easier and easier, it seems.

tdegeus:
I think that at the moment everything is still c++11. Personally I have no problems with c++14, but I'm not sure that this generalisation is worth changing standards.

eudoxos:
So this was an easy fix in the end, for loading, there is even no need to restrict the base template at all due to partial specialization. Unit tests for more Eigen types added.

I did not do it for xtensor yet (not being sure I understand how its classes are organized — simple templates for `xarray` and `xtensor` or something more involved?) but that should really be trivial.

eudoxos:
In a broad view, the (function) `dump` is more difficult as the compiler is using [overload resolution](https://en.cppreference.com/w/cpp/language/function_template#Function_template_overloading) (for templated functions), which is counter-intuitive for complicated templated types such as CRTP. There is no such thing as partial specialization for functions.

`load_impl` is a `struct`, thus [partial specialization](https://en.cppreference.com/w/cpp/language/partial_specialization) can be used; its rules feel more "natural" as there is the notion of "base" tempalte and (partially) specialized template which are preferred if possible.

It might make sense to wrap `dump` in a struct as well; that `is_easy` trait might go, in that case.

tdegeus:
Thanks! I played around, is this what you had in mind?
```cpp

#include <vector>
#include <type_traits>

#include <Eigen/Eigen>
#include <xtensor/xtensor.hpp>
#include <xtensor/xarray.hpp>

template <class... T>
struct make_void
{
    using type = void;
};

template <class... T>
using void_t = typename make_void<T...>::type;

namespace detail
{

    template<class, typename = void>
    struct type_dispatch : std::integral_constant<int, 0> {};

    template <class T>
    struct type_dispatch<T, std::enable_if_t<std::is_same<xt::xarray<typename T::value_type>, T>::value>>
        : std::integral_constant<int, 1> {};

    // template <class T>
    // struct type_dispatch<T, std::enable_if_t<std::is_same<xt::xtensor<typename T::value_type, size_t>, T>::value>>
    //     : std::integral_constant<int, 2> {};

    template <class T>
    struct type_dispatch<T, std::enable_if_t<std::is_same<std::vector<typename T::value_type>, T>::value>>
        : std::integral_constant<int, 3> {};

    template <class T>
    struct type_dispatch<T, std::enable_if_t<std::is_base_of<Eigen::DenseBase<T>, T>::value>>
        : std::integral_constant<int, 4> {};

    template <class T>
    inline auto dump(const T& t, std::integral_constant<int, 0>)
    {
        return "Other";
    }

    template <class T>
    inline auto dump(const T& t, std::integral_constant<int, 1>)
    {
        return "xarray";
    }

    // template <class T>
    // inline auto dump(const T& t, std::integral_constant<int, 2>)
    // {
    //     return "xtensor";
    // }

    template <class T>
    inline auto dump(const T& t, std::integral_constant<int, 3>)
    {
        return "vector";
    }

    template <class T>
    inline auto dump(const T& t, std::integral_constant<int, 4>)
    {
        return "Eigen";
    }

}

template <class T>
inline auto dump(const T& t)
{
    return detail::dump(t, detail::type_dispatch<T>{});
}

int main()
{
    Eigen::ArrayXf A = Eigen::ArrayXf::Random(50);
    Eigen::MatrixXd B = Eigen::MatrixXd::Random(20, 5);
    xt::xarray<double> C = {{1, 2, 3}, {4, 5, 6}};
    std::vector<double> D = {1, 2, 3};
    int E = 0;

    std::cout << dump(A) << std::endl;
    std::cout << dump(B) << std::endl;
    std::cout << dump(C) << std::endl;
    std::cout << dump(D) << std::endl;
    std::cout << dump(E) << std::endl;

    return 0;
}
```

eudoxos:
What you wrote looks funny, in a good sense :) There is the `type_dispatch` struct  used only as tag for the dump functions. Why not using a struct's member function directly? More elegant. That's what I was after, modeling `dump` after `load`? In this sense:

* rename `load_impl` to `io_impl` (or something like that)
* rename `load_impl::run` to `io_impl::load`
* add `io_impl::dump`, `io_impl::dump_overwrite` and `io_impl::dump_extend`.

Then each type of I/O would only have to specialize this `io_impl` class with 4 methods (`load`, `dump`, `dump_overwrite`, `dump_extend`). At the moment, that would be `H5Easy::detail::{scalar,eigen,xtensor}::io_impl`.

tdegeus:
I think you are right that that could be better indeed! Can you work on the plain example, thereafter everything should become straightforward?

eudoxos:
Please have a look here: https://gist.github.com/eudoxos/43f11cfb04d33b4f3e5cf309ad628365

tdegeus:
Nice! I added also `xt::xtensor` : https://github.com/tdegeus/cpp_examples/blob/master/sfinae/main.cpp
Shall update the implementation, or do you want to have a go at it?

eudoxos:
Let me try a bit, I will let you know :)

eudoxos:
Quick question: in code like this:

```
inline std::vector<size_t> shape(const T& data) {
	if (std::decay<T>::type::RowsAtCompileTime == 1) {
		return {static_cast<size_t>(data.cols())};
	}
	/* ... */
}
```
is there some reason for the `std::decay<T>::type`? Because one can just write `data::RowsAtCompileTime` as well.

eudoxos:
There are some issue with xtensor; I left them unresolved not being familiar with it. The rest works very nice, actually the code shrunk significantly.

tdegeus:
Nice! I think indeed the code is more elegant now! *And* Eigen is supported more generally. Great that we insisted a bit!

> Quick question: in code like this:
> 
> ```
> inline std::vector<size_t> shape(const T& data) {
> 	if (std::decay<T>::type::RowsAtCompileTime == 1) {
> 		return {static_cast<size_t>(data.cols())};
> 	}
> 	/* ... */
> }
> ```
> 
> is there some reason for the `std::decay<T>::type`? Because one can just write `data::RowsAtCompileTime` as well.

If `data::RowsAtCompileTime` works (also for objects with a fixed number of rows/columns like `Eigen::VectorXd` I see no objection. I didn't think it would.

> There are some issue with xtensor; I left them unresolved not being familiar with it. The rest works very nice, actually the code shrunk significantly.
> 
> the `io_impl_base` has the common functionality, then `io_impl<T>` inherit from it and implement their bits as needed. It is explained in comments.

Maybe I'm overlooking something, but everything is there no? Are all variants currently in included in the CI?

For the rest I will leave a review.

include/highfive/h5easy_bits/H5Easy_Eigen.hpp
```diff
@@ -19,91 +19,86 @@ namespace H5Easy {
...
+		}
+
+		throw detail::error(file, path, "H5Easy::load: Inconsistent rank");
+	}
+
+	// write to open DataSet of the correct size
+	// (use Eigen::Ref to convert to RowMajor; no action if no conversion is needed)
+	inline static void write(DataSet& dataset, const T& data) {
+		Eigen::Ref<
+			const Eigen::Array<
```
tdegeus, line 64: Just out or curiosity, what is the difference between `Array` and `Matrix`? I was under the impression that `Eigen` only had matrices.

include/highfive/h5easy_bits/H5Easy_scalar.hpp
```diff
@@ -16,27 +16,34 @@ namespace H5Easy {
...
+* dump_extend
+* load
+* load_part
 
-// create DataSet and write data
-template <class T, class E = void>
-struct dump_impl
+*/
+template<typename T>
+struct io_impl_base
```
tdegeus, line 34: What is the reason to this extra name? Is seems that it requires you to make an alias below. I would simply use one name

include/highfive/h5easy_bits/H5Easy_scalar.hpp
```diff
@@ -16,27 +16,34 @@ namespace H5Easy {
 
 namespace detail {
 
-namespace scalar {
+/*
+This base class presents fallback methods which will be automatically
+called from specializations, if not defined in the specialization.
+Every specialization of io_impl<T> should inherit from io_impl_base<T>.
```
tdegeus, line 22: I don't think this is entirely true, at least not for the moment. If you call `load_part` with e.g. an Eigen object you will get a compiler error because `load_part` is not in the `struct` enabled for `Eigen` objects. I don't see this as a problem a-priori: the methods could be added if so desired. For the moment defining the methods and putting a `static_assert` might improve readability.

include/highfive/h5easy_bits/H5Easy_scalar.hpp
```diff
@@ -94,13 +96,8 @@ struct dump_extend_impl
...
         return dataset;
     }
-};
-
-// load "scalar" as part of a larger DataSet
-template <class T>
-struct load_impl
-{
-    static T run(const File& file,
+    // load a part of a larger DataSet
```
tdegeus, line 99: The previous comment "a scalar" still remains true no? That being a said, a static_assert could be a nice extension here. 

Furthermore, it would be good to keep the white-line between functions.

include/highfive/h5easy_bits/H5Easy_vector.hpp
```diff
@@ -17,31 +17,30 @@ namespace H5Easy {
...
-template <class T, class E = void>
-struct overwrite_impl
-{
-    static DataSet run(File& file, const std::string& path, const std::vector<T>& data)
+/*
+This vector specialization does not implement load, load_part, dump_extend.
+Since it inherits from io_impl_base, implementations there will be called automatically.
+ */
+template<typename T>
+struct io_impl<T,typename std::enable_if<is_vector<T>::value>::type>: public io_impl_base<T> {
```
tdegeus, line 31: If I understand well, a nested vector is also capture here, right?

ferdonline:
Thanks @eudoxos for this contribution which looks really nice.
I didn't have time to review the details, I would just like to comment on the c++ standard.
We want to keep the widest compatibility possible across compilers/platform and C++11 is a good compromise. So far we didn't find strong enough reasons to move up. Eventually some constructs become slightly longer but that's a price we're willing to pay

include/highfive/h5easy_bits/H5Easy_Eigen.hpp
```diff
@@ -19,91 +19,86 @@ namespace H5Easy {
...
+				static_cast<size_t>(data.cols())};
+	}
+	
+	using EigenIndex = Eigen::DenseIndex;
+
+	// get the shape of a "DataSet" as size 2 "std::vector<Eigen::Index>"
+	inline static std::vector<EigenIndex> shape(const File& file,
+                                               const std::string& path,
+                                               const DataSet& dataset,
+                                               int RowsAtCompileTime) {
```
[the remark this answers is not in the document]
ferdonline: please dont use tabs. Find the conventions in `.clang-format`

include/highfive/h5easy_bits/H5Easy_Eigen.hpp
eudoxos: `Matrix` is... mathematical matrix (and `Vector*` is 1D matrix), Array (1D or 2D) is element-wise by default. Converting (compile-time, no allocation) between them is done with `.matrix()` and `.array()` methods. As we are using arrays just as a bunch of numbers grid, that intent is clear by using `Array`.

include/highfive/h5easy_bits/H5Easy_scalar.hpp
eudoxos: As far as I understood, `std::vector` uses some methods for "scalars" directly, this makes it possible; otherwise they would not be accessible, would have to be duplicated.

include/highfive/h5easy_bits/H5Easy_vector.hpp
eudoxos: Yes, of course. The idea is that e.g. `vector<set<T>>` will fail in the scalar routine called with `set<T>`.

eudoxos:
> We want to keep the widest compatibility possible across compilers/platform and C++11 is a good compromise. So far we didn't find strong enough reasons to move up. Eventually some constructs become slightly longer but that's a price we're willing to pay

Understood. Just saying that `xtensor` headers fail to be parsed as `-std=c++11`.

include/highfive/H5Easy.hpp
```diff
@@ -49,10 +49,10 @@ using HighFive::File;
...
 ///
 /// \brief Write mode for DataSets
-enum class DumpMode
-{
-  Create, /*!< Dump only if DataSet does not exist, otherwise throw. */
-  Overwrite /*!< If DataSet already exists overwrite it if data has the same shape, otherwise throw. */
+enum class DumpMode {
+    Create,   /*!< Dump only if DataSet does not exist, otherwise throw. */
+    Overwrite /*!< If DataSet already exists overwrite it if data has the same shape,
+                 otherwise throw. */
```
tdegeus, line 55: Maybe some rules can be broken ;)

include/highfive/h5easy_bits/H5Easy_Eigen.hpp
```diff
@@ -19,165 +19,105 @@ namespace H5Easy {
...
+    // (use Eigen::Ref to convert to RowMajor; no action if no conversion is needed)
+    inline static void write(DataSet& dataset, const T& data) {
+        Eigen::Ref<const Eigen::Array<typename std::decay<T>::type::Scalar,
+                                      std::decay<T>::type::RowsAtCompileTime,
+                                      std::decay<T>::type::ColsAtCompileTime,
+                                      std::decay<T>::type::ColsAtCompileTime == 1
+                                          ? Eigen::ColMajor
+                                          : Eigen::RowMajor,
+                                      std::decay<T>::type::MaxRowsAtCompileTime,
+                                      std::decay<T>::type::MaxColsAtCompileTime>,
```
tdegeus, line 71: I would keep one argument per line by adding a line-break before the first argument and changing the indentation (see original version)

include/highfive/h5easy_bits/H5Easy_Eigen.hpp
```diff
@@ -19,165 +19,105 @@ namespace H5Easy {
...
-struct load_impl {
-    static T run(const File& file, const std::string& path) {
+
+    // load from DataSet
+    // convert to ColMajor if needed (HDF5 always stores row-major)
+    static T load(const File& file, const std::string& path) {
         DataSet dataset = file.getDataSet(path);
-        std::vector<typename T::Index> dims = shape(file, path, dataset, T::RowsAtCompileTime);
+        std::vector<typename T::Index> dims = shape(file, path, dataset,
+                                                    T::RowsAtCompileTime);
```
tdegeus, line 104: Can be one line, this looks awful.

include/highfive/h5easy_bits/H5Easy_Eigen.hpp
```diff
@@ -19,165 +19,105 @@ namespace H5Easy {
...
-    static Eigen::Matrix<T,Rows,Cols,Options,MaxRows,MaxCols>
-    run(const File& file,
-        const std::string& path)
-    {
-        return detail::eigen::load_impl<
-            Eigen::Matrix<T,Rows,Cols,Options,MaxRows,MaxCols>>::run(file, path);
+        return Eigen::Map<
+            Eigen::Array<typename T::Scalar, T::RowsAtCompileTime, T::ColsAtCompileTime,
+                         T::ColsAtCompileTime == 1 ? Eigen::ColMajor : Eigen::RowMajor,
+                         T::MaxRowsAtCompileTime, T::MaxColsAtCompileTime>>(
```
tdegeus, line 115: One argument per line, see original version.

include/highfive/h5easy_bits/H5Easy_scalar.hpp
```diff
@@ -65,7 +62,8 @@ struct dump_extend_impl
             std::vector<size_t> shape = dims;
             if (dims.size() != idx.size()) {
                 throw detail::error(file, path,
-                    "H5Easy::dump: Rank of the index and the existing field do not match");
+                                    "H5Easy::dump: Rank of the index and the existing "
+                                    "field do not match");
```
tdegeus, line 66: Keep original

include/highfive/h5easy_bits/H5Easy_scalar.hpp
```diff
@@ -94,32 +92,17 @@ struct dump_extend_impl
...
-
-}  // namespace scalar
-
-// load from DataSet
-template <class T, class E = void>
-struct load_impl
-{
-    static T run(const File& file, const std::string& path)
-    {
+    // load entire DataSet
```
tdegeus, line 104: I think it always a good idea to keep whitelines between functions

eudoxos: Understood. I just used `.clang-format`.

include/highfive/h5easy_bits/H5Easy_scalar.hpp
tdegeus: I think is a reply to the next comment. (The one here is I think still important to discuss). My point was that the `io_impl` of `std::vector<T>` will be enabled when called with a `std::vector<T>` so I guess any function in `H5Easy_scalar` can not be reached with any `std::vector<T>`

eudoxos:
Feel free to undo the last commit (formatting-only) and format as you see fit yourself, perhaps easier.

alkino:
Really nice job @eudoxos

PR #332, opened 2020-04-21
expose close() function on Highfive file to allow force closing

Dekken:
Hi there,

We've noticed that we cannot overwrite h5 files during the same binary execution unless we close all handles to the files.

### DIFF
--- a/include/highfive/H5File.hpp
+++ b/include/highfive/H5File.hpp
@@ -69,7 +69,14 @@ class File : public Object,
     ///
     void flush();
 
- private:
+    ///
+    /// \brief close
+    ///
+    /// Flushes then closes file handle
+    ///
+    void close();
+
+  private:
     std::string _filename;
 };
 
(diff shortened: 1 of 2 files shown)


tristan0x:
Hello @Dekken 
Thank you very much for your contribution.

HighFive uses RAII paradigm to manage IO resources. It internally relies on HDF5 reference counting to close HDF5 file when the last C++ object using it is destructed. I am not an HDF5 expert but I am not very fond of this addition because  IMO it breaks this RAII paradigm. Besides your C++ H5File instance will be unusable after calling this `close` method. Can't you destruct your H5File objects when you don't need them instead?

Dekken:
We thought we were destructing them, as our fix is to call close() in the destructor of the wrapper class which contains a stack HighFive::File variable.
```
https://github.com/PHAREHUB/PHARE/blob/master/src/diagnostic/detail/h5file.h#L13
```

Dekken:
thanks for your time, I think our open handle was with h5py

PR #336, opened 2020-04-28
Add a mechanism to not include target HighFive several times

alkino:
I was inspired by Qt5CoreConfig.cmake.

We have to do that if the project is included several time, for example by Brion.
`Brion` include `HighFive` and `MVDTool`.
`MVDTool` include `HighFive`
This cause a failure because the target `HighFive` is defined twice.

### DIFF
--- a/CMake/HighFiveConfig.cmake.in
+++ b/CMake/HighFiveConfig.cmake.in
@@ -11,56 +11,58 @@ function(copy_interface_properties target source)
   endforeach()
 endfunction()
 
-# Get HighFive targets
-include("${CMAKE_CURRENT_LIST_DIR}/HighFiveTargets.cmake")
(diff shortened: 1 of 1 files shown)


alkino:
I add you as a reviewer @matz-e because the problem is shown by https://github.com/BlueBrain/spack/pull/782

alkino:
I don't see any scenario it can hide, except "double inclusion"

matz-e:
> I don't see any scenario it can hide, except "double inclusion"

Double inclusion of different versions? This could be checked by comparing the target properties with the installation location, I guess. If they are not identical, error out?

Also, I'm wondering if this is not masking another issue here. Why is Brion exposing a HighFive target? Isn't Brion a compiled library and thus this should be a private target?

Issue #301, opened 2020-02-27
Easy compression in H5Easy


spacescientist:
##### Detailed description:
It would be great to be able to easily enable compression when using H5Easy, especially given e.g. its Eigen support. Is there currently a way to specify `Chunking`, `Deflate` and `Shuffle` in H5Easy for Eigen, xtensor, etc..?

##### Tentative first step towards a solution:

As I couldn't find one, I've tried this (I'm not at all familiar with this, and I didn't find much on the subject at all, but #117 and `tests_high_five_base.cpp` were useful):
###### H5Easy.cpp
```cpp
// add the following:
using HighFive::Shuffle;
using HighFive::Deflate;
```
###### H5Easy_Eigen.cpp, `dump_impl`
```cpp
// create DataSet and write data
template <class T>
static DataSet dump_impl(File& file,
                         const std::string& path,
                         const T& data)
{
    using type = typename std::decay_t<T>::Scalar;
    detail::createGroupsToDataSet(file, path);

// NEW v [
    DataSetCreateProps props;
    props.add(Chunking(std::vector<hsize_t>{(hsize_t)data.rows(), (hsize_t)data.cols()})); // explicit cast to avoid complains from Eigen. Note that rows() and cols() are apparently 'long int'.
    props.add(Shuffle());
    const int deflate_level = 9; // hardcoded for now, should be given as an argument.
    props.add(Deflate(deflate_level));
    DataSet dataset = file.createDataSet<type>(path, DataSpace(shape(data)), props);
// NEW ^ ]
//    DataSet dataset = file.createDataSet<type>(path, DataSpace(shape(data)));
    detail::eigen::write(dataset, data);
    file.flush();
    return dataset;
}
```
(nb: I didn't take care of `overwrite_impl`)

###### main.cpp
```cpp
H5Easy::File output_hdf("out.hdf", H5Easy::File::Overwrite);
H5Easy::dump(output_hdf, dataset_name, eigen_map, H5Easy::DumpMode::Overwrite);
```

I've tested with a given dataset and, as a prototype, it seems work, namely: 
- the output file from the original `H5Easy::dump` is 12 Mb;
- the one using this prototype is 1.5 Mb;

and they are identical according to `h5diff` (or, equivalently, `diff` the outputs of `h5totxt`).

Cheers.

alkino:
In fact the function should take a `DataSetCreateProps` and not the `deflate_level`.

spacescientist:
[...]

PR #288, opened 2020-02-19
Proposal: Add Attributes to H5Easy

tdegeus:
This is a proposal for the extension of the H5Easy API. It would allow:

```cpp
#include <Eigen/Eigen>
#include <highfive/H5Easy.hpp>

int main()
{
    Eigen::Vector3d a;
    a(0) = 1.0;
    a(1) = 2.0;
    a(2) = 3.0;

    H5Easy::File data("test.h5", H5Easy::File::Overwrite);

    H5Easy::dump(data, "/a", a);
    H5Easy::dump_attribute(data, "/a", "desc", std::string("Foobar"));
    std::cout << H5Easy::load_attribute<std::string>(data, "/a", "desc") << std::endl;
}
```

Comments would be great!

PR #337, opened 2020-04-29
Adding missing function to H5Attribute; [H5Easy] Enabling compression & Adding attributes

### LINKED ISSUE #340: Dump without Flushing
Is there an alternative to `H5Easy::dump(...)` for an `Eigen::Matrix` that does not flush the file?

tdegeus:
Allows a variable number of dump-settings (in arbitrary order): The original `H5Easy::DumpMode`, the new `H5Easy::Compression`, and other further other options.

Can you give feedback? @spacescientist @eudoxos 

Fixes https://github.com/BlueBrain/HighFive/issues/301
Fixes #340

### DIFF
--- a/include/highfive/H5Easy.hpp
+++ b/include/highfive/H5Easy.hpp
@@ -54,17 +54,133 @@ using HighFive::Shuffle;
 ///
 /// \brief Write mode for DataSets
 enum class DumpMode {
-    Create, /*!< Dump only if DataSet does not exist, otherwise throw. (default) */
-    Overwrite /*!< If DataSet already exists overwrite it if data has the same shape, otherwise throw. */
+    Create = 0, /*!< Dump only if DataSet does not exist, otherwise throw. */
+    Overwrite = 1 /*!< Create or overwrite if DataSet of correct shape exists, otherwise throw. */
 };
 
 ///
-/// \brief Compression-level for writing DataSets
+/// \brief Enable/disable automatic flushing.
+enum class Flush
+{
+    False = 0, /*!< No automatic flushing. */
+    True = 1 /*!< Automatic flushing. */
+};
+
+///
+/// \brief Enable/disable compression for written DataSets.
 enum class Compression
 {
-    None, /*!< No compression. (default) */
-    Medium, /*!< Medium compression level. */
-    High /*!< High compression level. */
+    None = 0, /*!< No compression. (default) */
+    Medium = 5, /*!< Medium compression level. */
+    High = 9 /*!< High compression level. */
+};
+
+///
+/// \brief Set compression-level for written DataSets.
+class CompressionLevel
+{
+public:
+    CompressionLevel(unsigned deflate_level);
+
+    inline unsigned get() const;
+
+private:
+    unsigned m_deflate_level;
+};
+
+///
+/// Dump-settings.
+///
+/// By default:
+/// - DumpMode::Create
+/// - Flush::True
+/// - Compression::False
+/// - ChunkSize: automatic.
+class DumpOptions
+{
+public:
+    ///
+    /// \brief Dump-settings.
+    DumpOptions() = default;
+
+    ///
+    /// \brief Dump-settings
+    /// \param Any of DumpMode, Flush, Compression in arbitrary number and order.
+    template <class... Args>
+    DumpOptions(Args... args);
+
+    ///
+    /// \brief Set setting.
+    /// \param mode: DumpMode.
+    inline void set(DumpMode mode);
+
+    ///
+    /// \brief Set setting.
+    /// \param flush: Flush.
+    inline void set(Flush flush);
+
+    ///
+    /// \brief Set setting.
+    /// \param level: Compression.
+    inline void set(Compression level);
+
+    ///
+    /// \brief Set setting.
+    /// \param level: CompressionLevel
+    inline void set(const CompressionLevel& level);
+
+    ///
+    /// \brief Set settings.
+    /// \param Any of DumpMode, Flush, Compression in arbitrary number and order.
+    template <class T, class... Args>
+    inline void set(T arg, Args... args);
+
+    ///
+    /// \brief Set deflate level.
+    /// \param level: deflate level (0-9).
+    inline void setDeflateLevel(unsigned level);
+
+    ///
+    /// \brief Set chunk-size. If the input is rank zero, automatic chunking is enabled.
+    /// \param shape: chunk size along each dimension.
+    template <class T>
+    inline void setChunkSize(const std::vector<T>& shape);
+
+    ///
+    /// \brief Set chunk-size. If the input is rank zero, automatic chunking is enabled.
+    /// \param shape: chunk size along each dimension.
+    inline void setChunkSize(std::initializer_list<size_t> shape);
+
+    ///
+    /// \brief Check to overwrite.
+    inline bool Overwrite() const;
+
+    ///
+    /// \brief Check to flush.
+    inline bool Flush() const;
+
+    ///
+    /// \brief Check to compress.
+    inline bool Compress() const;
+
+    ///
+    /// \brief Get deflate-level.
+    inline unsigned DeflateLevel() const;
+
+    ///
+    /// \brief Check to compute the chunk-size automatically.
+    inline bool AutomaticChunkSize() const;
+
+    ///
+    /// \brief Get chunk size.
+    inline std::vector<hsize_t> ChunkSize() const;
+
+private:
+    bool m_overwrite = false;
+    bool m_flush = true;
+    bool m_compress = false;
+    unsigned m_deflate_level = 0;
+    std::vector<hsize_t> m_chunk_size = {};
 };
 
 ///
@@ -91,15 +207,31 @@ inline std::vector<size_t> getShape(const File& file, const std::string& path);
 /// \param file Writeable opened file
 /// \param path Path of the DataSet
 /// \param data Data to write
-/// \param write_options Write mode, e.g. DumpMode or Compression
+/// \param mode Write mode
 ///
 /// \return The newly created DataSet
 ///
-template <class T, class... Args>
+template <class T>
 inline DataSet dump(File& file,
                     const std::string& path,
                     const T& data,
-                    Args... write_options);
+                    DumpMode mode = DumpMode::Create);
+
+///
+/// \brief Write object (templated) to a (new) DataSet in an open HDF5 file.
+///
+/// \param file Writeable opened file
+/// \param path Path of the DataSet
+/// \param data Data to write
+/// \param options Dump options
+///
+/// \return The newly created DataSet
+///
+template <class T>
+inline DataSet dump(File& file,
+                    const std::string& path,
+                    const T& data,
+                    const DumpOptions& options);
 
 ///
 /// \brief Write a scalar to a (new, extendible) DataSet in an open HDF5 file.
@@ -116,6 +248,58 @@ inline DataSet dump(File& file,
                     const std::string& path,
                     const T& data,
                     const std::vector<size_t>& idx);
+
+///
+/// \brief Write a scalar to a (new, extendible) DataSet in an open HDF5 file.
+///
+/// \param file opened File (has to be writeable)
+/// \param path path of the DataSet
+/// \param data the data to write
+/// \param idx the indices to which to write
+///
+/// \return The newly created DataSet
+///
+template <class T>
+inline DataSet dump(File& file,
+                    const std::string& path,
+                    const T& data,
+                    const std::initializer_list<size_t>& idx);
+
+///
+/// \brief Write a scalar to a (new, extendible) DataSet in an open HDF5 file.
+///
+/// \param file opened File (has to be writeable)
+/// \param path path of the DataSet
+/// \param data the data to write
+/// \param idx the indices to which to write
+/// \param options Dump options
+///
+/// \return The newly created DataSet
+///
+template <class T>
+inline DataSet dump(File& file,
+                    const std::string& path,
+                    const T& data,
+                    const std::vector<size_t>& idx,
+                    const DumpOptions& options);
+
+///
+/// \brief Write a scalar to a (new, extendible) DataSet in an open HDF5 file.
+///
+/// \param file opened File (has to be writeable)
+/// \param path path of the DataSet
+/// \param data the data to write
+/// \param idx the indices to which to write
+/// \param options Dump options
+///
+/// \return The newly created DataSet
+///
+template <class T>
+inline DataSet dump(File& file,
+                    const std::string& path,
+                    const T& data,
+                    const std::initializer_list<size_t>& idx,
+                    const DumpOptions& options);
 
 ///
 /// \brief Load entry "(i,j)" from a rank-two DataSet in an open HDF5 file to a scalar.
@@ -147,16 +331,34 @@ inline T load(const File& file, const std::string& path);
 /// \param path Path of the DataSet
 /// \param key Name of the attribute
 /// \param data Data to write
-/// \param write_options Write mode, e.g. DumpMode
+/// \param mode Write mode
 ///
 /// \return The newly created DataSet
 ///
-template <class T, class... Args>
+template <class T>
 inline Attribute dump_attr(File& file,
                            const std::string& path,
                            const std::string& key,
                            const T& data,
-                           Args... write_options);
+                           DumpMode mode = DumpMode::Create);
+
+///
+/// \brief Write object (templated) to a (new) Attribute in an open HDF5 file.
+///
+/// \param file Writeable opened file
+/// \param path Path of the DataSet
+/// \param key Name of the attribute
+/// \param data Data to write
+/// \param options Dump options
+///
+/// \return The newly created DataSet
+///
+template <class T>
+inline Attribute dump_attr(File& file,
+                           const std::string& path,
+                           const std::string& key,
+                           const T& data,
+                           const DumpOptions& options);
 
 ///
 /// \brief Load a Attribute in an open HDF5 file to an object (templated).
@@ -174,6 +376,7 @@ inline T load_attr(const File& file, const std::string& path, const std::string&
 
 #include "h5easy_bits/H5Easy_Eigen.hpp"
 #include "h5easy_bits/H5Easy_misc.hpp"
+#include "h5easy_bits/H5Easy_public.hpp"
 #include "h5easy_bits/H5Easy_scalar.hpp"
 #include "h5easy_bits/H5Easy_vector.hpp"
 #include "h5easy_bits/H5Easy_xtensor.hpp"
--- a/include/highfive/H5Easy.hpp
+++ b/include/highfive/H5Easy.hpp
@@ -107,7 +107,10 @@ public:
     /// \brief Dump-settings
     /// \param Any of DumpMode, Flush, Compression in arbitrary number and order.
     template <class... Args>
-    DumpOptions(Args... args);
+    DumpOptions(Args... args)
+    {
+        set(args...);
+    }
 
     ///
     /// \brief Set setting.
@@ -117,7 +120,7 @@ public:
     ///
     /// \brief Set setting.
     /// \param flush: Flush.
-    inline void set(Flush flush);
+    inline void set(Flush mode);
 
     ///
     /// \brief Set setting.
@@ -153,27 +156,27 @@ public:
 
     ///
     /// \brief Check to overwrite.
-    inline bool Overwrite() const;
+    inline bool isOverwrite() const;
 
     ///
     /// \brief Check to flush.
-    inline bool Flush() const;
+    inline bool isFlush() const;
 
     ///
     /// \brief Check to compress.
-    inline bool Compress() const;
+    inline bool isCompress() const;
 
     ///
     /// \brief Get deflate-level.
-    inline unsigned DeflateLevel() const;
+    inline unsigned getDeflateLevel() const;
 
     ///
     /// \brief Check to compute the chunk-size automatically.
-    inline bool AutomaticChunkSize() const;
+    inline bool isAutomaticChunkSize() const;
 
     ///
     /// \brief Get chunk size.
-    inline std::vector<hsize_t> ChunkSize() const;
+    inline std::vector<hsize_t> getChunkSize() const;
 
 private:
     bool m_overwrite = false;
--- a/include/highfive/H5Easy.hpp
+++ b/include/highfive/H5Easy.hpp
@@ -255,7 +255,7 @@ inline DataSet dump(File& file,
 ///
 /// \brief Write a scalar to a (new, extendable) DataSet in an open HDF5 file.
 ///
-/// \param file opened File (has to be writeable)
+/// \param file open File (has to be writeable)
 /// \param path path of the DataSet
 /// \param data the data to write
 /// \param idx the indices to which to write
--- a/include/highfive/H5Easy.hpp
+++ b/include/highfive/H5Easy.hpp
@@ -39,6 +39,7 @@
 
 namespace H5Easy {
 
+using HighFive::Attribute;
 using HighFive::AtomicType;
 using HighFive::Chunking;
 using HighFive::DataSet;
@@ -85,12 +86,12 @@ inline size_t getSize(const File& file, const std::string& path);
 inline std::vector<size_t> getShape(const File& file, const std::string& path);
 
 ///
-/// \brief Write scalar/string to a new DataSet in an open HDF5 file.
+/// \brief Write object (templated) to a (new) DataSet in an open HDF5 file.
 ///
 /// \param file Writeable opened file
 /// \param path Path of the DataSet
 /// \param data Data to write
-/// \param write_options Write mode, any combination of DumpMode and Compression
+/// \param write_options Write mode, e.g. DumpMode or Compression
 ///
 /// \return The newly created DataSet
 ///
@@ -138,6 +139,36 @@ inline T load(const File& file, const std::string& path, const std::vector<size_
 ///
 template <class T>
 inline T load(const File& file, const std::string& path);
+
+///
+/// \brief Write object (templated) to a (new) Attribute in an open HDF5 file.
+///
+/// \param file Writeable opened file
+/// \param path Path of the DataSet
+/// \param key Name of the attribute
+/// \param data Data to write
+/// \param write_options Write mode, e.g. DumpMode
+///
+/// \return The newly created DataSet
+///
+template <class T, class... Args>
+inline Attribute dump_attr(File& file,
+                           const std::string& path,
+                           const std::string& key,
+                           const T& data,
+                           Args... write_options);
+
+///
+/// \brief Load a Attribute in an open HDF5 file to an object (templated).
+///
+/// \param file opened File (has to be writeable)
+/// \param path path of the DataSet
+/// \param key Name of the attribute
+///
+/// \return the read data
+///
+template <class T>
+inline T load_attr(const File& file, const std::string& path, const std::string& key);
 
 }  // namespace H5Easy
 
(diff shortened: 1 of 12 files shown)


tdegeus:
It might be good to check from conflicting settings. Current the user could for example specify:
```cpp
dump(file, path, data, H5Easy::Compression::None, H5Easy::Compression::High)
```
without getting an error

tdegeus:
@alkino Are you willing to review?

tdegeus:
Superseeds #288

tdegeus:
Can someone review the contribution of the core library? @matz-e @alkino @ferdonline

spacescientist:
Thank you very much for your work on this; sorry I haven't been on github for quite a while.

My main feedback is that I actually completely agree with your comment there: https://github.com/BlueBrain/HighFive/issues/301#issuecomment-616394989

Imagine some people that are collaborating and creating HDF files using different tools and languages, they might still *require the exact same compression level 1-9 for all files created within the collaboration*. With "Low", "Medium" and "High", that might not be possible for them though.

I strongly think it would be preferable to preserve this HDF5 standard. It'll be clear for everyone using HDF5 (facilitating the adoption of HighFive), give full flexibility to the users, and avoid creating a somewhat vague and unnecessary new terminology.

--

Or maybe have both if you want to keep the Compression enum, by specifying the values for the enums?
```cpp
enum Compression
{
    None = 0, /*!< No compression. (default) */
    Medium = 5, /*!< Medium compression level. */
    High = 9 /*!< High compression level. */
};
```
And then adapt the code to accept either values from 1-9 or such an enum, interpreting its values directly as being the deflate_level.
```cpp
inline void set(unsigned int mode)
{
	if(mode > 0)
	{
		deflate_level = mode;
		compress = true;
	}
	else compress = false;
}
```
I've done a mock test (adapting these lines of code ^, not the full thing), and it then accepts things like:
```cpp
Compression choice = Medium;
set(choice); 
// then printing deflate_level and compress give: 5 & 1

int defl = 8;
set(defl);
// then printing deflate_level and compress give: 8 & 1

set(7);
// then printing deflate_level and compress give: 7 & 1

Compression choice2 = None;
set(choice2);
// then printing deflate_level and compress give: 0 & 0
```

tdegeus:
@spacescientist I agree that using a standard that is there, or at least allowing it, makes sense. The challenge in the implementation is that I think it is nicest to allow a mix of options in arbitrary order. That way the user doesn't have to remember the specificities, and the code can be adapted later. It also has a benefit: one could have `H5Easy::Compression::High` and maybe something like `H5Easy::CompressionLevel(8)` as options?

spacescientist:
@tdegeus But allowing for an arbitrary ordering really is a separate issue, right? 

If you really want that, maybe make all options inherit from an abstract base class `H5Easy::OptionBase`, and then have an option handler in the dump method, for instance with a set of boolean flags (one for each family of options) which are all false by default. Set flagX to true when optionX is provided, and make it complain if it realises that it is given yet another optionX (i.e. if flag was already set to true: give an error).

In such a setting, `H5Easy::Compression::High` and `H5Easy::CompressionLevel(8)` should trigger the same flag (as they should belong to the same family of options); only one would be allowed at a time, and the ordering with respect to other options wouldn't matter thanks to polymorphism.
```cpp
// all would be valid:
dump(file, path, data, H5Easy::Compression::High, H5Easy::OtherOption);
dump(file, path, data, H5Easy::OtherOption, H5Easy::Compression::High);
dump(file, path, data, H5Easy::CompressionLevel(8), H5Easy::OtherOption);
dump(file, path, data, H5Easy::OtherOption, H5Easy::CompressionLevel(8));
```

I'm a bit concerned about the number of arguments though, should there be many kinds of possible options. Especially if the user only cares about one of them. How many do you have in mind?

Maybe rather make the user select all their options first, and then give that to `dump()`? For instance via an `H5Easy::Options` class which would handle the options (and check their uniqueness) inside its constructor and/or by applying an easy-to-use `.add_option()` method any number of times?

```cpp
H5Easy::Options options(H5Easy::SomeOption, H5Easy::OtherOption, ...);
// ^ either with constructor to be adapted and which would accept any options in any order
// and would take among other things either the deflate_level or the enum (if you want to keep these enums as well).

// and/or also give the possibility to add further options via a method, e.g.:
options.add_option(H5Easy::CompressionLevel(8));

// then, one could simply use:
dump(file, path, data, options); // dump with custom options
```
```cpp
// Making options an optional argument, this v would then still be valid:
dump(file, path, data); // dump with default options
```

tdegeus:
I like the idea of having a base-class for dump options, as it indeeds as the possibility to pre-combine options for many dump operations. Maybe then one does not even need the variadic options? If one wants a one-liner one can simply do it with the constructor of `H5Easy::DumpOptions`. Great idea! Thanks

spacescientist:
Yes, I think so. Glad if that can help!

Now, just going back to `deflate_level`: I really think that no more than one class/constructor is needed. `H5Easy::Compression::High` and `H5Easy::CompressionLevel(8)` do not need to be considered as separate options in reality since enums can be given a specific value and reinterpreted. See again https://github.com/BlueBrain/HighFive/pull/337#issuecomment-638090633.

That being said, refering again to https://github.com/BlueBrain/HighFive/issues/301#issuecomment-592158639: it might be actually better to set all the compression options at once as part of the same object: not just the `deflate_level`, but also the `chunking` etc. 

In any case, what I really wanted to stress is that for setting the deflate_level, integers (following the HDF5 standard) and enums (which may be introduced for convenience) might use the exact same code.

tdegeus:
What about this API @spacescientist ?
```cpp
#include <vector>
#include <type_traits>
#include <iostream>

enum class DumpMode {
    Create = 0,
    Overwrite = 1
};

enum class DumpCompression {
    None = 0,
    Medium = 5,
    High = 9
};

class DumpOptions
{
public:
    DumpOptions() = default;

    template <class... Args>
    DumpOptions(Args... args)
    {
        set(args...);
    }

    inline void set(DumpMode mode)
    {
    }

    inline void set(DumpCompression mode)
    {
    }

    template <class T, class... Args>
    inline void set(T arg, Args... args)
    {
        set(arg);
        set(args...);
    }

    inline void setDeflateLevel(int level)
    {
    }

    template <class T>
    inline void setChunkSize(const std::vector<T>& shape)
    {
    }

    inline void setChunkSize(std::initializer_list<size_t> shape)
    {
    }
};


template <class T>
inline void dump(const T& data, const DumpOptions& options)
{
}

template <class T>
inline void dump(const T& data, DumpMode mode = DumpMode::Create)
{
    return dump(data, DumpOptions(mode));
}

int main()
{
    DumpOptions options;
    options.setChunkSize({10, 10});

    dump(10);
    dump(10, DumpMode::Overwrite);
    dump(10, DumpOptions(DumpMode::Overwrite, DumpCompression::High));
    dump(10, options);

    return 0;
}
```

tdegeus:
* Another option that crossed my mind for the API would be change the overload for `DumpMode` with and `unsigned int`. Then one could enumerate clearly to allow from input like `DumpMde::Overwrite + DumpCompressionHigh`, or even something like `29` (`20` for Overwrite and `9` for Compression). But ultimately I think that the syntax is a bit obscure and that too many internals are exposed, causing future changes to result in less and less readable code. 
* What I would have liked was a single overload, but I without loosing backward compatibility. It would have been great to have `DumpMode` (or oven `DumpMode::Overwrite`) derive from `DumpOptions`. But I don't think this could be done without requiring the user to construct, requiring parentheses.

tdegeus:
@spacescientist I've integrated the new API. What do you think?

tdegeus:
Can someone at minimum review the contribution to `H5Attribute.hpp` and `bits/H5Attribute_misc.hpp`? @matz-e @alkino @ferdonline

include/highfive/H5Easy.hpp
```diff
@@ -39,20 +39,151 @@
...
+/// \brief Set compression-level for written DataSets.
+class CompressionLevel
+{
+public:
+    CompressionLevel(unsigned deflate_level);
+
+    inline unsigned get() const;
+
+private:
+    unsigned m_deflate_level;
```
matz-e, line 88: Compression or deflation? The naming here is inconsistent.

include/highfive/H5Easy.hpp
```diff
@@ -39,20 +39,151 @@
...
+    ///
+    /// \brief Set settings.
+    /// \param Any of DumpMode, Flush, Compression, CompressionLevel in arbitrary number and order.
+    template <class T, class... Args>
+    inline void set(T arg, Args... args);
+
+    ///
+    /// \brief Set deflate level.
+    /// \param level: deflate level (0-9).
+    inline void setDeflateLevel(unsigned level);
```
matz-e, line 144: See above… deflation or compression, pick one, please.

include/highfive/H5Easy.hpp
```diff
@@ -39,20 +39,151 @@
...
+    ///
+    /// \brief Check to flush.
+    inline bool isFlush() const;
+
+    ///
+    /// \brief Check to compress.
+    inline bool isCompress() const;
+
+    ///
+    /// \brief Get deflate-level.
```
matz-e, line 170:
```suggestion
    /// \brief Get deflation level.
```

include/highfive/h5easy_bits/H5Easy_misc.hpp
```diff
@@ -59,16 +48,144 @@ inline void createGroupsToDataSet(File& file, const std::string& path) {
...
+// get a opened DataSet: nd-array
+template <class T>
+inline DataSet init_dataset(File& file,
+                            const std::string& path,
+                            const std::vector<size_t>& shape,
+                            const DumpOptions& options)
+{
+    if (!file.exist(path)) {
+        detail::createGroupsToDataSet(file, path);
+        if (!options.isCompress()) {
```
matz-e, line 84: Naive question: can I have a chuncked dataset without compression?

include/highfive/h5easy_bits/H5Easy_public.hpp
```diff
@@ -0,0 +1,221 @@
...
+{
+    m_chunk_size = std::vector<hsize_t>(shape.begin(), shape.end());
+}
+
+inline void DumpOptions::setChunkSize(std::initializer_list<size_t> shape)
+{
+    m_chunk_size = std::vector<hsize_t>(shape.begin(), shape.end());
+}
+
+inline bool DumpOptions::isOverwrite() const
```
matz-e, line 94:
```suggestion
inline bool DumpOptions::overwrite() const
```
my preference

include/highfive/h5easy_bits/H5Easy_public.hpp
```diff
@@ -0,0 +1,221 @@
...
+{
+    return m_overwrite;
+}
+
+inline bool DumpOptions::isFlush() const
```
matz-e, line 99:
```suggestion
inline bool DumpOptions::flush() const
```

include/highfive/h5easy_bits/H5Easy_public.hpp
```diff
@@ -0,0 +1,221 @@
...
+{
+    return m_flush;
+}
+
+inline bool DumpOptions::isCompress() const
```
matz-e, line 104:
```suggestion
inline bool DumpOptions::compress() const
```

include/highfive/h5easy_bits/H5Easy_public.hpp
```diff
@@ -0,0 +1,221 @@
...
+{
+    return m_compress;
+}
+
+inline unsigned DumpOptions::getDeflateLevel() const
+{
+    return m_deflate_level;
+}
+
+inline bool DumpOptions::isChunkSize() const
```
matz-e, line 114:
```suggestion
inline bool DumpOptions::isChunked() const
```
? What exactly is this supposed to do?

include/highfive/H5Easy.hpp
```diff
@@ -39,20 +39,151 @@
...
+enum class Compression
+{
+    False = 0, /*!< No compression. */
+    Medium = 5, /*!< Medium compression (deflate level 5). */
+    High = 9 /*!< High compression (deflate level 9). */
+};
+
+///
+/// \brief Set compression-level for written DataSets.
+class CompressionLevel
```
matz-e, line 80: I'm wondering if this and `Compression` could not be folded into one, simpler class/enum; Numerical values count for compression level, certain values could be queried by i.e.:
```
bool compress() const { return m_level > 0; };
```
etc…

You could use a static construction method to get medium/high compression.

include/highfive/h5easy_bits/H5Easy_public.hpp
tdegeus: To distinct if the chunk-size if prescribed by the user, or if the dump routines should proceed to automatically choose it.

include/highfive/H5Easy.hpp
tdegeus: This I like. A question is if `deflate_level = 0` means no compression, also in the underlying library? Or if the user should be able to really specify `deflate_level = 0`.

src/examples/easy_attribute.cpp
```diff
@@ -0,0 +1,40 @@
...
+    H5Easy::File file("example.h5", H5Easy::File::Overwrite);
+
+    std::vector<double> A = {1.0, 2.0, 3.0};
+    std::string desc = "This is an important dataset.";
+    double T = 1.234;
+
+    H5Easy::dump(file, "/path/to/A", A);
+    H5Easy::dump_attr(file, "/path/to/A", "description", desc);
+    H5Easy::dump_attr(file, "/path/to/A", "temperature", T);
+
```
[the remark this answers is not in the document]
tdegeus: Will do!

tests/unit/tests_high_five_easy.cpp
```diff
@@ -99,16 +125,16 @@ BOOST_AUTO_TEST_CASE(H5Easy_vector3d)
...
+    a[0][0].push_back(0);
+    a[0][0].push_back(1);
+    a[0][1].push_back(2);
+    a[0][1].push_back(3);
+    a[1][0].push_back(4);
+    a[1][0].push_back(5);
+    a[1][1].push_back(6);
+    a[1][1].push_back(7);
+    a[2][0].push_back(8);
+    a[2][0].push_back(9);
```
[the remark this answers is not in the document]
tdegeus: Do you mean: use less values? Or construct differently?

include/highfive/h5easy_bits/H5Easy_public.hpp
tdegeus: OK, will update the PR.

include/highfive/h5easy_bits/H5Easy_public.hpp
tdegeus: OK, will update the PR.

include/highfive/h5easy_bits/H5Easy_public.hpp
tdegeus: OK, will update the PR.

include/highfive/h5easy_bits/H5Easy_public.hpp
tdegeus: In terms of naming, the `ChunkSize` can now also be used when writing extendable DataSets (to overwrite the default `10` along all dimensions). These DataSets are by definition chunked.

tdegeus:
Thanks for your review @matz-e . It is indeed a substantial PR, but at the same time it brings a lot of new things to H5Easy: compression, various customisations, and reading/writing of attributes.

tests/unit/tests_high_five_easy.cpp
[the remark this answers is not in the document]
matz-e: Sorry, should have been more clear. This is just a change in whitespace that could have been skipped…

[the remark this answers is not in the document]
matz-e: Now that you ask, though… I'm wondering if this could be constructed with a nested initializer list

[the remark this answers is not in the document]
tdegeus: Good idea! Will do.

include/highfive/H5Easy.hpp
tdegeus: I get the answer is `deflate_level = 0` should switch of shuffling as well: https://github.com/Unidata/netcdf-c/issues/23

include/highfive/h5easy_bits/H5Easy_public.hpp
tdegeus: In the end I think that `isChunked` is at least as good. So I changed it.

include/highfive/h5easy_bits/H5Easy_public.hpp
tdegeus: Done

include/highfive/h5easy_bits/H5Easy_public.hpp
tdegeus: Done

include/highfive/h5easy_bits/H5Easy_public.hpp
tdegeus: Done

include/highfive/H5Easy.hpp
tdegeus: About converting to one class, I love the idea, but I don't think it can be done though. C++ does not have static constructors, and having `enum` members would ruin the overload resolution.

Unless I'm overlooking something @matz-e ?

tests/unit/tests_high_five_easy.cpp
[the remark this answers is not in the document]
tdegeus: Done

src/examples/easy_attribute.cpp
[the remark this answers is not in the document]
tdegeus: I made a third suggestion.

include/highfive/H5Easy.hpp
matz-e: Just an idea or two:
```
#include <iostream>

class SomeLevel {
    size_t level_;
public:
    SomeLevel(size_t l) : level_(l) {};

    static const SomeLevel High;
    static SomeLevel Medium() {
        return SomeLevel{4};
    };

    size_t get() const { return level_; };
};

const SomeLevel SomeLevel::High{8};

int main(void) {
    auto foo = SomeLevel::High;
    std::cout << foo.get() << "\n";
    auto bar = SomeLevel::Medium();
    std::cout << bar.get() << "\n";
}
```

If there should really be a distinction between 0 and not setting any compression, I would think that -1/limits::max<unsigned> could be used?

tdegeus: Beautiful! Thanks a lot!

tdegeus: This really simplifies not only the implementation, but also the API!

include/highfive/H5Easy.hpp
```diff
@@ -105,6 +252,58 @@ inline DataSet dump(File& file,
...
+/// \param options Dump options
+///
+/// \return The newly created DataSet
+///
+template <class T>
+inline DataSet dump(File& file,
+                    const std::string& path,
+                    const T& data,
+                    const std::initializer_list<size_t>& idx,
+                    const DumpOptions& options);
```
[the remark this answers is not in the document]
tdegeus: (Sorry, this comment was hidden before)
I don't get exactly what you mean. Can you reformulate?

include/highfive/h5easy_bits/H5Easy_public.hpp
```diff
@@ -0,0 +1,221 @@
...
+}
+
+inline unsigned CompressionLevel::get() const
+{
+    return m_deflate_level;
+}
+
+inline void DumpOptions::set(DumpMode mode)
+{
+    if (mode == DumpMode::Create) {
```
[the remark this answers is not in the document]
tdegeus: Sure!

include/highfive/H5Easy.hpp
tdegeus: Well.. If you look where it is used 
```cpp
props.add(Deflate(options.getDeflateLevel()));
```
deflate makes sense (for at least some of the settings). But at the same time for most users I would say that "compression" is more logical. So my preference is what I did. But it would be good to hear your preference.

include/highfive/H5Easy.hpp
tdegeus: (See comment above)

include/highfive/h5easy_bits/H5Easy_misc.hpp
tdegeus: I suppose (but honestly this lies outside my normal use case). Anyways I could easily allow it

include/highfive/H5Easy.hpp
tdegeus: @matz-e This leads to a a compiler error "multiple definition of ". Do you know if and how that can be solved for a header-only library?

tdegeus:
@matz-e I decided to simplify the API, because I don't think the previous problems can be overcome. One can now use

```cpp
H5Easy::Compression()
H5Easy::Compression(true)
H5Easy::Compression(8)
```

Please tell me what you think. Thanks!

tdegeus:
@matz-e Are we good to go?

include/highfive/H5Easy.hpp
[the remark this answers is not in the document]
tdegeus: Ah sorry now I get what you mean. No it breaks backward compatibility, which I don't think is warranted by these new 'expert' options

matz-e:
> @matz-e I decided to simplify the API, because I don't think the previous problems can be overcome. One can now use
> 
> ```c++
> H5Easy::Compression()
> H5Easy::Compression(true)
> H5Easy::Compression(8)
> ```
> 
> Please tell me what you think. Thanks!

Thanks, looks very nice that way IMO. I got some other stuff on my plate right now, I'll try to give it another look tomorrow.

tdegeus:
Would be great @matz-e !

include/highfive/h5easy_bits/H5Easy_vector.hpp
```diff
@@ -28,29 +28,46 @@ using HighFive::details::type_of_array;
 template <typename T>
 struct io_impl<T, typename std::enable_if<is_vector<T>::value>::type> {
 
-    static DataSet dump(File& file, const std::string& path, const T& data) {
-        using type_name = typename type_of_array<T>::type;
-        detail::createGroupsToDataSet(file, path);
-        DataSet dataset = file.createDataSet<type_name>(path, DataSpace::From(data));
+    inline static DataSet dump(File& file,
+                        const std::string& path,
```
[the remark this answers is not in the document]
tdegeus: Thanks, I'll change. This the reason why I don't like the selected style: indentation should not have to change when something is renamed. That is one on the reasons many styles don't align for instance at `=` in a sequence of assignment operations. On top of that I personally find the chosen style very difficult to read, code is often dangling almost of the right of the line.

include/highfive/H5Easy.hpp
```diff
@@ -128,10 +318,59 @@ inline T load(const File& file, const std::string& path, const std::vector<size_
...
+/// \param file Writeable opened file
+/// \param path Path of the DataSet
+/// \param key Name of the attribute
+/// \param data Data to write
+/// \param mode Write mode
+///
+/// \return The newly created DataSet
+///
+template <class T>
+inline Attribute dump_attr(File& file,
```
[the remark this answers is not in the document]
tdegeus: Fine!

include/highfive/h5easy_bits/H5Easy_Eigen.hpp
```diff
@@ -24,6 +24,30 @@ struct io_impl<
...
+                typename std::decay<T>::type::Scalar,
+                std::decay<T>::type::RowsAtCompileTime,
+                std::decay<T>::type::ColsAtCompileTime,
+                std::decay<T>::type::ColsAtCompileTime == 1 ? Eigen::ColMajor : Eigen::RowMajor,
+                std::decay<T>::type::MaxRowsAtCompileTime,
+                std::decay<T>::type::MaxColsAtCompileTime>,
+            0,
+            Eigen::InnerStride<1>>;
+
+        using col_major = Eigen::Map<Eigen::Array<
```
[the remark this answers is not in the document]
tdegeus: Probably safer to use `std::decay<T>::type` everywhere. I'll change.

include/highfive/h5easy_bits/H5Easy_scalar.hpp
```diff
@@ -62,89 +85,49 @@ struct io_impl {
...
         detail::createGroupsToDataSet(file, path);
         std::vector<size_t> shape = idx;
         const size_t unlim = DataSpace::UNLIMITED;
         std::vector<size_t> unlim_shape(idx.size(), unlim);
-        std::vector<hsize_t> chuncks(idx.size(), 10);
+        std::vector<hsize_t> chunks(idx.size(), 10);
+        if (options.isChunked()) {
+            chunks = options.getChunkSize();
+            if (chunks.size() != idx.size()) {
+                throw error(file, path, "H5Easy::dump: Incorrect rank ChunkSize");
```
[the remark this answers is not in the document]
tdegeus: For me `rank` always reminds me of a tensor. I'll use `dimension` instead as it is part of the API as well.

tdegeus:
Thanks @matz-e , that should do it.

include/highfive/h5easy_bits/H5Easy_Eigen.hpp
```diff
@@ -58,66 +84,68 @@ struct io_impl<
...
+        if (options.flush()) {
+            file.flush();
+        }
+        return attribute;
+    }
+
+    inline static T loadAttribute(const File& file,
+                                  const std::string& path,
+                                  const std::string& key) {
+        DataSet dataset = file.getDataSet(path);
```
ferdonline, line 133: Some of these methods may return a ref. To avoid copies and potentially improve performance, consider declaring const-ref when applicable.

include/highfive/h5easy_bits/H5Easy_misc.hpp
```diff
@@ -59,16 +48,146 @@ inline void createGroupsToDataSet(File& file, const std::string& path) {
...
+            "H5Easy: path exists, but does not correspond to a Dataset. Dump not possible.");
+    }
+}
 
-inline size_t getSize(const File& file, const std::string& path) {
-    return file.getDataSet(path).getElementCount();
+// get a opened DataSet: nd-array
+template <class T>
+inline DataSet initDataset(File& file,
+                           const std::string& path,
```
ferdonline, line 78: At a glance I see four implementations that share similar concepts. Not a blocker, I'm more in favour of unified solutions if that could be applied here.

tdegeus: I'm not against such a clean up. One would have to see if it becomes something readable/maintainable. Maybe you could open a PR when this is PR is merged?

include/highfive/h5easy_bits/H5Easy_Eigen.hpp
tdegeus: Do you mean using `const DataSet dataset = file.getDataSet(path);` etc. ? Here I trust your expertise. But let's merge this first and do a separate PR. This make the extra API features public. Also, once this PR is merged I can finish https://github.com/BlueBrain/HighFive/pull/343 . Thereafter there are only internal changes.

A related point is that a minor release could be considered soon, as there are quite some new features (also in https://github.com/BlueBrain/HighFive/pull/346 ) and it would be great to make them available to (conda) package users.

tdegeus:
Thanks @ferdonline !

tdegeus:
Do you also approve @matz-e ?

include/highfive/h5easy_bits/H5Easy_Eigen.hpp
matz-e: None of the actual data is actually copied, though? I think in that case the penalty here is negligible, even though I agree that `const Dataset&` would be more appropriate.

tdegeus: Let's do it in a separate PR!

spacescientist:
That's great; thanks a lot @tdegeus!

Issue #314, opened 2020-03-11
Can I get the path to a DataSet?


tdegeus:
Can I get the path to a DataSet from a `HighFive::DataSet` object?

alkino:
I think no because it is possible to have several path to the same DataSet.

tdegeus:
Make senses. I guess the only possible way would be if a `HighFive::DataSet` instance would keep memory of the path with which is was called. But I see no urgent reason for it.

KerimMatlab:
I think this could be done via [H5I_GET_NAME(obj_id, name, size)](https://portal.hdfgroup.org/display/HDF5/H5I_GET_NAME)

> I think no because it is possible to have several path to the same DataSet.

this is said in documentation that folows the link above but at least one path is returned. Ttis may be enough.

I do not provide any code for this task because I'am beginner at programming (it is only few months that I work with C/C++). It is hard to change the source code without knowing the core of library.

alkino:
This is the right function to do it, thanks I will give a try

tdegeus:
Thanks @KerimMatlab ! And thanks @alkino for giving it a try !

KerimMatlab:
@tdegeus I added this functionality to the [pull request ](https://github.com/BlueBrain/HighFive/pull/346)
#314

matz-e:
Closed in #314.

PR #346, opened 2020-05-27
Add method to retrieve path from Node or Dataset (issue #314). Add method "rename"

KerimMatlab:
Sometimes you need to retrieve the name and the path of the node (File/Group) or dataset and attribute.
For `File/Group/Dataset` hdf5 uses [function](https://portal.hdfgroup.org/display/HDF5/H5I_GET_NAME): `ssize_t H5Iget_name( hid_t obj_id, char *name, size_t size )` and this returnes path to these objects
For attribute it uses [another function](https://portal.hdfgroup.org/display/HDF5/H5A_GET_NAME): `ssize_t H5Aget_name(hid_t attr_id, size_t buf_size, char *buf )`

To check the implemented function you can run:
```C++
#include <H5File.hpp>
#include <H5Group.hpp>
#include <H5DataSet.hpp>
#include <H5DataSpace.hpp>
#include <H5Attribute.hpp>

using namespace HighFive;

int main(int argc, char *argv[])
{
    /* We are going to create group in root directory then add
     * dataset to this group and attach attribute to the dataset.
     * Then we are trying to get path to the root, group dataset
     * and the name of the dataset.
     * Secondly we will move dataset with attached attribute to
     * some destination path. Every '/' sign means that a group
     * will be created there (it is done automatically).
     * When we have moved the dataset we can check if dataset
     * object is still valid? We do this by creating second
     * attribute. There is a tricky part at the end when we
     * move Group with its dataset */

    // Create a new file using the default property lists.
    HighFive::File file("names.h5", File::ReadWrite | File::Create | File::Truncate);

    // Create a group
    HighFive::Group group = file.createGroup("group");

    // Create a dummy dataset of one single integer
    DataSet dataset = group.createDataSet("data", DataSpace(1), AtomicType<int>());
    dataset.write(100);

    // Now let's add a attribute to this dataset
    std::string string_list("very important Dataset !");
    Attribute attribute = dataset.createAttribute<std::string>("attribute", DataSpace::From(string_list));
    attribute.write(string_list);

    // Get path and names
    std::cout << "root path: \t" << file.getPath() << std::endl;
    std::cout << "group path: \t" << group.getPath() << std::endl;
    std::cout << "dataset path: \t" << dataset.getPath() << std::endl;
    std::cout << "attribute name: \t" << attribute.getName() << std::endl;
    std::cout << std::endl;

    // move dataset with its attribute to another destination path
    group.moveObject(file, "data", "/NewGroup/SubGroup/movedData");

    // as you can see to reach destination path new groups were created as well
    std::cout << "dataset new path: \t" << dataset.getPath() << std::endl;
    std::cout << std::endl;

    // we can still use moved dataset
    // let's create new attribute
    Attribute attributeNew = dataset.createAttribute<std::string>("attributeNew", DataSpace::From(string_list));
    attribute.write(string_list);
    std::cout << "attribute new name: \t" << attributeNew.getName() << std::endl;
    std::cout << std::endl;

    // move the folder with its content to other place
    file.moveObject("/NewGroup/SubGroup", "/FinalDestination");

    // here is the important moment. Old 'dataset' variable tells us
    // that dataset directory wasn't changed
    std::cout << "dataset new path wasn't changed: \t" << dataset.getPath() << std::endl;
    std::cout << std::endl;

    // but actually it was moved we just need to update variable
    std::cout << "actually it was moved we just need to update it: \t" << file.getDataSet("/FinalDestination/movedData").getPath() << std::endl;
    std::cout << std::endl;

    /* The conclusion: if you move a Group always update the varibles
     * to its content :) */

    file.flush();
}
```


KerimMatlab:
Also I added the method `moveObject(...)` wich is based on `herr_t H5Lmove( hid_t src_loc_id, const char *src_name, hid_t dest_loc_id, const char *dest_name, hid_t lcpl_id, hid_t lapl_id )` hdf5 library function
Only `File` and `Group` has this method.

tdegeus:
A first question is : could you add some tests in `HighFive/tests/unit`. That would help verification as well as grasping the functionality.

KerimMatlab:
@tdegeus just added [here](https://github.com/KerimMatlab/HighFive/blob/master/tests/unit/test_getPath_and_moveObject.cpp)

tdegeus:
Thanks! Do you think the tests could be added to `tests_high_five_base.cpp`? Then they would be immediately included in the continuous integration.

KerimMatlab:
@tdegeus actually I don't know how test should look like
I just made simple example how to implement written function
If my test is badly written probably I could change it. At least rewrite the comments to make it more clear and more readable probably

include/highfive/bits/H5Attribute_misc.hpp
```diff
@@ -27,6 +27,24 @@
 
 namespace HighFive {
 
+inline std::string Attribute::getAttributeName() const {
+    const size_t maxLength = 255;
+    char  buffer[maxLength + 1];
+    ssize_t retcode = H5Aget_name(
```
matz-e, line 33: Seems to me that this code repeats itself several times, save for this little function. Could probably be deduplicated? Also the naming could be shortened to `getName` and `getPath` without repeating the class name in the method name…

tests/unit/test_getPath_and_moveObject.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include <H5DataSet.hpp>
+#include <H5DataSpace.hpp>
+#include <H5Attribute.hpp>
+#include <H5PropertyList.hpp>
+
+using namespace HighFive;
+
+int main(int argc, char *argv[])
+{
+    QCoreApplication a(argc, argv);
```
matz-e, line 12: Seems Qt specific?

See the other tests in the directory and `CMakeLists.txt` for a better testing structure.

include/highfive/bits/H5Attribute_misc.hpp
KerimMatlab: Hi,
I absolutely agree with your proposition about changing the names of the methods. So now they are called `getName` and `getPath`
> Seems to me that this code repeats itself several times, save for this little function. Could probably be deduplicated?

I still don't fully understand how this library works because I have huge number of errors inside the project :)
I guess this is beacause one namespace is divided in many `.hpp` files and when I compile the code it works because those files are combined I think. 
But editing the code is difficult in such conditions. In this case creating new small functions that decrease the number of code lines would make the code less readable for me and probably for other developpers who probably would read this code.
So if you don't mind I would leave it as is.

tests/unit/test_getPath_and_moveObject.cpp
KerimMatlab: I just googled what is C++ tests... it is to comlicated for me now. Have not done this before. Also I don't know CMake, I use QMake (I'm beginner).
So I edited the example and put it to the [/src/examples](https://github.com/KerimMatlab/HighFive/tree/master/src/examples) folder. Here is this [example](https://github.com/KerimMatlab/HighFive/blob/master/src/examples/renaming_and_moving_objects.cpp). I hope it is clear and clean enough
I'm sorry I can't do tests now

tdegeus:
@KerimMatlab Nice features you introduced! 

I do have some suggestions to improve. If you agree I'll directly push them to the PR.

My comments are that:
* `getPath` and `getName` could be moved to `Object` to potentially widen their scope. Both `Attribute` and `DataSet` are derived from it.
* I would keep the API `file.moveObject("/old/absolute/path", "/new/absolute/path")` and `file.moveObject(group, "old/relative/path", "new/relative/path")`.
* I don't see a use for `getPath` and `getName` in `NodeTraits`, but maybe I'm overlooking something?

KerimMatlab:
Why there are three declined status (Review required, All checks have failed, Merging is blocked)?
It is my first time adding pull request. Can't understand what should I do

tdegeus:
The "1 failing check" is because the continuous integration (CI) is firing an error. In this project all compiler warning do this. You can click on details to check out the reason. Furthermore this project requires at least one reviewer to explicitly approve before the PR can be merged by somebody with the proper rights.

tdegeus:
Let me know if you're ok if I push some suggestions (we can always revert some of them)

KerimMatlab:
I'm sorry I didn't see your post
> * `getPath` and `getName` could be moved to `Object` to potentially widen their scope. Both `Attribute` and `DataSet` are derived from it.

At the beginning I wanted to do so. But then I have found that `File, Group and DataSet` inherit from `Object` and the `Object` has **protected** constructors. As I know there is no way to create object of type `Object` (using High Five) of course, original hdf5 library allows that). 
Such protected constructors protects us from making bugs but we pay for that by the impossibility of use Object. That is how I see that. And this is someone's decision I think, someone who know the library better. So I decided not to make `Object` constructor public.

> I would keep the API file.moveObject("/old/absolute/path", "/new/absolute/path") and file.moveObject(group, "old/relative/path", "new/relative/path").

Yes I see. I'm going to change it. 

> I don't see a use for getPath and getName in NodeTraits, but maybe I'm overlooking something?

Could you explain? You have `NodeTraits` typed object and you can't use those functions? Or you can't imagine `NodeTraits` would need these functionality? If so I haven't used `NodeTraits` but I know that `File` and `Group` inherits from this class and most method for `File` and `Group` are written there. It's just convinient to write methods for `File` and `Group` there.
By the way if a group is not empty and you move this group then all the objects from this group are moved as well! BUT you need to create new variables for those objects in this group after they were moved (old variables for those objects will tell you that their path wasn't changed). Maybe it is a something like a bug in hdf5 because in this case the old variable for group automatically changes its path.
Maybe I should expand the example to point this.

KerimMatlab:
> The "1 failing check" is because the continuous integration (CI) is firing an error. In this project all compiler warning do this. You can click on details to check out the reason. Furthermore this project requires at least one reviewer to explicitly approve before the PR can be merged by somebody with the proper rights.

Ok
We will be in touch. I don't understand exactly how project merges but I want that this functionality is included.

include/highfive/H5DataSet.hpp
```diff
@@ -31,6 +31,11 @@ class DataSet : public Object,
 
     const static ObjectType type = ObjectType::Dataset;
 
+    ///
+    /// \brief return the path to the current dataset
+    /// \return the path to the dataset
+    std::string getPath() const;
+
```
tdegeus, line 38: Move to `Object`: `DataSet`, `Group`, and `Attribute` derive from it: all would have this functionality

include/highfive/bits/H5Attribute_misc.hpp
```diff
@@ -27,6 +27,24 @@
 
 namespace HighFive {
 
+inline std::string Attribute::getName() const {
+    const size_t maxLength = 255;
+    char  buffer[maxLength + 1];
```
tdegeus, line 32:
```suggestion
    char buffer[maxLength + 1];
```

include/highfive/bits/H5Attribute_misc.hpp
```diff
@@ -27,6 +27,24 @@
...
+    ssize_t retcode = H5Aget_name(
+        _hid, static_cast<hsize_t>(maxLength) + 1, buffer);
```
tdegeus, line 34:
```suggestion
    ssize_t retcode = H5Aget_name(_hid, static_cast<hsize_t>(maxLength) + 1, buffer);
```

include/highfive/bits/H5Attribute_misc.hpp
```diff
@@ -27,6 +27,24 @@
...
+    if (retcode < 0) {
+        HDF5ErrMapper::ToException<GroupException>("Error accessing object name");
+    }
+    const size_t length = static_cast<std::size_t>(retcode);
+        if (length <= maxLength) {
```
tdegeus, line 39:
```suggestion
    if (length <= maxLength) {
```

include/highfive/bits/H5Attribute_misc.hpp
```diff
@@ -27,6 +27,24 @@
...
+        return std::string(buffer, length);
+    }
+    std::vector<char> bigBuffer(length + 1, 0);
+    H5Aget_name(
+        _hid, static_cast<hsize_t>(length) + 1, bigBuffer.data() );
```
tdegeus, line 44:
```suggestion
    H5Aget_name(_hid, static_cast<hsize_t>(length) + 1, bigBuffer.data());
```

include/highfive/bits/H5DataSet_misc.hpp
```diff
@@ -26,6 +26,24 @@
 
 namespace HighFive {
 
+inline std::string DataSet::getPath() const {
+    const size_t maxLength = 255;
+    char  buffer[maxLength + 1];
+    ssize_t retcode = H5Iget_name(
+        _hid, buffer, static_cast<hsize_t>(maxLength) + 1);
```
tdegeus, line 33:
```suggestion
    ssize_t retcode = H5Iget_name(_hid, buffer, static_cast<hsize_t>(maxLength) + 1);
```

include/highfive/bits/H5DataSet_misc.hpp
```diff
@@ -26,6 +26,24 @@
 
 namespace HighFive {
 
+inline std::string DataSet::getPath() const {
+    const size_t maxLength = 255;
+    char  buffer[maxLength + 1];
```
tdegeus, line 31:
```suggestion
    char buffer[maxLength + 1];
```

include/highfive/bits/H5DataSet_misc.hpp
```diff
@@ -26,6 +26,24 @@
...
+    ssize_t retcode = H5Iget_name(
+        _hid, buffer, static_cast<hsize_t>(maxLength) + 1);
+    if (retcode < 0) {
+        HDF5ErrMapper::ToException<GroupException>("Error accessing object name");
+    }
+    const size_t length = static_cast<std::size_t>(retcode);
+        if (length <= maxLength) {
```
tdegeus, line 38:
```suggestion
    if (length <= maxLength) {
```

include/highfive/bits/H5DataSet_misc.hpp
```diff
@@ -26,6 +26,24 @@
...
+        return std::string(buffer, length);
+    }
+    std::vector<char> bigBuffer(length + 1, 0);
+    H5Iget_name(
+        _hid, bigBuffer.data(), static_cast<hsize_t>(length) + 1 );
```
tdegeus, line 43:
```suggestion
    H5Iget_name(_hid, bigBuffer.data(), static_cast<hsize_t>(length) + 1);
```

include/highfive/bits/H5DataSet_misc.hpp
```diff
@@ -26,6 +26,24 @@
 
 namespace HighFive {
 
+inline std::string DataSet::getPath() const {
```
tdegeus, line 29: Like above: this should be moved to `Object`

include/highfive/bits/H5Node_traits.hpp
```diff
@@ -110,7 +110,38 @@ class NodeTraits {
     ///
     /// \brief return the name of the object with the given index
     /// \return the name of the object
-    std::string getObjectName(size_t index) const;
+    std::string getName(size_t index) const;
```
tdegeus, line 113:
```suggestion
    std::string getObjectName(size_t index) const;
```
Otherwise you break backward compatibility

include/highfive/bits/H5Node_traits.hpp
```diff
@@ -110,7 +110,38 @@ class NodeTraits {
...
+
+    ///
+    /// \brief return the path to the current object
+    /// \return the path to the object
+    std::string getPath() const;
```
tdegeus, line 118: By placing `getPath` in Object you will have it in `DataSet`, `DataSpace`, `Group`, Attribute`. Is there still functionality missing? If not, I think this should be removed.

include/highfive/bits/H5Node_traits.hpp
```diff
@@ -110,7 +110,38 @@ class NodeTraits {
...
+
+    ///
+    /// \brief moves an object and its content within an HDF5 file. Moved object is still valid but content - NOT!
```
tdegeus, line 121:
```suggestion
    /// \brief moves an object and its content within an HDF5 file.
```

I don't understand this remark. I have the feeling that it is not entirely true either.

include/highfive/bits/H5Node_traits.hpp
```diff
@@ -110,7 +110,38 @@ class NodeTraits {
...
+    bool moveObject(const File& dst_parent, const std::string& src_name, const std::string& dst_path, bool parents = true) const;
+
+    ///
+    /// \brief moves an object and its content within an HDF5 file. Moved object is still valid but content - NOT!
+    /// \param dst_parent destination object
+    /// \param dst_path relative path from dst_parent
+    /// \param parents Whether it shall create intermediate groups if
+    ///      necessary. Default: true
+    /// \return the group object
+    bool moveObject(const Group& dst_parent, const std::string& src_name, const std::string& dst_path, bool parents = true) const;
```
tdegeus, line 144:
```suggestion
///
    /// \brief moves an object within an HDF5 file.
    /// \param group group to which the file-paths are relative
    /// \param source_path relative path of the object
    /// \param dest_path new relative path for the object
    /// \param parents if true necessary intermediate groups are created. Default: true
    /// \return boolean that is true if the move was successful
    bool moveObject(const Group& group,
                    const std::string& source_path,
                    const std::string& dest_path,
                    bool parents = true) const;
```

If you want another overload I would suggest only path relative to a group. This corresponds to:

```cpp
file.moveObject(group, "data", "sub/new_data");
```

Given `group` having path `"/group" this would correspond to the move `"/group/data"` -> `""/group/sub/new_data"`

include/highfive/bits/H5Node_traits.hpp
```diff
@@ -110,7 +110,38 @@ class NodeTraits {
...
+    /// \brief return the path to the current object
+    /// \return the path to the object
+    std::string getPath() const;
+
+    ///
+    /// \brief moves an object and its content within an HDF5 file. Moved object is still valid but content - NOT!
+    /// \param dst_path relative path from dst_parent
+    /// \param parents Whether it shall create intermediate groups if
+    ///      necessary. Default: true
+    /// \return the group object
```
tdegeus, line 125:
```suggestion
    /// \param source_path absolute path of the object
    /// \param dest_path new absolute path for the object
    /// \param parents if true necessary intermediate groups are created. Default: true
    /// \return boolean that is true if the move was successful  
```

include/highfive/bits/H5Node_traits.hpp
```diff
@@ -110,7 +110,38 @@ class NodeTraits {
...
+    bool moveObject(const std::string& src_name, const std::string& dst_path, bool parents = true) const;
```
tdegeus, line 126: Formatting:

```suggestion
    bool moveObject(const std::string& source_path,
                    const std::string& dest_path,
                    bool parents = true) const;
```

include/highfive/bits/H5Node_traits_misc.hpp
```diff
@@ -142,7 +142,7 @@ inline size_t NodeTraits<Derivate>::getNumberObjects() const {
 }
 
 template <typename Derivate>
-inline std::string NodeTraits<Derivate>::getObjectName(size_t index) const {
+inline std::string NodeTraits<Derivate>::getName(size_t index) const {
```
tdegeus, line 145:
```suggestion
inline std::string NodeTraits<Derivate>::getObjectName(size_t index) const {
```
See earlier comment

include/highfive/bits/H5Node_traits_misc.hpp
```diff
@@ -162,6 +162,78 @@ inline std::string NodeTraits<Derivate>::getObjectName(size_t index) const {
...
+    const size_t length = static_cast<std::size_t>(retcode);
+        if (length <= maxLength) {
+        return std::string(buffer, length);
+    }
+    std::vector<char> bigBuffer(length + 1, 0);
+    H5Iget_name(
+        static_cast<const Derivate*>(this)->getId(), bigBuffer.data(), static_cast<hsize_t>(length) + 1 );
+    return std::string(bigBuffer.data(), length);
+}
+
```
tdegeus, line 183:
```suggestion
```
See earlier comment

include/highfive/bits/H5Node_traits_misc.hpp
```diff
@@ -162,6 +162,78 @@ inline std::string NodeTraits<Derivate>::getObjectName(size_t index) const {
...
+template <typename Derivate>
+inline bool NodeTraits<Derivate>::moveObject(const std::string& src_name,
+                                             const std::string& dst_path, bool parents) const {
+    RawPropertyList<PropertyType::LINK_CREATE> lcpl;
+    if (parents) {
+        lcpl.add(H5Pset_create_intermediate_group, 1u);
+    }
+    herr_t status = H5Lmove(static_cast<const Derivate*>(this)->getId(), src_name.c_str(),
+                            static_cast<const Derivate*>(this)->getId(), dst_path.c_str(), lcpl.getId(), H5P_DEFAULT);
+    if (status  < 0) {
```
tdegeus, line 193:
```suggestion
    if (status < 0) {
```
Formatting

include/highfive/bits/H5Node_traits_misc.hpp
```diff
@@ -162,6 +162,78 @@ inline std::string NodeTraits<Derivate>::getObjectName(size_t index) const {
...
+        HDF5ErrMapper::ToException<GroupException>(
+                    std::string("Unable to move link to \"") + dst_path + "\":");
+        return false;
+    }
+    return true;
+}
+
+
+
+
```
tdegeus, line 235:
```suggestion
template <typename Derivate>
inline bool NodeTraits<Derivate>::moveObject(const Group& group,
                                             const std::string& src_path,
                                             const std::string& dst_path,
                                             bool parents) const {
    RawPropertyList<PropertyType::LINK_CREATE> lcpl;
    if (parents) {
        lcpl.add(H5Pset_create_intermediate_group, 1u);
    }
    herr_t status = H5Lmove(group.getId(), src_path.c_str(),
                            group.getId(), dst_path.c_str(),
                            lcpl.getId(), H5P_DEFAULT);
    if (status < 0) {
        HDF5ErrMapper::ToException<GroupException>(
            std::string("Unable to move link to \"") + dst_path + "\":");
        return false;
    }
    return true;
}
```
See suggestion above

src/examples/renaming_and_moving_objects.cpp
```diff
@@ -0,0 +1,75 @@
+#include <H5File.hpp>
+#include <H5Group.hpp>
+#include <H5DataSet.hpp>
+#include <H5DataSpace.hpp>
+#include <H5Attribute.hpp>
+
+using namespace HighFive;
+
+int main(int argc, char *argv[])
```
tdegeus, line 9:
```suggestion
int main(void)
```
This is why the CI is failing

src/examples/renaming_and_moving_objects.cpp
```diff
@@ -0,0 +1,75 @@
...
+     * Secondly we will move dataset with attached attribute to
+     * some destination path. Every '/' sign means that a group
+     * will be created there (it is done automatically).
+     * When we have moved the dataset we can check if dataset
+     * object is still valid? We do this by creating second
+     * attribute. There is a tricky part at the end when we
+     * move Group with its dataset */
+
+    // Create a new file using the default property lists.
+    HighFive::File file("names.h5", File::ReadWrite | File::Create | File::Truncate);
```
tdegeus, line 24:
```suggestion
   File file("names.h5", File::ReadWrite | File::Create | File::Truncate);
```
Since you're using the namespace (which personally I hate, but ok, the same is done in other examples.

src/examples/renaming_and_moving_objects.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+    // Create a group
+    HighFive::Group group = file.createGroup("group");
```
tdegeus, line 27:
```suggestion
   Group group = file.createGroup("group");
```

src/examples/renaming_and_moving_objects.cpp
```diff
@@ -0,0 +1,75 @@
...
+     * dataset to this group and attach attribute to the dataset.
+     * Then we are trying to get path to the root, group dataset
+     * and the name of the dataset.
+     * Secondly we will move dataset with attached attribute to
+     * some destination path. Every '/' sign means that a group
+     * will be created there (it is done automatically).
+     * When we have moved the dataset we can check if dataset
+     * object is still valid? We do this by creating second
+     * attribute. There is a tricky part at the end when we
+     * move Group with its dataset */
```
tdegeus, line 21:
```suggestion
    /* We are going to create group in root directory then add
     * dataset to this group and attach attribute to the dataset.
     * Then we are trying to get path to the root, group dataset
     * and the name of the dataset.
     * Secondly we will move dataset with attached attribute to
     * some destination path.
     * To check if dataset object is still valid, we create a
     * second attribute */
```

src/examples/renaming_and_moving_objects.cpp
```diff
@@ -0,0 +1,75 @@
...
+    attribute.write(string_list);
+
+    // Get path and names
+    std::cout << "root path: \t" << file.getPath() << std::endl;
+    std::cout << "group path: \t" << group.getPath() << std::endl;
+    std::cout << "dataset path: \t" << dataset.getPath() << std::endl;
+    std::cout << "attribute name: \t" << attribute.getName() << std::endl;
+    std::cout << std::endl;
+
+    // move dataset with its attribute to another destination path
```
tdegeus, line 45:
```suggestion
    // Move dataset with its attribute to another destination path
```

src/examples/renaming_and_moving_objects.cpp
```diff
@@ -0,0 +1,75 @@
...
+    group.moveObject(file, "data", "/NewGroup/SubGroup/movedData");
+
+    // as you can see to reach destination path new groups were created as well
```
tdegeus, line 48:
```suggestion
    // As you can see to reach destination path new groups were created as well
```

src/examples/renaming_and_moving_objects.cpp
```diff
@@ -0,0 +1,75 @@
...
+    std::cout << "dataset new path: \t" << dataset.getPath() << std::endl;
+    std::cout << std::endl;
+
+    // we can still use moved dataset
```
tdegeus, line 52:
```suggestion
    // We can still use moved dataset
```

src/examples/renaming_and_moving_objects.cpp
```diff
@@ -0,0 +1,75 @@
...
+    Attribute attributeNew = dataset.createAttribute<std::string>("attributeNew", DataSpace::From(string_list));
+    attribute.write(string_list);
+    std::cout << "attribute new name: \t" << attributeNew.getName() << std::endl;
+    std::cout << std::endl;
+
+    // move the folder with its content to other place
+    file.moveObject("/NewGroup/SubGroup", "/FinalDestination");
+
+    // here is the important moment. Old 'dataset' variable tells us
+    // that dataset directory wasn't changed
```
tdegeus, line 63:
```suggestion
    // Here is the important moment. The old 'dataset' variable tells us
    // that dataset directory wasn't changed
```

src/examples/renaming_and_moving_objects.cpp
```diff
@@ -0,0 +1,75 @@
...
+    std::cout << "dataset new path wasn't changed: \t" << dataset.getPath() << std::endl;
+    std::cout << std::endl;
+
+    // but actually it was moved we just need to update variable
```
tdegeus, line 67:
```suggestion
    // But actually it was moved we just need to update variable
```

src/examples/renaming_and_moving_objects.cpp
```diff
@@ -0,0 +1,75 @@
...
+    std::cout << "actually it was moved we just need to update it: \t" << file.getDataSet("/FinalDestination/movedData").getPath() << std::endl;
+    std::cout << std::endl;
+
+    /* The conclusion: if you move a Group always update the varibles
+     * to its content :) */
+
```
tdegeus, line 73:
```suggestion
```

tdegeus:
> At the beginning I wanted to do so. But then I have found that `File, Group and DataSet` inherit from `Object` and the `Object` has **protected** constructors. As I know there is no way to create object of type `Object` (using High Five) of course, original hdf5 library allows that).
> Such protected constructors protects us from making bugs but we pay for that by the impossibility of use Object. That is how I see that. And this is someone's decision I think, someone who know the library better. So I decided not to make `Object` constructor public.

I don't think the constructors should worry you at all. You are just interested to add a method that all of the derived classes can use.

KerimMatlab:
@tdegeus I've just accepted some of your changes.
Here is my point of view:

`HighFive` library for `hdf5` is a wrapper for `hdf5` library. `hdf5` library has great documentation but it is written in C language wich is inconvenient to use (for me for example). In the same time now `HighFive` has some structure (that I like) wich can be interpreted as one `hdf5` function is equal to one `HighFive` method for some of the object (File, Group, Dataset, Attribute etc..).
You propose to use `getPath()` and `getName()` to `HighFive::Object` because `File`, `Group`, `Dataset`, `Attribute` inherit from `HighFive::Object`. So it may seem easier to implement those functions to `HighFive::Object`. But it is not. If we implement `getPath()` in `HighFive::Object` then all other types that inherit from `HighFive::Object` (`Attribute`, `Dataspace`) is going to have the same method `getPath()`. Do we need `getPath()` for `Attribute` or `Dataspace`? 
For attribute to `getName()` we use [H5Aget_name(...)](https://portal.hdfgroup.org/display/HDF5/H5A_GET_NAME). To `getPath(...)` to `Dataset` etc we should use [H5Iget_name(...)](https://portal.hdfgroup.org/display/HDF5/H5I_GET_NAME). Maybe we could handle this and puth these functions to `HighFive::Object` but this immediatle leads to complicated of code. Thus we can lost the basic idea of `HighFive` **one method = one hdf5 function**
Moreover `Object` constructor is protected so we cannot get an object of type `Object` somwhere outside of inherited classes. It is hard to say wether it is good or not but while it is as it is I don't see an importance to move `getPath()` and `getName()` to `Object`. I think `NodeTraits` were created for such tasks.

About overloading of `NodeTraits::moveObject()`. I strongly believe that there should be at least three methods:

- bool moveObject(const std::string& src_path, const std::string& dest_path, bool parents = true) const;
- bool moveObject(const File& file,  const std::string& src_path, const std::string& dest_path, bool parents = true) const;
- bool moveObject(const Group& group, const std::string& src_path, const std::string& dest_path, bool parents = true) const;

First of all I dont see any inconveniences using them. `hdf5` is something like a filesystem and usual filesystem is based on a tree models. So in model/view programming `moveObject(const File& file,...)` may faciliate the code.

Sorry for late answer.
Thank you for commenting and editing!

src/examples/renaming_and_moving_objects.cpp
```diff
@@ -0,0 +1,70 @@
+#include <H5File.hpp>
+#include <H5Group.hpp>
+#include <H5DataSet.hpp>
+#include <H5DataSpace.hpp>
+#include <H5Attribute.hpp>
```
tdegeus, line 5: This causes the error in the CI:
```suggestion
#include <highfive/H5File.hpp>
#include <highfive/H5Group.hpp>
#include <highfive/H5DataSet.hpp>
#include <highfive/H5DataSpace.hpp>
#include <highfive/H5Attribute.hpp>
```

src/examples/renaming_and_moving_objects.cpp
```diff
@@ -0,0 +1,70 @@
...
+    attribute.write(string_list);
+    std::cout << "attribute new name: \t" << attributeNew.getName() << std::endl;
+    std::cout << std::endl;
+    
+    // move the folder with its content to other place
+    file.moveObject("/NewGroup/SubGroup", "/FinalDestination");
+    
+    // Here is the important moment. The old 'dataset' variable tells us
+    // that dataset directory wasn't changed
+    std::cout << "_____Moving folder with its content_____" << std::endl;
```
tdegeus, line 61:
```suggestion
```
It's not really a folder is it?

src/examples/renaming_and_moving_objects.cpp
```diff
@@ -0,0 +1,70 @@
...
+    std::cout << "dataset new path wasn't changed: \t" << dataset.getPath() << std::endl;
+    std::cout << std::endl;
+    
+    // But actually it was moved we just need to update variable
+    std::cout << "actually it was moved we just need to update it: \t" << file.getDataSet("/FinalDestination/movedData").getPath() << std::endl;
```
tdegeus, line 66:
```suggestion
    std::cout << "actually it was moved we just need to update it: " << std::endl;
    std::cout << file.getDataSet("/FinalDestination/movedData").getPath() << std::endl;
```

KerimMatlab:
@tdegeus I can't write test. As I remember you proposed some test did you? If so we could add it

> Something strange happened to H5Node_traits*.hpp in one of your latest commit. Git seems to think you changed the entire file. It would be great if you could fix that.

Yes I reuploaded these files. I don't fully understand how GitHub works so it is my first experience of doing such things there.

tdegeus:
Here they are: I guess you could add them to `tests/unit/tests_high_five_base.cpp`

```cpp
BOOST_AUTO_TEST_CASE(HighFiveMove) {

    File file("move.h5", File::ReadWrite | File::Create | File::Truncate);

    int number = 100;

    {
        Group group = file.createGroup("group");
        DataSet dataset = group.createDataSet("data", DataSpace(1), AtomicType<int>());
        dataset.write(number);
        std::string path = dataset.getPath();
        BOOST_CHECK_EQUAL("/group/data", path);
    }

    file.moveObject("/group/data", "/new/group/new/data");

    {
        DataSet dataset = file.getDataSet("/new/group/new/data");
        std::string path = dataset.getPath();
        BOOST_CHECK_EQUAL("/new/group/new/data", path);
        int read;
        dataset.read(read);
        BOOST_CHECK_EQUAL(number, read);
    }
}

BOOST_AUTO_TEST_CASE(HighFiveMoveRelative) {

    File file("move.h5", File::ReadWrite | File::Create | File::Truncate);
    Group group = file.createGroup("group");

    int number = 100;

    {
        DataSet dataset = group.createDataSet("data", DataSpace(1), AtomicType<int>());
        dataset.write(number);
        std::string path = dataset.getPath();
        BOOST_CHECK_EQUAL("/group/data", path);
    }

    file.moveObject(group, "data", "sub/new_data");

    {
        DataSet dataset = group.getDataSet("sub/new_data");
        std::string path = dataset.getPath();
        BOOST_CHECK_EQUAL("/group/sub/new_data", path);
        int read;
        dataset.read(read);
        BOOST_CHECK_EQUAL(number, read);
    }
}
```

As to the history, your editor seems to have stored some character at the end of each line, so once you remove them the history will be ok again:

Screenshot 2020-05-31 at 15 41 50

KerimMatlab:
@tdegeus I added test, thank you!
Also now example should look more attractive.

> As to the history, your editor seems to have stored some character at the end of each line, so once you remove them the history will be ok again:

This caracters should be EOL symbols for Windows. I'm trying to understand how git hub EOL handles for users with different platform.

tdegeus:
@KerimMatlab Great! Did you notice that it now says: all checks have passed? 

To change the line-endings you may want to consider this : https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings (whereby you can skip the per-repository settings) or https://stackoverflow.com/a/10419350/2646505

KerimMatlab:
@tdegeus yes I noticed that. This is great!
Thank you for the link. I need some time to tackle this because it is all about Git Hub staff which is I poorely know yet...

tdegeus:
@KerimMatlab If I'm not mistaken, I think it is mostly about a peculiarity that Windows saves line-endings differently that Unix based systems. This first two relevant hits I found are http://www.cs.toronto.edu/~krueger/csc209h/tut/line-endings.html and https://www.edwardthomson.com/blog/git_for_windows_line_endings.html . Now it is often so that project use the Unix based line-endings, and you should therefore upload as such. So if you fix the settings on your git client, and do some empty commit I think you should be again good to go

tdegeus:
(Note that I approve the content, but I still think the issue with the history of `H5node*.hpp` should be solved.)

Personally I'm still not convinced with having an overload
```cpp
group.moveObject(file, "old", "new")
```
makes any sense, but you seem to be quite convinced. Indeed, having it strictly doesn't impede other usage.  Maybe others can comment as well? @ferdonline @matz-e @alkino

KerimMatlab:
> (Note that I approve the content, but I still think the issue with the history of `H5node*.hpp` should be solved.)

I afree with that. Seems like Git Hub can handle EOL for everyone who downloads the project. I tried to do as it said [there](https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings). I hope that now if you download the project there will be no problems with line endings.

> group.moveObject(file, "old", "new")

Personally me needs this. When working with model/view programming it is much easier to store full path to the object inside `h5` file and the file itself. 
You don't need to store every parent group for each object in tree model

tdegeus:
> 
> > group.moveObject(file, "old", "new")
> 
> Personally me needs this. When working with model/view programming it is much easier to store full path to the object inside `h5` file and the file itself.
> You don't need to store every parent group for each object in tree model

I should maybe clarify. I see the point of using relative paths. My question was simply, since there is already
```cpp
file.moveObject(group, "old", "new")
```
I wonder what the benefit is of having also
```cpp
group.moveObject(file, "old", "new")
```

KerimMatlab:
@tdegeus thank you! Now I see what is the problem :)
Probably you are rigth, we just need to leave:
```
file.moveObject("old", "new")
group.moveObject("old", "new")
```
Where "old" and "new" are full path to the object relative to `file` and `group` respectively.
Sorry for being stubborn.
Lets wait one-two days for other member comments or should I change it now? How do you think?

tdegeus:
> @tdegeus thank you! Now I see what is the problem :)
> Probably you are rigth, we just need to leave:
> 
> ```
> file.moveObject("old", "new")
> group.moveObject("old", "new")
> ```
> 
> Where "old" and "new" are full path to the object relative to `file` and `group` respectively.
> Sorry for being stubborn.
> Lets wait one-two days for other member comments or should I change it now? How do you think?

I like it! But's indeed let wait for other comments

KerimMatlab:
@tdegeus are EOL fine now? haven't you checked?

tdegeus:
I don't think so. If you look under "Files changed" you still see many seeming changes on lines you did not edit because of the line endings. (Also you didn't make a commit to the PR for a while, so by definition no changes ;))

KerimMatlab:
@tdegeus I changed EOL. Now it should by UNIX style
Also I removed methods like `file.moveObject(group, "old", "new")`
And Check the `tests_high_five_base.cpp` I commented lines 1241-1265 because with new syntaxes it unuseful I think

tests/unit/tests_high_five_base.cpp
```diff
@@ -1212,6 +1212,58 @@ typedef struct {
...
+
+//     {
+//         DataSet dataset = group.getDataSet("sub/new_data");
+//         std::string path = dataset.getPath();
+//         BOOST_CHECK_EQUAL("/group/sub/new_data", path);
+//         int read;
+//         dataset.read(read);
+//         BOOST_CHECK_EQUAL(number, read);
+//     }
+// }
```
tdegeus, line 1265: We could do:
```suggestion
BOOST_AUTO_TEST_CASE(HighFiveMoveRelative) {

    File file("move.h5", File::ReadWrite | File::Create | File::Truncate);
    Group group = file.createGroup("group");

    int number = 100;

    {
        DataSet dataset = group.createDataSet("data", DataSpace(1), AtomicType<int>());
        dataset.write(number);
        BOOST_CHECK_EQUAL("/group/data", dataset.getPath());
    }

    group.moveObject("data", "new_data");

    {
        DataSet dataset = group.getDataSet("new_data");
        BOOST_CHECK_EQUAL("/group/new_data", dataset.getPath());
        int read;
        dataset.read(read);
        BOOST_CHECK_EQUAL(number, read);
    }
}
```

KerimMatlab:
Thank you
I hope this pull request will be accepted and merged

tdegeus:
(Closing and opening to trigger the CI)

include/highfive/bits/H5Node_traits.hpp
```diff
@@ -112,6 +112,21 @@ class NodeTraits {
...
+    /// \return the path to the object
+    std::string getPath() const;
+
+    ///
+    /// \brief moves an object and its content within an HDF5 file.
+    /// \param src_path relative path of the object to current File/Group
+    /// \param dest_path new relative path of the object to current File/Group
+    /// \param parents if true necessary intermediate groups are created. Default: true
+    /// \return boolean that is true if the move was successful
+    bool moveObject(const std::string& src_path,
```
matz-e, line 126: Personal nitpick: I find the `Object` a little redundant, could this be just `move`, or better `rename`, since I don't think that any of the data is actually moved?

include/highfive/bits/H5Node_traits_misc.hpp
```diff
@@ -162,6 +162,42 @@ inline std::string NodeTraits<Derivate>::getObjectName(size_t index) const {
     return std::string(bigBuffer.data(), length);
 }
 
+template <typename Derivate>
+inline std::string NodeTraits<Derivate>::getPath() const {
```
matz-e, line 166: The code here seems to be largely identical with the above, could you please remind me why this could not be simplified?

include/highfive/bits/H5Node_traits.hpp
KerimMatlab: `rename` sounds interesting but for me if I saw for a first time the `File/Group` method `rename` (or `move`) then I would think that I'm renaming (moving) this `File/Group`. How do you think?

tdegeus: `move` might in the end be closed to the underlying `H5Lmove`. On the other-hand `rename` might be closest to the description of `H5Lmove`: "Renames a link within an HDF5 file.". `moveObject` is a bit a name on its own. Probably I would vote for `rename`. Or if you're hesitant that it is not sufficiently clear, I could still suggest `renameLink` which is short of the description of the underlying function.

KerimMatlab: I can do as you want. It doesn't matter a lot for me
Everyone wants it is to be called `rename`?

matz-e:
>  My point of view: if you want that moveObj would be rename (or some other) just say 'please rename moveObj() to rename'.

I wasn't sure if the naming should be changed. You're already passing in an object, so in my opinion, the `Object` in the name is a bit superfluous. But ultimately, I'm not the only that is responsible for this repository. Hence me not flat out requesting this to be changed. Maybe @ohm314 could give his opinion if `highfive::Group::moveObject` could have the `Object` dropped?

I'll have a look deduplicating the other code…

include/highfive/bits/H5Node_traits_misc.hpp
matz-e: It's more about avoiding repeated code. Also, if you read the same pattern several times in the code, you wonder if there is any difference, and why… so I for example spend more energy doing mental diffs to see if there's anything special in similar looking code…

I have an example locally, just thinking about ways to make it more intuitive before I push it somewhere.

include/highfive/bits/H5Node_traits.hpp
matz-e: Yes, please.

include/highfive/bits/H5Node_traits_misc.hpp
matz-e: See [this commit](https://github.com/BlueBrain/HighFive/commit/fedd857059a21ce43ed263bc3ee4da0f685bb9c0). It's somewhat hideous, but so is the nature of C++, and the fact that HDF5 has two pretty much identical functions (in purpose) with arguments swapped in the signature.

KerimMatlab:
@matz-e I just replaced `moveObj()` by `rename()`. 
Thank you, now the code looks cleaner and I accepted your commit. As I understood you used lambda expression (which I don't know yet) but there was a place where you forgot to replace the code. So I did it by myself and as I don't know lambda expressions please look at [the commit](https://github.com/BlueBrain/HighFive/pull/346/commits/fc484f1c686d7fae78d1813306a52fd5e8a9a3f9) have I done it correctly?

tdegeus:
You did it almost correctly! The only thing that needed improvement was that you did not allow for a variably length, by hand-specifying it. I corrected it, along with another issue that appeared for a reason unclear to me. Anyway since it was a bug it is good that it is now fixed.

tdegeus:
Very nice cleanup @KerimMatlab and @matz-e !

matz-e:
> Very nice cleanup @KerimMatlab and @matz-e !

I was just giving it another look to approve; One last thing (sorry!): could we have a test for `Attribute::getName`, too? Then I'm ready to approve myself

matz-e:
Merging after no additional dissenting feedback received.

matz-e:
Thank you again for your contributions!

Issue #305, opened 2020-03-03
Give the possibility to use the parameter 'op' with hyperslab


alkino:
https://support.hdfgroup.org/HDF5/doc/RM/RM_H5S.html#Dataspace-SelectHyperslab

https://github.com/BlueBrain/HighFive/blob/master/include/highfive/bits/H5Slice_traits_misc.hpp#L74-L96

alkino:
Hyperslab are really powerful but our API limit this. Try to find a way to have more flexibility.

KerimMatlab:
@alkino What I think `High-Five` should provide to when selecting hyperslabs:
1) working with as many dimensions as it needs
2) selecting intersection of rows/cols in any dimension
3) selecting points by coordinates. It is not an square area generated by intersection of rows and cols but it rather a bunch of points.
4) probably we don't need to have two step: `select(...)` and then `read()` (or write). Probably we would need have the syntaxis like `read(rows, cols)`

To solve these statements I think we need to have a method with arbitrary number of input parameters. So we could have the syntax:
```
std::vector<size_t> ind0;
std::vector<size_t> ind1;
std::vector<size_t> ind2;
ind0 << 0 << 1 << 2;
ind1 << 0 << 1 << 2;
ind2 << 0 << 1 << 2;
read(ind0, ind1, ind2);
```
Here index at the end (0, 1, 2) is the dimension.
We could use `variadic templates` to create method with any namber of input parameters. I have not used `variadic templates` before but mu gut tells me that the staff we need.

`variadic templates` could be used in read by coordinates mthods as well:
`readCoord(ind0, ind1, ...)`

What do you think?

tdegeus:
A nice API could be using slices:

* E.g. https://www.tutorialspoint.com/numpy/numpy_indexing_and_slicing.htm
* https://xtensor.readthedocs.io/en/latest/view.html

KerimMatlab:
@tdegeus as I understood from the `numpy` example `slice` is defined by a `start`, `stop` and `step` values
I think it is unconvenient.
I'm from Matlab and here is very simple and comprehensive [indexing](https://www.mathworks.com/company/newsletters/articles/matrix-indexing-in-matlab.html) idea.
You can make any slice using it. You only need to specify indexes you want to extract along each dimension. Then you retrieve the area where those andexes intersect.

alkino:
WIth the last patch I think we can start trying to do more complicate stuff.

PR #351, opened 2020-06-10
Hyperslab

alkino:
Beginning of #305 
Any thoughts?
You can see example of usage there


alkino:
We still have to implement `NotB` and `NotA`. I was trying to add a `~` operator to `HyperSlab::Select`.
But I'm not sure to be able to implement everything combination ;)

tdegeus:
Nice!

include/highfive/bits/H5Slice_traits_misc.hpp
```diff
@@ -75,22 +75,9 @@ template <typename Derivate>
...
-
     DataSpace space = slice.getSpace().clone();
-    if (H5Sselect_hyperslab(space.getId(), H5S_SELECT_SET, offset_local.data(),
-                            stride.empty() ? NULL : stride_local.data(),
-                            count_local.data(), NULL) < 0) {
-        HDF5ErrMapper::ToException<DataSpaceException>("Unable to select hyperslap");
-    }
+    HyperSlab(HyperSlab::Select(offset, count, stride)).apply(space);
 
     return Selection(DataSpace(count), space, details::get_dataset(slice));
```
alkino, line 82: can someone explain me, what means this count?

include/highfive/bits/H5Slice_traits.hpp
```diff
@@ -49,6 +49,122 @@ class ElementSet {
...
+        ret.selects.emplace_back(sel, Op::Xor);
+        return ret;
+    }
+
+    HyperSlab& operator^=(const Select& sel) {
+        selects.emplace_back(sel, Op::Xor);
+        return *this;
+    }
+
+    void apply(const DataSpace& space) {
```
alkino, line 107: I should make a copy of this DataSpace and return it, right?
From this new DataSpace we can retrieve new dims for the `Selection` class.

alkino:
The problem is that HighFive does not support irregular hyperslab, so this patch is hard to do.

ferdonline:
Progress here looks nice. Do you think we could have tests running?

include/highfive/bits/H5Slice_traits_misc.hpp
```diff
@@ -71,54 +71,38 @@ inline ElementSet::ElementSet(const std::vector<std::vector<std::size_t>>& eleme
     }
 }
 
+template <typename Derivate>
+inline Selection SliceTraits<Derivate>::select(const HyperSlab& hyperslab) const {
+    const auto& slice = static_cast<const Derivate&>(*this);
+    auto space = hyperslab.apply(slice.getSpace());
+
+    return Selection(space, space, details::get_dataset(slice));
```
alkino, line 79: I think the error is here: before the first arg was `DataSpace(count)` that means nothing here...

include/highfive/bits/H5Slice_traits.hpp
```diff
@@ -49,10 +49,159 @@ class ElementSet {
...
+    DataSpace apply(const DataSpace& space_) const {
+        auto space = space_.clone();
+        for (const auto& sel: selects) {
+            if (sel.op == Op::None) {
+                H5Sselect_none(space.getId());
+            } else {
+                std::vector<hsize_t> offset_local(sel.offset.size());
+                std::vector<hsize_t> count_local(sel.count.size());
+                std::vector<hsize_t> stride_local(sel.stride.size());
+                std::vector<hsize_t> block_local(sel.stride.size());
```
sergiorg-hpc, line 128: Oups...

PR #377, opened 2020-10-02
optional documentation building in cmake

mmmaat:
* **problem**

  I use HighFive in another project with documentation that I build with "make doc".  But cmake complains a target "doc" already exists in HighFive.

* **solution** 

  With this PR one can disable HighFive documentation building with `set(HIGHFIVE_DOC OFF)` just before `add_subdirectory(/path/to/HighFive)`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,7 +35,7 @@ option(HIGHFIVE_USE_OPENCV "Enable OpenCV testing" ${USE_OPENCV})
 option(HIGHFIVE_UNIT_TESTS "Enable unit tests" ON)
 option(HIGHFIVE_EXAMPLES "Compile examples" ON)
 option(HIGHFIVE_PARALLEL_HDF5 "Enable Parallel HDF5 support" OFF)
-option(HIGHFIVE_DOC "Enable documentation building" ON)
+option(HIGHFIVE_BUILD_DOCS "Enable documentation building" ON)
 
 # In deplomyents we probably don't want/cant have dynamic dependencies
 option(HIGHFIVE_USE_INSTALL_DEPS "End applications by default use detected dependencies here" OFF)
@@ -115,6 +115,6 @@ if(HIGHFIVE_UNIT_TESTS)
   add_subdirectory(tests/unit)
 endif()
 
-if(HIGHFIVE_DOC)
+if(HIGHFIVE_BUILD_DOCS)
   add_subdirectory(doc)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,7 @@ option(HIGHFIVE_USE_OPENCV "Enable OpenCV testing" ${USE_OPENCV})
 option(HIGHFIVE_UNIT_TESTS "Enable unit tests" ON)
 option(HIGHFIVE_EXAMPLES "Compile examples" ON)
 option(HIGHFIVE_PARALLEL_HDF5 "Enable Parallel HDF5 support" OFF)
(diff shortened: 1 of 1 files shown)


tristan0x:
I would rather not import all HighFive Makefile targets when importing the CMake project:

```
add_subdirectory(/path/to/highfive EXCLUDE_FROM_ALL)
```
See [add_subdirectory documentation](https://cmake.org/cmake/help/v3.0/command/add_subdirectory.html)

matz-e:
> I would rather not import all HighFive Makefile targets when importing the CMake project:
> 
> ```
> add_subdirectory(/path/to/highfive EXCLUDE_FROM_ALL)
> ```
> 
> See [add_subdirectory documentation](https://cmake.org/cmake/help/v3.0/command/add_subdirectory.html)

Very useful hint! Would maybe be good to have this on the README, then.

tdegeus:
The tip is indeed very useful. At the same time I agree that it makes a lot of sense to have the option to (not) build the docs.

mmmaat:
Thanks for tip but actually I still have the same issue even when using `add_subdirectory(/path/to/highfive EXCLUDE_FROM_ALL)`
```
CMake Error at doc/CMakeLists.txt:33 (add_custom_target):
  add_custom_target cannot create target "doc" because another target with
  the same name already exists.  The existing target is a custom target
  created in source directory
  ".../external/HighFive/doc".  See documentation for policy CMP0002 for more details.
```

matz-e:
It's great to have namespaces! /s

Rather than adding an option, an alternative would be to only enable the `doc` target when we are built as a top-level project, and not included. [CLI11 does this in several places.](https://github.com/CLIUtils/CLI11/blob/master/CMakeLists.txt#L280)

CMakeLists.txt
```diff
@@ -35,6 +35,7 @@ option(HIGHFIVE_USE_OPENCV "Enable OpenCV testing" ${USE_OPENCV})
 option(HIGHFIVE_UNIT_TESTS "Enable unit tests" ON)
 option(HIGHFIVE_EXAMPLES "Compile examples" ON)
 option(HIGHFIVE_PARALLEL_HDF5 "Enable Parallel HDF5 support" OFF)
+option(HIGHFIVE_DOC "Enable documentation building" ON)
```
matz-e, line 38:
```suggestion
option(HIGHFIVE_BUILD_DOCS "Enable documentation building" ON)
```
…would be more fitting, IMO.

ferdonline:
there is already some logic for the DOCS target. If dependencies are not met then it won't make the target available.
Having this flag, if false, we should even avoid looking for doc dependencies and avoid issuing the warning

ferdonline:
> Rather than adding an option, an alternative would be to only enable the `doc` target when we are built as a top-level project, and not included. [CLI11 does this in several places.](https://github.com/CLIUtils/CLI11/blob/master/CMakeLists.txt#L280)

I also don't think it makes sense to build HighFive docs as a subproject. That would avoid the manual var which is something we have wanted to avoid. We preferred a target that is manually built

matz-e:
> there is already some logic for the DOCS target. If dependencies are not met then it won't make the target available.
> Having this flag, if false, we should even avoid looking for doc dependencies and avoid issuing the warning

The issue here is namespace pollution… i.e., HighFive targets "leak" into the global scope and squat common target names, and dependency searching for the intended documentation build may first create the HighFive target with the same name. I ran into this at some point, too…

tdegeus (approved):

tdegeus:
Ping @matz-e @alkino @ferdonline

PR #366, opened 2020-08-07
Recursives converters

alkino:

Issue #285, opened 2020-02-11
Unify read/write routines for Datasets and Attributes


ferdonline:
Datasets and Attributes duplicate much of the code to read and write data.
In #277 initial work has been done separating read/writing logic into its own source file.
If would be desirable to make such routines to work for Attributes as well, with eventual minor changes.

alkino:
You would like that Attribute inherit from Slice_traits?

PR #378, opened 2020-10-05
Unify interface of H5Attribute with H5Slice_traits

alkino:


matz-e (commented):
So no data conversion for raw operations?

alkino:
> So no data conversion for raw operations?
What do you mean, it's here only for `read_raw` (called `read(T*)`) and `write_raw`.

ferdonline:
This kinda addresses #285, right? they use the same API, although they don't share code (yet!) :)

alkino:
This patch was extracted from #366 to simplify it.

The next patch is #366 and deduplicate most things.

ferdonline (approved):

PR #381, opened 2020-10-13
Move array_dims in the future manipulator

alkino:
Fix vector<Eigen::Matrix> and boost::multi_array<Eigen::Matrix>, from now
Eigen::Matrix has 2 dimensions and no more 1 with hackish things.

### DIFF
--- a/include/highfive/bits/H5Utils.hpp
+++ b/include/highfive/bits/H5Utils.hpp
@@ -38,62 +38,89 @@ class FixedLenStringArray;
 
 
 namespace details {
-
-// determine at compile time number of dimensions of in memory datasets
 template <typename T>
-struct array_dims {
-    static constexpr size_t value = 0;
+struct manipulator {
+    using type = T;
+
+    static const size_t n_dims = 0;
+    static const size_t r_n_dims = n_dims;
 };
 
-template <std::size_t N>
-struct array_dims<FixedLenStringArray<N>> {
-    static constexpr size_t value = 1;
-};
+template <size_t N>
+struct manipulator<FixedLenStringArray<N>> {
+    using type = FixedLenStringArray<N>;
 
-template <typename T>
-struct array_dims<std::vector<T> > {
-    static constexpr size_t value = 1 + array_dims<T>::value;
-};
-
-template <typename T>
-struct array_dims<T*> {
-    static constexpr size_t value = 1 + array_dims<T>::value;
-};
-
-template <typename T, std::size_t N>
-struct array_dims<T[N]> {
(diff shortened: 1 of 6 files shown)


include/highfive/bits/H5Utils.hpp
```diff
@@ -38,63 +38,90 @@ class FixedLenStringArray;
...
-
-// determine at compile time number of dimensions of in memory datasets
 template <typename T>
-struct array_dims {
-    static constexpr size_t value = 0;
+struct manipulator {
+    using type = T;
+
+    static const size_t n_dims = 0;
+    static const size_t r_n_dims = n_dims;
```
tdegeus, line 46: To me, the variable names are not really descriptive...

tdegeus (commented):
Very nice that you are moving all details to a central location. This will simplify the lib in the future.

include/highfive/bits/H5Utils.hpp
alkino: Help me with that, please, I'm not good.

tdegeus: Some suggestions:

* `n_dims` -> `rank` / `dimension` or `outer_rank` / `outer_dimension` 
* `r_n_dims` -> `recursive_rank` / `total_rank` / `full_rank` / `*_dimension` or `rank` / `dimension`

Or... something even better ;) I'd be happy to iterate.

ferdonline: Maybe also
 `manipulator` -> `array_info` ?
Or you intent that in the future to manipulate something? (but the word is not much used I guess)

ferdonline:
I like this improvement! Thanks @alkino

include/highfive/bits/H5Utils.hpp
ferdonline: for the properties, I would keep consistency. In DataSpace we have the following methods:
```
size_t getNumberDimensions() const;
size_t getElementCount() const;
```
so, what about
`.number_dimensions`
`.element_count`

alkino: `n_dims` => number of dimensions of the Object itself (e.g `std::vector` => 1)
`r_n_dims` => total number of dimensions (e.g `std::vector<std::vector>` => 2),
It's not an  `element_count`

alkino: it's not only `array_info`, i  will add methods to resize array, `ŧransform_read`, `transform_write`, etc...

alexsavulescu (commented):
LGTM, nice work.

ferdonline (approved):

PR #382, opened 2020-10-13
Put type_of_array in inspector

alkino:


include/highfive/bits/H5Utils.hpp
```diff
@@ -37,10 +37,15 @@ template <std::size_t N>
...
 
+template <typename T>
+using unqualified_t = typename std::remove_const<typename std::remove_reference<T>::type
+        >::type;
+
 namespace details {
 template <typename T>
 struct manipulator {
     using type = T;
+    using hdf5_type = unqualified_t<T>;
```
ferdonline, line 48: to me this is a `base_type` from a composite type. Is there any connection to hdf5?

alkino: hdf5_type is the type than HDF5 want to see internally.
the type behind the `void*` in `write(void* value)`

alkino: I don't remember right now when, but sometimes it differs from the `h5_type` (futur) that is the type to store the array.

ohm314 (approved):

include/highfive/bits/H5Utils.hpp
alkino: but base_type can be a good name.

include/highfive/bits/H5Utils.hpp
```diff
@@ -37,10 +37,15 @@ template <std::size_t N>
 class FixedLenStringArray;
 
 
+template <typename T>
+using unqualified_t = typename std::remove_const<typename std::remove_reference<T>::type
+        >::type;
```
matz-e, line 42: That looks…weird

ohm314: looks fine to me?

ohm314: with C++20: https://en.cppreference.com/w/cpp/types/remove_cvref

ferdonline (approved):

ohm314 (approved):
lgtm

include/highfive/bits/H5Utils.hpp
ferdonline: LGTM as well

matz-e: Well, I hear complaints about
```
):
```
In Python code… so `>::type;` looks just as weird to me :)

PR #383, opened 2020-10-13
get_dim_vector in manipulator

alkino:


include/highfive/bits/H5Utils.hpp
```diff
@@ -109,6 +145,14 @@ struct manipulator<Eigen::Matrix<T, M, N>> {
 
     static const size_t n_dims = 2;
     static const size_t r_n_dims = n_dims + manipulator<value_type>::r_n_dims;
+
+    static std::vector<size_t> get_sizes(const type& val) {
```
ferdonline, line 149: cant we do
`std::array<size_t, n_dims>` ? (maybe ndims must be constexpr

alkino: ok, for `dimensions`,
I can return the `std::array`

alkino: for the `std::array` we agree that it will not be the `n_dims` but the `r_n_dims`, right?

include/highfive/bits/H5Utils.hpp
```diff
@@ -51,7 +51,7 @@ struct inspector {
     static constexpr size_t recursive_number_dimensions = number_dimensions;
 
     static std::array<size_t, recursive_number_dimensions> getDimensions(const type& /* val */) {
-        return std::array<size_t, recursive_number_dimensions>{};
+        return std::array<size_t, recursive_number_dimensions>();
```
matz-e, line 54:
```suggestion
        return std::array<size_t, recursive_number_dimensions>();
```
I'd say `recursive_number_of_dimensions` or `recursive_dimensions` (IMO `getDimensions()` → `dimensions()` would also be nice. STL has `.size()`, and not `.getSize()` :) )

alkino: hm,
`getDimensions` to follow DataSpace: `getDimensions`, `getNumberDimensions`

matz-e: Ok, consistency is good. `getNumberDimensions` seems like a bad name to me, though… could there also be `getStringDimensions`?

matz-e: Just saying that IMO, `getNumberDimensions` reads a little weird. `getNumberOfDimensions` would have been better. (I misread the former as the dimensions of things that are numbers due to the missing "of", hence my question about the dimensions of things that are "strings" :) )

TL;DR: `DataSpace` already has these methods, my lingual confusion should be ignored.

tdegeus: I agree with your @alkino in the sense that also I find `dimension()` (no 's'?) and `size()` a lot better than `getSize()` and `getDimensions()`. If the overhaul involves more API changes I think such a change could be discussed, but at the same time renaming is not super urgent. One could also phase in the change by having a deprecation warning for the old API. 

Then, `getNumberDimensions` I also find completely confusing: I immediately start thinking that this must be something special that I do not know about. In comparions `getNumberOfDimensions` is clear, but quite long and verbose. 

So my vote would be for an API change (following STL):
* `getSize()` -> `size()`
* `getDimensions()` -> `dimension()`

and for an internal use of
* `recursive_number_dimensions` -> `recursive_dimension` or `dimension_recursive`
* `number_dimensions` -> `dimension`

include/highfive/bits/H5Utils.hpp
```diff
@@ -68,6 +76,15 @@ struct inspector<std::vector<T>> {
 
     static constexpr size_t number_dimensions = 1;
     static constexpr size_t recursive_number_dimensions = number_dimensions + inspector<value_type>::recursive_number_dimensions;
+
+    static std::array<size_t, recursive_number_dimensions> getDimensions(const type& val) {
```
tdegeus, line 80: But then here I am confused, I thought I understood well but I'm starting to develop doubts... Isn't this `shape`??

alkino: Yes, but we call it `getDimensions`:
https://github.com/BlueBrain/HighFive/blob/master/include/highfive/H5DataSet.hpp#L87-L94

It's written here `getDimensions return shape`

tdegeus: OK. Not a fan ;) I would say an nd-array has:

* `shape` : size along each axis
* `size == prod(shape)`
* `dimension == rank == ndim == shape.size()` (here names do vary a bit in different libraries)

include/highfive/bits/H5Utils.hpp
matz-e: > `getDimensions()` -> `dimensions()`

Should keep the plural there, no? And yes, these would have to be added, and the older ones deprecated properly.

tdegeus: @matz-e If they are added my confusion below should be taken into account. I think that what is being called `dimensions` here should be called `shape`

ferdonline: Guys I don’t think we should at this point take care of prettifying the API. We have a lot of work already and dealing with deprecations and new API just for renaming... it might not be perfect but I at least is consistent.
Wrt internal api sure, we can add the "of" to leave no doubts.

ferdonline (approved):
I personally have nothing against getDimensions. Dimensions has been the word used in highfive since ever afaik. Switching to a synonym like “shape” is a matter of taste so IMO we shouldn’t do it as would break api and consistency. Further it makes it sound like a property, something that afaik has been avoided in Highfive.
If however you introduce a getNumberOfDimensions I agree with that naming (including the “of”)

include/highfive/bits/H5Utils.hpp
matz-e: @tdegeus ah, yes… I agree, being closer to numpy would be less confusing overall.

@ferdonline I think the one "of" should be in there. Otherwise, as per my tl;dr:, consistency trumps other stuff for right now.

include/highfive/bits/H5Dataspace_misc.hpp
```diff
@@ -150,12 +150,14 @@ inline DataSpace DataSpace::From(const ScalarValue& scalar) {
 
 template <typename Value>
 inline DataSpace DataSpace::From(const std::vector<Value>& container) {
-    return DataSpace(details::get_dim_vector<Value>(container));
+    auto dims = details::inspector<std::vector<Value>>::getDimensions(container);
+    return DataSpace(std::vector<size_t>(dims.begin(), dims.end()));
```
ferdonline, line 154: Adding a `std::array` ctor to DataSpace makes sense?

alkino: yes ;)

tdegeus:
> I personally have nothing against getDimensions. Dimensions has been the word used in highfive since ever afaik. Switching to a synonym like “shape” is a matter of taste so IMO we shouldn’t do it as would break api and consistency. Further it makes it sound like a property, something that afaik has been avoided in Highfive.
> If however you introduce a getNumberOfDimensions I agree with that naming (including the “of”)

Makes sense. But then why not harmonise the internal and external naming:

* `number_dimensions` -> `dimensions`
* `recursive_number_dimensions` -> `recursive_dimensions`

?

ferdonline:
> number_dimensions -> dimensions
> recursive_number_dimensions -> recursive_dimensions

`dimensions` to me sounds the dimensions themselves (numpy shape). if you want to have it short, then `ndim`, otherwise as suggested, `number_of_dimensions` also ok with me.
similar thoughts for the other one. Either as is, or one of `recursive_ndim`, `recursive_number_of_dimensions`

tdegeus:
> > number_dimensions -> dimensions
> > recursive_number_dimensions -> recursive_dimensions
> 
> `dimensions` to me sounds the dimensions themselves (numpy shape). if you want to have it short, then `ndim`, otherwise as suggested, `number_of_dimensions` also ok with me.
> similar thoughts for the other one. Either as is, or one of `recursive_ndim`, `recursive_number_of_dimensions`
> Or @tdegeus , was your suggestion `ndimensions` (with n)? Would have nothing against it.

`ndim` sounds good. Or otherwise `rank` could be considered

ferdonline:
`ndim` sounds good. Or otherwise `rank` could be considered

I would prefer ndim since it's a short for dimensions. Also, the fact that it is the same as numpy is nice

tdegeus:
> the fact that it is the same as numpy is nice

Agreed!

alkino:
Should I change it?
n_dims r_n_dims?

ferdonline:
So, last version is `ndim` and `recursive_ndim`

ferdonline (approved):

PR #384, opened 2020-10-13
From can be automatic now

alkino:


matz-e (approved):
Nice and simple!

include/highfive/bits/H5Dataspace_misc.hpp
```diff
@@ -125,100 +125,16 @@ inline std::vector<size_t> DataSpace::getMaxDimensions() const {
...
-        "  boost_container_types = "
-        "boost::numeric::ublas::matrix<all_basic_types> | "
-        "boost::multi_array<all_basic_types> \n"
-        "  all_supported_types = all_basic_types | stl_container_types | "
-        "boost_container_types"
-        "  eigen_matrix_type = Eigen::Matrix<signed_arithmetic_type> | Eigen::VectorXX");
-    return DataSpace(DataSpace::datascape_scalar);
+template <typename T>
+inline DataSpace DataSpace::From(const T& value) {
+    return DataSpace(details::manipulator<T>::get_sizes(value));
```
ferdonline, line 130: get_sizes() -> get_dimensions?

include/highfive/bits/H5Dataspace_misc.hpp
```diff
@@ -125,100 +125,16 @@ inline std::vector<size_t> DataSpace::getMaxDimensions() const {
     return details::to_vector_size_t(maxdims);
 }
 
-template <typename ScalarValue>
-inline DataSpace DataSpace::From(const ScalarValue& scalar) {
-    (void)scalar;
-    static_assert(
```
ferdonline, line 131: As much as I like simplifying, the assert message could be helpful, better than a whole lot of template errors.

include/highfive/H5DataSpace.hpp
```diff
@@ -110,47 +110,12 @@ class DataSpace : public Object {
     /// Create a dataspace matching a single element of a basic type
     ///  supported type are integrals (int,long), floating points (float,double)
     ///  and std::string
-    template <typename ScalarValue>
-    static DataSpace From(const ScalarValue& scalar_value);
-
-    /// Create a dataspace matching the container dimensions and size
-    /// Supported Containers are:
```
ohm314, line 117: I guess you'd like to keep some of the doc since the templated from supports more than just integrals?

ohm314 (approved):
Very nice, I like it

include/highfive/bits/H5Dataspace_misc.hpp
matz-e: or just `dimensions()`?

ferdonline: DataSpace API has `getDimensions()`. no need to be strictly consistent since this is internal API, but still...

alkino: Resolved in the previous patch.

include/highfive/bits/H5Dataspace_misc.hpp
alkino: What others think of it?

include/highfive/H5DataSpace.hpp
alkino: I write a new one

tdegeus: I would create a central location in the docs where the datatypes are listed, and then refer to it from here. In the future the supported datatypes will probably be extended and we'll probably forget to add new types to the docs here.

include/highfive/H5DataSpace.hpp
```diff
@@ -107,51 +107,13 @@ class DataSpace : public Object {
...
-    static DataSpace From(const std::array<Value, N>&);
-
-    template <typename ValueT, std::size_t N>
-    static DataSpace From(const ValueT(&container)[N]);
+    /// Create a dataspace matching a type accepted by details::inspector
+    template <typename T>
+    static DataSpace From(const T& value);
 
     template <std::size_t N, std::size_t Width>
     static DataSpace FromCharArrayStrings(const char(&)[N][Width]);
```
tdegeus, line 115: Could this be added to the generic template as well?

include/highfive/bits/H5Dataspace_misc.hpp
tdegeus: Having a `static_assert` is nice, but following my earlier comment for maintainability when new types are added it would be nice if it follows from the `inspector` (not sure have that could be done though?).

tdegeus (commented):
The overhaul is playing out nicely! Very nice simplifications and cleanup. It seems that most of the `H5Easy` implementation can be merged soon, exposing the core API to several new types.

include/highfive/bits/H5Dataspace_misc.hpp
alkino: I disagree with static_assert here ^^

include/highfive/H5DataSpace.hpp
alkino: not now, ArrayStrings are a bit complicate.

include/highfive/bits/H5Dataspace_misc.hpp
```diff
@@ -129,38 +129,9 @@ inline std::vector<size_t> DataSpace::getMaxDimensions() const {
...
-        (std::is_arithmetic<ScalarValue>::value ||
-         std::is_enum<ScalarValue>::value ||
-         std::is_same<std::string, ScalarValue>::value),
-        "Only the following types are supported by DataSpace::From: \n"
```
ferdonline, line 139: @alkino I thought a bit and we should keep this for users. For ourselves maybe we can disable this, but many users of HighFive don’t know templating and will be lost

ferdonline: Do you think there is a way, not too complicated to show the message in case `details::inspector<T>` doesnt support T ?

alkino: My concern is that it is hard to tell which types are accepted for `T` (for containers it is easy).

matz-e (approved):

include/highfive/bits/H5Dataspace_misc.hpp
alkino: And when it is compiled without Eigen (but you don't think about it) and the message is "Eigen::matrix is accepted".

ferdonline: ok

ferdonline (approved):

PR #385, opened 2020-10-15
Add and use datatype.

alkino:
Support new types in inspector: std::string, char[N]

### DIFF
--- a/tests/unit/tests_high_five_base.cpp
+++ b/tests/unit/tests_high_five_base.cpp
@@ -1466,10 +1466,11 @@ BOOST_AUTO_TEST_CASE(HighFiveFixedString) {
     }
 
     {  // Cant convert flex-length to fixed-length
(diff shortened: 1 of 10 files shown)


tests/unit/tests_high_five_base.cpp
```diff
@@ -1466,10 +1466,11 @@ BOOST_AUTO_TEST_CASE(HighFiveFixedString) {
     }
 
     {  // Cant convert flex-length to fixed-length
-        const char* buffer[] = {"abcd", "1234"};
-        SilenceHDF5 silencer;
-        BOOST_CHECK_THROW(file.createDataSet<char[10]>("ds5", DataSpace(2)).write(buffer),
-                          HighFive::DataSetException);
+       // From now error at compilation
```
ferdonline, line 1469: I like compile-time checks! :)

PR #386, opened 2020-10-19
Add Reference and std::string inside inspector

alkino:
On this PR I will try to reduce the power of `data_converter`.

### Types inside inspector
I firstly have to add `element_type` for `std::string`.
Inside `inspector`:
* `type`: the type inspector is taking care.
* `value_type` (for container only), the next type to inspect.
* `base_type`: the type that will be given to `create_and_check_datatype`.
* `element_type`: the type that HDF5 want to manage.

`base_type` and `element_type` are recursive.

Example why we need `base_type` and `element_type`:
`std::string` is converted to Variable-length string inside HDF-5 by `create_and_check_datatype`, but to `read` or `write` them we need to give `const char*` variable.
So `base_type` is `std::string` and `element_type` is `const char*`.

### Storage
I add a class `storage` because in some `data_converter` (those which are not handling contiguous types) we need to store a continuous vector of `element_type`. All `data_converter` are handling those vectors differently so it's hard to write a generic `data_converter` based on `inspector`. This storage class will disappear in future.

### Inspectors and Storage
Currently we have specialization of `data_converter` for **each** type! (e.g.: `H5::Reference`, `std::vector<T>` and `std::vector<H5::Reference>`).
So this PR is trying to remove things from `data_converter` (not recursive) to put them inside `inspector` (recursive).
With the new storage we add members inside inspectors to move from type to the storage and from the storage to the type.

### DIFF
--- a/include/highfive/bits/H5Converter_misc.hpp
+++ b/include/highfive/bits/H5Converter_misc.hpp
@@ -27,8 +27,8 @@
 #include <H5Dpublic.h>
 #include <H5Ppublic.h>
 
-#include "../H5Reference.hpp"
 #include "H5Utils.hpp"
+#include "../H5Reference.hpp"
 
 namespace HighFive {
 
@@ -253,7 +253,7 @@ struct data_converter<boost::numeric::ublas::matrix<T>, void>
 template <typename T>
 struct data_converter<std::vector<T>,
                       typename std::enable_if<(is_container<T>::value)>::type> {
-    using value_type = typename inspector<T>::base_type;
+    using value_type = typename inspector<T>::element_type;
 
     inline data_converter(const DataSpace& space)
         : _dims(space.getDimensions()) {}
@@ -275,14 +275,14 @@ struct data_converter<std::vector<T>,
     }
 
     std::vector<size_t> _dims;
-    std::vector<typename inspector<T>::base_type> _vec_align;
+    std::vector<value_type> _vec_align;
 };
 
 
 // apply conversion to scalar string
 template <>
 struct data_converter<std::string, void> {
-    using value_type = const char*;  // char data is const, mutable pointer
+    using value_type = typename inspector<std::string>::element_type;  // char data is const, mutable pointer
 
     inline data_converter(const DataSpace& space) noexcept
         : _c_vec(nullptr)
@@ -317,7 +317,7 @@ struct data_converter<std::string, void> {
 // apply conversion for vectors of string (dereference)
 template <>
 struct data_converter<std::vector<std::string>, void> {
-    using value_type = const char*;
+    using value_type = typename inspector<std::vector<std::string>>::element_type;
 
     inline data_converter(const DataSpace& space) noexcept
         : _space(space) {}
@@ -362,8 +362,27 @@ struct data_converter<FixedLenStringArray<N>, void>
     using container_converter<FixedLenStringArray<N>, char>::container_converter;
 };
 
+template <typename T>
+struct storage {
+    void resize(size_t size) {
+        _vec_align.resize(size);
+    };
+
+    T* pointer() {
+        return _vec_align.data();
+    }
+
+    const T* pointer() const {
+        return _vec_align.data();
+    }
+
+    std::vector<T> _vec_align;
+};
+
 template <>
 struct data_converter<std::vector<Reference>, void> {
+    using element_type = typename inspector<std::vector<Reference>>::element_type;
+
     inline data_converter(const DataSpace& space)
         : _dims(space.getDimensions()) {
         if (!is_1D(_dims)) {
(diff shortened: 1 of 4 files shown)


include/highfive/bits/H5Converter_misc.hpp
```diff
@@ -161,24 +161,24 @@ struct data_converter<CArray,
...
 
-    // Ship (pseudo)1D implementation
     inline value_type* transform_read(Container& vec) const {
         auto&& dims = _space.getDimensions();
-        if (!is_1D(dims))
-            throw DataSpaceException("Dataset cant be converted to 1D");
-        vec.resize(compute_total_size(dims));
-        return vec.data();
+        if (is_1D(dims)) { // This is a bad things, but keep for compatibility
+            dims = std::vector<size_t>{compute_total_size(dims)};
```
alkino, line 174: What a nightmare of this!

This patch is broken because of this violent things here!

What could we do?

include/highfive/bits/H5Converter_misc.hpp
```diff
@@ -161,24 +161,24 @@ struct data_converter<CArray,
...
+        }
+        inspector<Container>::resize(vec, dims);
+        return inspector<Container>::first(vec);
     }
 
     inline const value_type* transform_write(const Container& vec) const noexcept {
-        return vec.data();
+        return inspector<Container>::first(vec);
```
tdegeus, line 181: Just to understand what you are doing, is this a free function that mimics e.g. `vec.begin()` such that it can be done for all types (also for those that do not provide such a method)?

A remark that pops up now that I see this code is that the names `Container` and `vec` (together) confuse me a bit. Wouldn't `inspector<T>::begin(container)` or `inspector<T>::begin(data)` be more descriptive? Or did I misunderstood what this code does?

include/highfive/bits/H5ConverterEigen_misc.hpp
```diff
@@ -40,31 +40,17 @@ inline size_t compute_total_size(const boost::multi_array<T, Dims>& vec) {
 
 // apply conversion to eigen matrix
 template <typename T, int M, int N>
-struct data_converter<Eigen::Matrix<T, M, N>, void> {
+struct data_converter<Eigen::Matrix<T, M, N>, void>
+    : public container_converter<Eigen::Matrix<T, M, N>> {
 
-    typedef Eigen::Matrix<T, M, N> MatrixTMN;
+    using MatrixTMN = Eigen::Matrix<T, M, N>;
```
tdegeus, line 46: I would opt for a generic name maybe:
```cpp
using value_type = Eigen::Matrix<T, M, N>
```
(or so)

include/highfive/bits/H5Converter_misc.hpp
```diff
@@ -161,24 +161,24 @@ struct data_converter<CArray,
 };
 
 // Generic container converter
-template <typename Container, typename T = typename inspector<Container>::base_type>
+template <typename Container, typename T = typename inspector<Container>::element_type>
 struct container_converter {
-    typedef T value_type;
+    using value_type = T;
```
tdegeus, line 166: Seems strange to alias for a longer name...

tdegeus (commented):
I just wanted to answer to your Eigen question, but I ended up making some generic remarks too.

BTW HighFive's implementation is really getting a lot cleaner.

include/highfive/bits/H5Converter_misc.hpp
alkino: The way you fix it in easy is really interested but here it's for all types. !

A 3 x 3 x 1 space can be read as `std::vector<std::vector>` !

include/highfive/bits/H5Converter_misc.hpp
alkino: begin return an iterator.

The reason is to be generic for boost ublas matrix, this is implemented as `matrix.data(0, 0)`

include/highfive/bits/H5ConverterEigen_misc.hpp
alkino: I fix things little by little, at the end those data_converter would be deleted. The generic part is inside `H5Utils` with `inspector`.

include/highfive/bits/H5Converter_misc.hpp
alkino: alias with a name having a meaning ;)

I don't fix `data_converter` for now, i'm trying to remove what is not generic from them.

I cannot do everything in the same time, because it's a lot of little things locally *fixed*.

tdegeus: Sure ;) I just didn't want to not say.

alkino:
Endly I stop trying to fix Eigen in this patch ;) (it will comes maybe later).

include/highfive/bits/H5Reference_misc.hpp
```diff
@@ -64,6 +64,33 @@ inline Object Reference::get_ref(const Object& location) const {
     return Object(res);
 }
 
+namespace details {
+template <>
+struct inspector<Reference> {
```
ohm314, line 69: maybe not a bad idea to move this to the other `inspector` template classes.

include/highfive/bits/H5Converter_misc.hpp
```diff
@@ -392,39 +362,53 @@ struct data_converter<FixedLenStringArray<N>, void>
     using container_converter<FixedLenStringArray<N>, char>::container_converter;
 };
 
+template <typename T>
+struct storage {
```
ohm314, line 366: I don't quite understand what is the purpose of this class, do we really need it? Also, if it's really only needed for the `data_converter<std::vector<Reference.>`, why not make it an inner class?

include/highfive/bits/H5Converter_misc.hpp
ohm314: I would still call `first` actually `begin` to conform more to typical naming...?

ohm314 (commented):
Very nice!
Here are some first remarks:

include/highfive/bits/H5Reference_misc.hpp
alkino: There is a problem of includes. And I think it's better, at the end, there will be the `H5::Reference` class only declared and used here.

include/highfive/bits/H5Converter_misc.hpp
alkino: Little by little it will be used by all classes. And data_converter will be deleted.

Issue #387, opened 2020-10-20
CMake error dependencies


tdegeus:
I have a project with 

```bash
CMakeLists.txt
foo/CMakeLists.txt
bar/CMakeLists.txt
```

with `foo` and `bar` are made to work on its own, both include

```cmake
find_package(HighFive REQUIRED)
```

On the top simply have:

```cmake
if (foo)
    add_subdirectory(foo)
endif()

if (bar)
    add_subdirectory(bar)
endif()
```

With this I get the following error:

```none
CMake Error at /home/tdegeus/miniconda3/envs/code_velocity_s6g1/share/HighFive/CMake/HighFiveTargetDeps.cmake:5 (add_library):
  add_library cannot create target "libdeps" because another target with the
  same name already exists.  The existing target is an interface library
  created in source directory
  "/some/path/foo".  See documentation for
  policy CMP0002 for more details.
Call Stack (most recent call first):
  /home/tdegeus/miniconda3/envs/code_velocity_s6g1/share/HighFive/CMake/HighFiveConfig.cmake:67 (include)
  bar/CMakeLists.txt:31 (find_package)
```

matz-e:
Looks like we need some sort of include guard?

tdegeus:
Thanks @matz-e , I guess that https://github.com/BlueBrain/HighFive/pull/389 should do it.

PR #389, opened 2020-10-21
Adding CMake include guard

tdegeus:
Fixes https://github.com/BlueBrain/HighFive/issues/387

### DIFF
--- a/CMake/HighFiveTargetDeps.cmake
+++ b/CMake/HighFiveTargetDeps.cmake
@@ -1,79 +1,82 @@
 # Link against target system libs
 # -------------------------------
 
-# Independent target to make it possible to have new dependencies each build
-add_library(libdeps INTERFACE)
+if(NOT TARGET libdeps)
 
-# HDF5
-if(NOT DEFINED HDF5_C_LIBRARIES)
-  set(HDF5_NO_FIND_PACKAGE_CONFIG_FILE TRUE)  # Consistency
-  set(HDF5_PREFER_PARALLEL ${HIGHFIVE_PARALLEL_HDF5})
-  find_package(HDF5 REQUIRED)
-endif()
+  # Independent target to make it possible to have new dependencies each build
+  add_library(libdeps INTERFACE)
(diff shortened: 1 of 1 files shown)


CMake/HighFiveTargetDeps.cmake
```diff
@@ -1,79 +1,82 @@
 # Link against target system libs
 # -------------------------------
 
-# Independent target to make it possible to have new dependencies each build
-add_library(libdeps INTERFACE)
+if(NOT TARGET libdeps)
```
matz-e, line 4: This makes me wonder… if I declare my own target `libdeps` (the name seems pretty generic), and *then* include HighFive, wouldn't this lead to collision / failure, too?

tdegeus: Sure. This is the reason I have always been favour of having a makeshift namespace. Firstly, one should:
* `libdeps` -> `HighFive::dependecies`
Secondly, one could allow a finer precision:
* `HighFive::hdf5`
* `HighFive::use_Eigen`
* `HighFive::use_xtensor`
* `HighFive::use_Boost`

matz-e: That seems reasonable to me… also found this:

https://cmake.org/cmake/help/v3.10/command/include_guard.html?highlight=include_guard

unfortunately pretty late, only in 3.10 onwards.

tdegeus: `include_guard` seem nice. Just as good as the if though. Let me switch to `HighFive::dependencies` here, and then to the more refined things in separate PR.

tdegeus: That change is going to take a bit more work than I hoped. Personally I find HighFive's CMake support a mess... Anyway, let's do the bugfix now and improvement separately.

tdegeus:
Ping @matz-e @alkino @ferdonline

alkino (approved):

PR #391, opened 2020-10-23
Undefine max macro on windows in order to compile with msvc

txstc55:
Windows MSVC may have max defined as a macro, undefine this macro to use the max function in std

### DIFF
--- a/include/highfive/bits/H5DataType_misc.hpp
+++ b/include/highfive/bits/H5DataType_misc.hpp
@@ -16,6 +16,10 @@
 #include <H5Ppublic.h>
 #include <H5Tpublic.h>
 
+// undefine the macro on windows platform
+#ifdef _WIN32
+#undef max
+#endif
 
 namespace HighFive {
 


alkino:
Thanks for this patch, I know this problem.
There is a nice solution to fix it by defining the macro `NOMINMAX`.

txstc55:
> Thanks for this patch, I know this problem.
> There is a nice solution to fix it by defining the macro `NOMINMAX`.

I tried to put a 
```c++
#define NOMINMAX
```
in the file without the undef stuff and failed, look at these stackoverflow issues though: [issue 1](https://stackoverflow.com/questions/13416418/define-nominmax-using-stdmin-max), [issue 2](https://stackoverflow.com/questions/4913922/possible-problems-with-nominmax-on-visual-c).

All in all, I was using this project as a cmake project, and it would be nice to use it out of box without extra compiler flag or messing with the code. So please consider incorporate a solution for msvc (doesn't have to be this pull request, anything would work)

PR #395, opened 2020-12-28
Update HighFiveTargetDeps.cmake

kerim371:
`NO_MODULE` expect that `hdf5_DIR` points to the dir containing `hdf5-config.cmake` which greatly simplifies finding hdf5.
I can't see a reasonable way to build HighFive on Windows without it


kerim371:
Cant understand how to specify HDF5_DIR in Travis ci when cmake configure

tdegeus:
@kerim371 , I think the best solution would be to make HighFive's CMake target modular, such that you have a choice to use whatever HighFive provides, *or* handle only HighFive's header and leave the rest up to user.

That being said, just to understand your original problem. You say that you experience problems on Windows. Although I'm not a Windows user, I do use HighFive in some Windows CIs without problems. Maybe you experience

kerim371:
@tdegeus hi,
I just understood that I can build HighFive master branch. So generally this pull request is useless until I plan to use HighFive with application that includes two different HDF5 libraries. In this case I would need to choose which HDF5 lib to wrap and there may some problems be revealed.
I decided to postpone this pull request until I start to work with such application (few month).

Issue #396, opened 2021-01-07
Using references


plusangel:
Dear all,

first of all, I would like to thank the author and the contributors for this amazing work.
C++ API for HDF5 is not that straightforward and your work is absolutely amazing.
I had tried other libraries too...

I work in a library for NDT (Non-Destructive Testing) testing that uses HDF5 under the hood.
I have to use extensive references and I had a hard time.

My suggestion is to include the following example in the relevant directory, to help future users on that.
I took one of your example scripts and I turned it to use references, like that:

```cpp
#include <iostream>
#include <string>
#include <vector>

#include <highfive/H5DataSet.hpp>
#include <highfive/H5DataSpace.hpp>
#include <highfive/H5File.hpp>
#include <highfive/H5Reference.hpp>

using namespace HighFive;

const std::string FILE_NAME("dataset_integer.h5");
const std::string SOURCE_INT_DATASET_NAME("source_dateset");
const std::string REFERENCE_DATASET_NAME("reference_dateset");
const size_t size_dataset = 20;

// create a dataset 1D from a vector of string
void write_dataset() {

  // we create a new hdf5 file
  File file(FILE_NAME, File::ReadWrite | File::Create | File::Truncate);

  // we create a new group
  Group group = file.createGroup("a_group");

  std::vector<int> data(size_dataset);
  for (size_t i = 0; i < data.size(); ++i) {
    data[i] = int(i);
  }

  // let's create a dataset of native integer with the size of the vector
  // 'data' inside the group
  DataSet dataset = group.createDataSet<int>(SOURCE_INT_DATASET_NAME, DataSpace::From(data));
  dataset.write(data);

  // create a reference to the dateset containing the integers
  Reference ref = Reference(group, dataset);
  std::vector<Reference> ref_container{ref};

  // in similar fashion, we store as dataset the reference that we want
  DataSet ref_set = group.createDataSet<Reference>(REFERENCE_DATASET_NAME, DataSpace::From(ref_container));
  ref_set.write(ref_container);
}

// read our data back
void read_dataset() {
  // we open the existing hdf5 file we created before
  File file(FILE_NAME, File::ReadOnly);

  // we load the group
  Group my_group = file.getGroup("a_group");

  // we load the dataset that contains the reference
  DataSet ref_dataset = my_group.getDataSet(REFERENCE_DATASET_NAME);

  // we load the vector of references
  std::vector<Reference> expected_references;
  ref_dataset.read(expected_references);

[...]

PR #397, opened 2021-01-16
An example using hdf5 references (#396)

plusangel:
I included an example to demonstrate how references work in HighFive. PR suggested by @tdegeus 

Thank you guys

### DIFF
--- a/src/examples/read_write_vector_dataset_references.cpp
+++ b/src/examples/read_write_vector_dataset_references.cpp
@@ -1,0 +1,93 @@
+/*
+ *  Copyright (c), 2017, Adrien Devresse
+ *
+ *  Distributed under the Boost Software License, Version 1.0.
+ *    (See accompanying file LICENSE_1_0.txt or copy at
+ *          http://www.boost.org/LICENSE_1_0.txt)
+ *
+ */
+#include <iostream>
+#include <string>
+#include <vector>
+
+#include <highfive/H5DataSet.hpp>
+#include <highfive/H5DataSpace.hpp>
+#include <highfive/H5File.hpp>
+#include <highfive/H5Reference.hpp>
+
+using namespace HighFive;
+
+const std::string FILE_NAME("dataset_integer.h5");
+const std::string SOURCE_INT_DATASET_NAME("source_dateset");
+const std::string REFERENCE_DATASET_NAME("reference_dateset");
+const size_t size_dataset = 20;
+
+// create a dataset 1D from a vector of string
+void write_dataset() {
+
+    // we create a new hdf5 file
+    File file(FILE_NAME, File::ReadWrite | File::Create | File::Truncate);
+
+    // we create a new group
+    Group group = file.createGroup("a_group");
+
+    std::vector<int> data(size_dataset);
+    for (size_t i = 0; i < data.size(); ++i) {
+        data[i] = int(i);
+    }
+
+    // let's create a dataset of native integer with the size of the vector
+    // 'data' inside the group
+    DataSet dataset = group.createDataSet<int>(SOURCE_INT_DATASET_NAME, DataSpace::From(data));
+    dataset.write(data);
+
+    // create a reference to the dateset containing the integers
+    Reference ref = Reference(group, dataset);
+    std::vector<Reference> ref_container{ref};
+
+    // in similar fashion, we store as dataset the reference that we want
+    DataSet ref_set = group.createDataSet<Reference>(REFERENCE_DATASET_NAME, DataSpace::From(ref_container));
+    ref_set.write(ref_container);
+}
+
+// read our data back
+void read_dataset() {
+    // we open the existing hdf5 file we created before
+    File file(FILE_NAME, File::ReadOnly);
+
+    // we load the group
+    Group my_group = file.getGroup("a_group");
+
+    // we load the dataset that contains the reference
+    DataSet ref_dataset = my_group.getDataSet(REFERENCE_DATASET_NAME);
+
+    // we load the vector of references
+    std::vector<Reference> expected_references;
+    ref_dataset.read(expected_references);
+
+    // we use the stored reference and dereference it to gain access in the integers' dataset
+    DataSet expected_dataset = expected_references[0].dereference<DataSet>(my_group);
+
+    // as usual, we load the vector with numbers from the extracted dataset
+    std::vector<int> read_data;
+    expected_dataset.read(read_data);
+
+    // and voila, the payload we excepted
+    for (size_t i = 0; i < read_data.size(); ++i) {
+        std::cout << read_data[i] << " ";
+    }
+}
+
+int main(void) {
+
+    try {
+        write_dataset();
+        read_dataset();
+
+    } catch (Exception &err) {
+        // catch and print any HDF5 error
+        std::cerr << err.what() << std::endl;
+    }
+
+    return 0;// successfully terminated
+}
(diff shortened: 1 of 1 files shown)


tdegeus (commented):
I'm probably not the best person to review, because I don't actually know references in HDF5. That being said, it didn't become clear from this example either... Some line of text would surely help there: describe what the example is doing in a few lines of comment somewhere. 

Where I really got lost though is with the `  std::vector<Reference>`. The question is: Are reference in HDF5 something similar to those in C++? If so, wouldn't it be easier to (also) have a basic example with one reference?

src/examples/read_write_vector_dataset_references.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include <iostream>
+#include <string>
+#include <vector>
+
+#include <highfive/H5DataSet.hpp>
+#include <highfive/H5DataSpace.hpp>
+#include <highfive/H5File.hpp>
+#include <highfive/H5Reference.hpp>
+
+using namespace HighFive;
```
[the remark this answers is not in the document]
ohm314: I agree, please remove the using namespace declaration. As @tdegeus said, this just makes things less clear and is generally a bad practice that should be avoided.

src/examples/read_write_vector_dataset_references.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+const std::string FILE_NAME("dataset_integer.h5");
+const std::string SOURCE_INT_DATASET_NAME("source_dateset");
```
ohm314, line 21:
```suggestion
const std::string SOURCE_INT_DATASET_NAME("source_dataset");
```

src/examples/read_write_vector_dataset_references.cpp
```diff
@@ -0,0 +1,93 @@
...
+const std::string REFERENCE_DATASET_NAME("reference_dateset");
```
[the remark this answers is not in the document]
ohm314: I don't mind either. Defining the strings doesn't hurt, but doesn't add much. Either is fine.

src/examples/read_write_vector_dataset_references.cpp
```diff
@@ -0,0 +1,93 @@
...
+const size_t size_dataset = 20;
+
+// create a dataset 1D from a vector of string
```
ohm314, line 25:
```suggestion
// create a dataset 1D from a vector of int
```

src/examples/read_write_vector_dataset_references.cpp
```diff
@@ -0,0 +1,93 @@
...
+void write_dataset() {
+
+    // we create a new hdf5 file
+    File file(FILE_NAME, File::ReadWrite | File::Create | File::Truncate);
+
+    // we create a new group
+    Group group = file.createGroup("a_group");
```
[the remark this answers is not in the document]
ohm314: why do you define const strings for the file name and dataset names, but not the group. I would be consistent, whatever you choose.

src/examples/read_write_vector_dataset_references.cpp
```diff
@@ -0,0 +1,93 @@
...
+    // let's create a dataset of native integer with the size of the vector
+    // 'data' inside the group
+    DataSet dataset = group.createDataSet<int>(SOURCE_INT_DATASET_NAME, DataSpace::From(data));
+    dataset.write(data);
+
+    // create a reference to the dateset containing the integers
+    Reference ref = Reference(group, dataset);
+    std::vector<Reference> ref_container{ref};
+
+    // in similar fashion, we store as dataset the reference that we want
```
ohm314, line 48:
```suggestion
    // in similar fashion, we store as dataset the vector of references that we want
```

ohm314 (changes requested):
Hi @plusangel, sorry for taking so long to review this. Overall the example is small enough and a good idea to add. We actually have in the unit tests already something very similar (but I understand that looking at test cases is a pain, especially when you're just getting into the library).
@tdegeus gave already some very good feedback. Could I ask you to please implement his and my suggestions into an additional patch. I"ll try to then re-review fairly quickly and have this merged!
Here's the test case, for reference: https://github.com/BlueBrain/HighFive/blob/master/tests/unit/tests_high_five_base.cpp#L1587

plusangel:
Thank you both for your valuable comments and your time to review it. I think I catch them all. Keep up the good work. Thank you!

src/examples/read_write_vector_dataset_references.cpp
```diff
@@ -0,0 +1,97 @@
+/*
+ *  Copyright (c), 2017, Adrien Devresse
```
ferdonline, line 2:
```suggestion
 *  Copyright (c), 2021 Blue Brain Project - EPFL
```

src/examples/read_write_vector_dataset_references.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include <highfive/H5DataSet.hpp>
+#include <highfive/H5DataSpace.hpp>
+#include <highfive/H5File.hpp>
+#include <highfive/H5Reference.hpp>
+
+// const std::string FILE_NAME("dataset_integer.h5");
+//const std::string SOURCE_INT_DATASET_NAME("source_dateset");
+//const std::string REFERENCE_DATASET_NAME("reference_dataset");
+//const size_t size_dataset = 20;
+
```
ferdonline, line 22:
```suggestion
```

src/examples/read_write_vector_dataset_references.cpp
```diff
@@ -0,0 +1,97 @@
...
+// create a dataset 1D from a vector of int
+void write_dataset() {
+
+    // we create a new hdf5 file
+    HighFive::File file("dataset_integer.h5", HighFive::File::ReadWrite |
+                                                  HighFive::File::Create |
+                                                  HighFive::File::Truncate);
```
ferdonline, line 29:
```suggestion
    HighFive::File file("dataset_integer.h5", HighFive::File::Overwrite);
```

src/examples/read_write_vector_dataset_references.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+    // we create a new group
+    HighFive::Group group = file.createGroup("a_group");
+
+    std::vector<int> data(20);
+    for (size_t i = 0; i < data.size(); ++i) {
+        data[i] = int(i);
+    }
```
ferdonline, line 37:
```suggestion
    std::vector<int> data(20);
    std::iota(data.begin(), data.end(), 0);
```

src/examples/read_write_vector_dataset_references.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+    // let's create a dataset of native integer with the size of the vector
+    // 'data' inside the group
+    HighFive::DataSet dataset = group.createDataSet<int>("source_dateset",
+                                                         HighFive::DataSpace::From(data));
+    dataset.write(data);
```
ferdonline, line 43:
```suggestion
    auto dataset = group.createDataSet("source_dataset", data);
```

src/examples/read_write_vector_dataset_references.cpp
```diff
@@ -0,0 +1,97 @@
...
+        write_dataset();
+        read_dataset();
+
+    } catch (HighFive::Exception& err) {
+        // catch and print any HDF5 error
+        std::cerr << err.what() << std::endl;
+    }
+
+    return 0;  // successfully terminated
+}
```
ferdonline, line 97:
```suggestion
}

```

ferdonline (changes requested):

ferdonline:
@plusangel Do you have an estimate on when you could address this PR suggestions?

plusangel:
Sorry for the delay @ferdonline. Done :)

ferdonline (dismissed):

tdegeus (approved):

ohm314 (approved):
lgtm

PR #403, opened 2021-02-13
Uploading docs to gh-pages using GitHub Actions

tdegeus:
Since recently I have been using GitHub Actions to build my docs and upload them to GitHub pages. I set this now to be done when something is pushed to the master, but it could also be done upon releasing.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -1,4 +1,4 @@
-name: Doxygen -> gh-pages
+name: gh-pages
 
 on:
   push:
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -42,5 +42,7 @@ jobs:
       with:
         target_branch: gh-pages
         build_dir: build/doc/html
+        jekyll: false
+        keep_history: false
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -1,0 +1,46 @@
+name: Doxygen -> gh-pages
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+
+  publish:
+
+    runs-on: ubuntu-latest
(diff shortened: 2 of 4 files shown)


matz-e (dismissed):
Thanks, this is very nice!

I was just checking the documentation… and was kind of missing the "customary" documentation badge on top of the readme. Does Github actions provide us with such a thing?

Also reminds me that we should move away from Travis, too…

tdegeus:
@matz-e Indeed GitHub supports badges, and it's a good idea to have it. It needed a minor rename, but it is included here!

See https://github.com/tdegeus/cppcolormap where I verified that it works

tdegeus:
BTW, I moved to GH Action for CI on almost all my projects. I mussed say, it makes a huge difference. I can highly recommend the use of conda (particularly of conda-forge and mamba) for ease of use an maintainability.

matz-e (approved):
Nice, thanks again!

I like Github actions, although I've mostly used it for my Rust explorations. Pretty easy to set up…

Seems to me that you use conda like a Python virtualenv here. I'm wondering if for our use case here this has advantages over OS-specific package managers… i.e., can we run with different versions of HDF5 (1.10, 1.12) with conda using the Github Action matrix?

tdegeus:
@matz-e It is true that using different package managers could be more efficient in the case that you want to use different library versions. Indeed, you can do this in the build matrix with conda, but it will probably involve removing/reinstalling. 

My remark indeed derives from the case that you use a single version of the library. In that case conda is very easy because its syntax is platform independent, and because you can test the environment also locally (and it gives you a way for future users to establish the same environment easily).

tdegeus:
@matz-e With my other sites I noticed that I missing an option to the action that I fixed here. The option takes care of always having the `.nojekyll` file to avoid GitHub to ignore files starting with `_` (generated a lot by doxygen). Furthermore I included a default option just as a reminder that the gh-pages branch is replaced with the new doxygen output (i.e. the action create, replaces, and removes files)

matz-e (approved):

PR #404, opened 2021-02-13
Avoiding GitHub to ignore files that start with an underscore

tdegeus:
See https://github.community/t/im-get-a-404-file-not-found-on-my-github-pages-for-files-starting-with-an-underscore/160564/2


matz-e (approved):
Seems legit. We use doxygen, anyways…

PR #414, opened 2021-03-03
Fix use before initialization

alkino:
I extracted this fix from the PR #400 of @kerim371


tdegeus:
Am I right to conclude that the same bug occurs also here:

https://github.com/BlueBrain/HighFive/blob/90e6316c171861fb925e7f995533114a4d206614/include/highfive/bits/H5Converter_misc.hpp#L297

alkino:
> Am I right to conclude that the same bug occurs also here:
> 
> https://github.com/BlueBrain/HighFive/blob/90e6316c171861fb925e7f995533114a4d206614/include/highfive/bits/H5Converter_misc.hpp#L297

No before of the use of "insert"

tdegeus (approved):

matz-e (approved):

PR #416, opened 2021-03-08
Introducing GitHub Actions CI

### LINKED ISSUE #402: Move to Github Actions
As other projects in the BlueBrain organization are moving to Github actions, so should this repository…

ferdonline:
**Description**

This PR introduces  GitHub actions-based CI

The strategy is to test for several Linux (Ubuntu) systems, for several compilers, Mac and Windows
Total: 10 jobs

**Combinations**
```
Linux_MPI:
    os: [ubuntu-16.04, ubuntu-18.04, ubuntu-20.04]
    compiler: (default/old)
    options: PARALLEL_HDF5 + XTENSOR + EIGEN

Linux_Compilers:
    runs-on: ubuntu-18.04
    compilers: [clang-8, clang-9, gcc-8, gcc-9,  gcc-10]
    options: EIGEN

MacOS: (default compiler + EIGEN)

Windows: (Default compiler + EIGEN + Xtensor)
```

Fixes #402

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,77 @@
+name: CMake
+
+on:
+  push:
+    branches:
+      - live-debug*
+      - ci_test
+      - release/**
+  pull_request:
+    branches:
+      - master
+      - release/**
+    paths-ignore:
+      - '**.md'
+      - '**.rst'
+      - 'docs/**'
+
+env:
+  HOMEBREW_NO_AUTO_UPDATE: 1  # for reproducibility, dont autoupdate
+  BUILD_TYPE: RelWithDebInfo
+  INSTALL_DIR: install
+  HIGHFIVE_USE_XTENSOR: False
+  HIGHFIVE_USE_OPENCV: False
+  HIGHFIVE_PARALLEL_HDF5: False
+  MINCONDA_VERSION: latest
+  MINCONDA_LINUX: Linux-x86_64
+  MINCONDA_OSX: MacOSX-x86_64
+
+jobs:
+  CI:
+    runs-on: ${{matrix.os}}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, ubuntu-20.04, ubuntu-latest]
+        compilers_eval: ["", "CC=clang-8 CXX=clang++-8"]
+        include:
+          - os: macos-10.15
+          - os: ubuntu-16.04
+          - os: ubuntu-18.04
+            compilers_eval: "CC=gcc-8 CXX=g++-8"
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,77 @@
+name: CMake
+
+on:
+  push:
+    branches:
+      - live-debug*
+      - ci_test
```
alexsavulescu, line 7:
```suggestion
      - master
```

ferdonline: I think it is a good idea to leave master out, since push to master is only allowed with PRs, which are already tested.
having a special `ci_test` is debatable, the idea is to have a special branch so that we run CI every time we push, which is useful mostly to develop the CI itself. Do you suggest a better way?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,77 @@
...
+      - '**.rst'
+      - 'docs/**'
+
+env:
+  HOMEBREW_NO_AUTO_UPDATE: 1  # for reproducibility, dont autoupdate
+  BUILD_TYPE: RelWithDebInfo
+  INSTALL_DIR: install
+  HIGHFIVE_USE_XTENSOR: False
+  HIGHFIVE_USE_OPENCV: False
+  HIGHFIVE_PARALLEL_HDF5: False
```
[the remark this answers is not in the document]
ferdonline: not yet! :)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,77 @@
...
+  MINCONDA_VERSION: latest
```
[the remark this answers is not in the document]
ferdonline: Same as bellow. For future usage. Will drop it

.github/workflows/ci.yml
alexsavulescu: I'd say it's standard practice to run CI against `master`. It's true it would virtually be redundant wrt PRs, regardless of merge strategy. Additional steps can be done if `isMaster()` like storing artifacts, publishing documentation, displaying CI badges and so on. I guess we can look at it also as a sanity check. 

> having a special `ci_test` is debatable, the idea is to have a special branch so that we run CI every time we push, which is useful mostly to develop the CI itself. Do you suggest another way? @alexsavulescu

The CI will run for each PR every time a new commit is pushed even when working on the CI, so there is no need for a special branch. To work on the CI itself, the most straightforward (and isolated way) is to use a fork, where you can modify `master` directly or even use PRs (if you don't want to add `master` branch on push).

alexsavulescu (approved):
LGTM

Issue #401, opened 2021-02-01
Support checking if two objects target the same thing


kerim371:
Hi,

I'm looking for a way to compare objects. For example I want to know whether two groups are the same and reffered to the same file?

I think it used to be possible with `ObjectInfo::getAddress()` but it is deprecated for now.

matz-e:
Hi, would it possible for you to just compare the `Group::getId()` between the two groups/objects?

kerim371:
@matz-e unfortunately the same `Group` (or any object) stored in different variables gives different `id`

alkino:
What is the good way to compare two objects with different ifs following you?

kerim371:
@alkino there is two types of objects:
1) usually stored in file (`File`, `Group`, `DataSet`)
2) other like `DataSpace`, `DataType`

For the `File` I think the best way is to compare whether the file is the same? Since HDF5 1.12.0 [there is a function](https://portal.hdfgroup.org/display/HDF5/H5F_GET_FILENO) wich (relying on documentation and my test but not 100% sure) can be used to compare files: `herr_t H5Fget_fileno (hid_t file_id, unsigned long *fnumber )` 
For older HDF5 versions probably we need to compare filenames wich is inefficient but I can't see another way.

For `Group`, `DataSet` and `Attribute` we may compare do they have the same address and if they do then also compare do they belong to the same file? 
Here the problem that retrieved address as a [deprecated struct member](https://portal.hdfgroup.org/display/HDF5/H5O_GET_INFO1) of `H5Oget_info1()`. But I've found the [question on HDF5 forum](https://forum.hdfgroup.org/t/check-if-two-identifiers-refer-to-the-same-object/7160) posted last year where author tells that it possible to use deprecated address.

For `DataSpace` we can compare the dimensionality. So there is no need to compare the address because usually (or always) `DataSpace` doesn't belong to any file.

And `DataType` already has overloaded `==` and `!=` operators

alkino:
I think, it's a bad idea, because we will never be sure. If it's not possible by design in hdf5, we will do bad things.

kerim371:
@alkino lately I asked similar question on [HDF5 forum](https://forum.hdfgroup.org/t/h5l-info2-t-me-meaning-of-member-variables/8162/2?u=kerim.khemraev) and I think that this could be done via token.
Though I have not tested it yet

ferdonline:
[...]

PR #417, opened 2021-03-11
Generalizing getPath and adding getFile as PathTraits

ferdonline:
**Problem**
1. `getFile()` would be useful to get back the file object from any Group/Dataset/Attribute.
2. `getPath` was duplicated in some entities.
3. With MPI-IO, closing a file invalidates open objects.

**This PR**
Addresses all three problems by creating a PathTraits class and applying it to Group, Dataset and Attribute.
Trait keeps a file obj ref so that it's count never drops to 0.

*Extra Required fixes*:
 - Impl `operator==` in object
 - Annotate Traits to not construct an empty `Attribute`, but use `Object(hid)` ctr instead
 - Drop Clang tests in Travis since we'r covering all of them in the new GitHub Actions CI. Kept only the very old systems tests

### DIFF
--- a/include/highfive/H5Object.hpp
+++ b/include/highfive/H5Object.hpp
@@ -63,6 +63,11 @@ class Object {
     ///
     ObjectType getType() const;
 
+    // Check if refer to same object
+    bool operator==(const Object& other) const noexcept {
+        return _hid == other._hid;
+    }
+
   protected:
     // empty constructor
     Object();
@@ -76,6 +81,7 @@ class Object {
     // Init with an low-level object id
     explicit Object(hid_t);
 
+    // Copy-Assignment operator
     Object& operator=(const Object& other);
 
     hid_t _hid;
--- a/include/highfive/bits/H5Annotate_traits_misc.hpp
+++ b/include/highfive/bits/H5Annotate_traits_misc.hpp
@@ -25,14 +25,14 @@ inline Attribute
 AnnotateTraits<Derivate>::createAttribute(const std::string& attribute_name,
                                           const DataSpace& space,
                                           const DataType& dtype) {
-    Attribute attribute;
-    if ((attribute._hid = H5Acreate2(
-             static_cast<Derivate*>(this)->getId(), attribute_name.c_str(),
-             dtype._hid, space._hid, H5P_DEFAULT, H5P_DEFAULT)) < 0) {
+    auto attr_id = H5Acreate2(static_cast<Derivate*>(this)->getId(),
+                              attribute_name.c_str(),
+                              dtype._hid, space._hid, H5P_DEFAULT, H5P_DEFAULT);
+    if (attr_id < 0) {
         HDF5ErrMapper::ToException<AttributeException>(
             std::string("Unable to create the attribute \"") + attribute_name + "\":");
     }
-    return attribute;
+    return Attribute(attr_id);
 }
 
 template <typename Derivate>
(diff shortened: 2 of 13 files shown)


include/highfive/H5Object.hpp
```diff
@@ -63,6 +63,11 @@ class Object {
     ///
     ObjectType getType() const;
 
+    // Check if refer to same object
+    bool operator==(const Object& other) const noexcept {
+        return _hid == other._hid;
+    }
+
```
alkino, line 70: https://github.com/BlueBrain/HighFive/issues/401 https://forum.hdfgroup.org/t/h5l-info2-t-me-meaning-of-member-variables/8162/2?u=kerim.khemraev
objects with different hid can be the same.

include/highfive/bits/H5Annotate_traits_misc.hpp
```diff
@@ -25,14 +25,14 @@ inline Attribute
...
-             dtype._hid, space._hid, H5P_DEFAULT, H5P_DEFAULT)) < 0) {
+    auto attr_id = H5Acreate2(static_cast<Derivate*>(this)->getId(),
+                              attribute_name.c_str(),
+                              dtype._hid, space._hid, H5P_DEFAULT, H5P_DEFAULT);
+    if (attr_id < 0) {
         HDF5ErrMapper::ToException<AttributeException>(
             std::string("Unable to create the attribute \"") + attribute_name + "\":");
     }
-    return attribute;
+    return Attribute(attr_id);
```
alkino, line 35: nicier !

include/highfive/H5Object.hpp
ohm314: I'm not sure I understand the comment.. but does that mean one should call https://portal.hdfgroup.org/display/HDF5/H5I_OBJECT_VERIFY for `_hid` and `other._hid` and compare the addresses?

alkino: I don't know, I'm asking what does it mean for user `Obj1 == Obj2` ?

include/highfive/bits/H5Path_traits_misc.hpp
```diff
@@ -0,0 +1,29 @@
...
+#include "H5Utils.hpp"
+#include "H5Path_traits.hpp"
+
+namespace HighFive{
+
+template <typename Derivate>
+inline PathTraits<Derivate>::PathTraits() {
+    const hid_t hid = static_cast<Derivate*>(this)->getId();
+    hid_t file_id = H5Iget_file_id(hid);
+    if (file_id > 0 and file_id != hid) {
```
ohm314, line 12:
```suggestion
    if ((file_id > 0) && (file_id != hid)) {
```
https://github.com/BlueBrain/HighFive/pull/417/checks?check_run_id=2085551915#step:5:98
maybe this will appease visual studio compiler...

ohm314 (commented):
Looks great, thank you! A small fix is needed to get it built on windows too.

include/highfive/H5Object.hpp
ferdonline: > objects with different hid can be the same.

But I wonder if we should say they are the same... Even if they point to the same thing in disk, I think in the general case it's good to compare the hid. If, in the case of datasets, we can do a better comparison, sure. But remember this definition is shared among any kind of hdf5 object.

include/highfive/bits/H5Annotate_traits_misc.hpp
ferdonline: I had to do it anyway since constructing an empty Attribute would lead to problems (since it can't get the parent).
That makes me think we should delete empty constructor in these cases.

include/highfive/bits/H5Path_traits_misc.hpp
ferdonline: 😆 too much python :p

ohm314 (approved):
lgtm (so clearly I did not understand the error message of vscode ;) )

Issue #357, opened 2020-07-13
Add function to link a DataSet to another location (Hard Links)


L0ric0:
Edit: The `hdf5` library has a function `link` which takes a type and two keys. This function links the first key to the location of the second. It would be nice to see such a function in HighFive so that one doesn't have to fall back to raw HDF5

Is it possible to link a DataSet to another location in the hdf5 file with this library?

matz-e:
Yes, see [here](https://github.com/BlueBrain/HighFive/blob/master/src/examples/renaming_objects.cpp#L43).

L0ric0:
well hdf5 supports `file->link(H5L_TYPE_HARD, original_key, link_key);` if that is not implemented then i will rewrite this issue as a feature request

matz-e:
Please go ahead. I think I dropped to bare HDF5 before when needing to link like that.

1uc:
@Quark-X10 thanks for implementing the feature.

PR #421, opened 2021-03-21
Soft | External links (minimal API)

### LINKED ISSUE #154: Ability to creating Soft link
Is it possible to create a soft-link to a dataset using HighFive? 
I couldn't find such feature in current release.

Imagine it could do:
```c++
rootGroup.createSoftLink(std::string name,const HighFive::DataSet & targetDataSet)
or
rootGroup.createSoftLink(std::string name, std::string pathToDataSetObject /* /group1/group2/dataset*/)
```

Thanks in advance for any help.

kerim371:
**Description**

- [x] Added minimal API to create Soft and External links

Fixes #154

**How to test this?**

See tests HighFiveSoftLink, HighFiveardHLink, HighFiveExternalLink

```bash
cmake ..
make -j8
make test
```

**Test System**
 - OS: [e.g. Ubuntu 20.04]


kerim371:
Test fails beacause file doesn't return its name.

I'm just interested, why you decided to keep `_filename` private variable instead of using hdf5 functions? 
And why you didn't implement `getFileName()` from my previous PR that allows to get file name directly from objects without doing `getFile().getName()` (hdf5 provides funtion to do that)?

ohm314:
Hi @kerim371

Thanks for your PR! Regarding your question on filename. As @ferdonline explained [here](https://github.com/BlueBrain/HighFive/pull/400#issuecomment-790511578) we actually follow the example of h5py by providing access to the file object from everywhere, which ones its name. We believe this is cleaner as the name is a property of the file and not of the group or dataset.
Can you run the tests locally to debug what is going on exactly for the failing tests?

kerim371:
@ohm314 Hi,

I'm sorry I'm completely disagree with the philosophy of filename implemented in HighFive. If I modify something connected with filename I believe I will change it in my point of view wich is different from yours. You will not accept that.

I think the problem is that when File is initialized from `object.getFile()` the variable `filename` is not initialized.

ferdonline:
> I'm just interested, why you decided to keep _filename private variable instead of using hdf5 functions?

That change was simply not addressed yet. There's a generic helper function to read names which handles the size limit problem (`std::string details::get_name(T fct)`)

include/highfive/H5Object.hpp
```diff
@@ -32,6 +32,16 @@ enum class ObjectType {
...
+///
+/// \brief The possible types of group entries (link concept)
+///
+enum class LinkType {
+    Hard,
+    Soft,
+    External,
+    Other  // Reserved or User-defined
+};
+
```
ferdonline, line 44: Is there a good reason to move "LinkType" to another header? Please don't introduce breaking changes, unless it's last resort.

include/highfive/bits/H5Node_traits.hpp
```diff
@@ -148,6 +148,23 @@ class NodeTraits {
     /// \param node_name The entry to check, path relative to the current group
     inline ObjectType getObjectType(const std::string& node_name) const;
 
+    template<typename Node,
+            typename std::enable_if<
+                std::is_same<Node, File>::value |
+                std::is_same<Node, Group>::value>::type* = nullptr>
+    Group createLink(
```
ferdonline, line 155: I find `static_assert` better where it can be used. Readability and debugging ease

ferdonline: Please pass `.clang-format`. style here is not consistent with rest

ferdonline (commented):
Some minimal remarks. I'd like anyway to discuss API. I'm not so convinced by handling all link types in the same function.

ferdonline:
Regarding the dynamic filename, there you go: https://github.com/BlueBrain/HighFive/pull/424

include/highfive/H5Object.hpp
kerim371: I think for now it can reside in its usual header. I will try to move it back.

include/highfive/bits/H5Node_traits.hpp
kerim371: Agree with clang format.
`enable_if` allows you to check whether your type is apropriate when you write code. If you try to pass inapropriate type then you will see that this line of code is underlined by red as I mistake. `static_assert` only throws error at a compile time so it is always a confusion what classes are accepted.

kerim371:
@ferdonline I find the most actual questions are:
1) what API and for what link type you would like to see. For now link type is defined as an enum argument to a function. Probably there are people who would like to see API like `createSoftLink(...)` and `createExternalLink(...)`. It is fine but in this case we woulod need to declare 6 new methods differentiated by return type (`Group createSoftLink(...)`, `File createSoftLink(...)`, `DataSet createSoftLink(...)` etc with external link). Previously (locally) I've tried to reduce the number of these methods by involving templates according to return type but I didn't succed.
2) Do I need to use `static_assert` instead of `enable_if`. As for me `enable_if` is easier to use beacause you can always see what types are appropriate to the function/class. If it is inapropriate then in your editor you can see underlined line as error. `static_assert` as I know throws error only at a compile time and it is difficult to guess what types are acceptable.

ferdonline:
Hi @kerim371 The best is probably to have independent API calls. You can see that due to all `ifs` it really feels like it's three functions in one, which doesn't feel right. Also, maybe it's not worth doing too much magic, i.e., soft and external links may receive the path as a string instead of an actual object, that will simplify and is good enough, and (I also like that) is similar to h5py.

kerim371:
@ferdonline ok, I'm busy a little bit now, I will change it soon.

kerim371:
Waiting for comments

include/highfive/bits/H5Node_traits_misc.hpp
```diff
@@ -97,14 +97,14 @@ NodeTraits<Derivate>::createDataSet(const std::string& dataset_name,
...
-                                 const DataSetAccessProps& accessProps) const {
-    const auto hid = H5Dopen2(static_cast<const Derivate*>(this)->getId(),
-                              dataset_name.c_str(), accessProps.getId());
-    if (hid < 0) {
-        HDF5ErrMapper::ToException<DataSetException>(
-            std::string("Unable to open the dataset \"") + dataset_name + "\":");
-    }
-    return DataSet(hid);
+                                const DataSetAccessProps& accessProps) const {
+   const auto hid = H5Dopen2(static_cast<const Derivate*>(this)->getId(),
```
ferdonline, line 101: I guess this reformat was a mistake. large diff and against our coding style... We can't accept changes in this state.

include/highfive/bits/H5Node_traits.hpp
```diff
@@ -152,6 +152,17 @@ class NodeTraits {
     /// \param node_name The entry to check, path relative to the current group
     inline ObjectType getObjectType(const std::string& node_name) const;
 
+    void createSoftLink(const std::string& objPath,
```
ferdonline, line 155: I don't see the definition/impl of hard links as mentioned by this PR

ferdonline (changes requested):

include/highfive/bits/H5Node_traits.hpp
kerim371: @ferdonline agree but I thought that there is no need to add hard link as hard links are those that are created in usual way: for example `file.createGroup(...)` should create hard link to the group (if I'm not mistaken). How do you think, should I add method `createHardLink(...)`?

ferdonline: Hard links are useful to add additional paths to datasets, or even move, see #357
Anyway its implementation can be a little bit more complicated and its urgency might be lower. So we will go with soft and external, while Hard links will be implemented in another PR.

ferdonline:
Sorry @kerim371 this was blocking the release so I went on to clean it up

ohm314 (approved):
This looks good to me!

Issue #419, opened 2021-03-17
Revise Property lists for completeness


ferdonline:
Property Lists have been made high level (template class `PropertyList`) and for each list type an alias has been created.

It seems however that some property lists are still missing, e.g. for `PropertyType::LINK_CREATE`.
Such property lists seem more important for link creating, which has been a long standing request (#154 #357)

https://support.hdfgroup.org/HDF5/doc/RM/RM_H5P.html#LinkCreatePropFuncs
- [ ] H5P_ATTRIBUTE_CREATE
- [ ] H5P_DATASET_ACCESS
- [ ] H5P_DATASET_CREATE
- [ ] H5P_DATASET_XFER
- [ ] H5P_DATATYPE_ACCESS
- [ ] H5P_DATATYPE_CREATE
- [ ] H5P_FILE_ACCESS
- [ ] H5P_FILE_CREATE
- [ ] H5P_FILE_MOUNT
- [ ] H5P_GROUP_ACCESS
- [ ] H5P_GROUP_CREATE
- [ ] H5P_LINK_ACCESS
- [ ] H5P_LINK_CREATE
- [ ] H5P_OBJECT_COPY
- [ ] H5P_OBJECT_CREATE
- [ ] H5P_STRING_CREATE

Note: This issue has been created following the discussion in #400

PR #423, opened 2021-03-26
Add property CreateIntermediateGroup

alkino:
Add a new property in the high level class of properties.
As documented in https://support.hdfgroup.org/HDF5/doc/RM/RM_H5P.html#Property-SetCreateIntermediateGroup
This property let you decide if missing parent's groups should be created when creating a group or a dataset.
As it is already use internally, I though it was a good idea to add it.

This is to complete a bit : #419

### DIFF
--- a/include/highfive/H5PropertyList.hpp
+++ b/include/highfive/H5PropertyList.hpp
@@ -74,11 +74,13 @@ class PropertyList {
     hid_t _hid;
 };
 
+typedef PropertyList<PropertyType::OBJECT_CREATE> ObjectCreateProps;
 typedef PropertyList<PropertyType::FILE_CREATE> FileCreateProps;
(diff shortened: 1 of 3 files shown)


matz-e (approved):

tdegeus:
Does this mean that https://github.com/BlueBrain/HighFive/blob/739c61484071b08feb54a6f360447daa55668e16/include/highfive/h5easy_bits/H5Easy_misc.hpp#L44 can be deprecated ?

alkino:
Why?

You should have fix the createDataSet API instead of creating your own function to do the job ^^

tdegeus:
If I understand well, currently HighFive *can* create all groups needed to create a dataset ?

Agreed. I'm not sure why we never did that at the time I implemented this.

alkino:
It's not highfive it's HDF5. It's not yet visible in the `createDataSet` of HighFive, but it needs only to expose the option.

PR #424, opened 2021-03-26
Implementing getting the filename dynamically

ferdonline:
**Description**

This enables recreating file objects and still being able to
call getName()
_filename becomes a cache variable, so we dont change API nor ABI

**Fixes**
Issue when recreating file object from any dataset/group ->getFile()

**How to test this?**
ctest (already covered, just reimplemented)


ohm314 (approved):
Looks good!

matz-e (approved):

PR #425, opened 2021-03-30
Add option *parents* to createDataSet

alkino:
This way the parents group can be created automatically if they don't exist.

Deprecate the function *createGroupsToDataSet* inside the easy part and stop using it internally

### DIFF
--- a/include/highfive/bits/H5Node_traits.hpp
+++ b/include/highfive/bits/H5Node_traits.hpp
@@ -37,7 +37,8 @@ class NodeTraits {
                   const DataSpace& space,
                   const DataType& type,
                   const DataSetCreateProps& createProps = DataSetCreateProps(),
-                  const DataSetAccessProps& accessProps = DataSetAccessProps());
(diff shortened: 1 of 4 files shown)


tdegeus:
Amazing ! 

Since `createGroupsToDataSet` was in the `detail` namespace I would say that it could be removed immediately. Alternatively it can be kept for a while but with a compiler warning using ``[[ deprecated ]]``

alkino:
> Amazing !
> 
> Since `createGroupsToDataSet` was in the `detail` namespace I would say that it could be removed immediately. Alternatively it can be kept for a while but with a compiler warning using `[[ deprecated ]]`

Don't ask me twice to delete dead code!

tdegeus:
> Don't ask me twice to delete dead code!

In that case... I think that with this change `getParentName` can go too. It was only used by `createGroupsToDataSet` and also in the `detail` namespace. So unless that is a which for it to be part of the public API I think that it can be removed!

PR #428, opened 2021-03-31
Fix for unexpected issue in HDF5 with ASCII reading

sergiorg-hpc:
**Description**

When reading ASCII-based datasets without first accessing other datasets, HDF5 produces unexpected converter issues. The patch provides the following fixes:

- [x] Hides HDF5 string converter issues when memory and dataset types do not match, by forcing the character set of the strings to the one available in the dataset.

The issue was reported here https://github.com/BlueBrain/libsonata/issues/139.

**How to test this?**
Through `libsonata`, it can be easily tested with Python:

```python
import libsonata
filename = "/gpfs/bbp.cscs.ch/project/proj12/jenkins/cellular/circuit-1k/nodes.h5"
pop = libsonata.NodeStorage(filename).open_population("All")
pop.get_attribute('region', libsonata.Selection((1,2)) )
```

Alternatively, the following C++ source code example can be compiled as well:

```C++
vector<std::string> buffer;
HighFive::File file("/gpfs/bbp.cscs.ch/project/proj12/jenkins/cellular/circuit-1k/nodes.h5");
file.getDataSet("/nodes/All/0/@library/region").read(buffer);
```

Without the fix, HighFive will throw several exceptions.

**Test System**
 - OS: RHEL v7.6
 - Compiler: GCC 9.3.0
 - Dependency versions: Tested with HDF5 v1.10.7 and v1.12.0

### DIFF
--- a/include/highfive/bits/H5ReadWrite_misc.hpp
+++ b/include/highfive/bits/H5ReadWrite_misc.hpp
@@ -42,9 +42,9 @@ inline static DataType getDataType(const DataType& element_type, const DataType&
     // TEMP. CHANGE: Ensure that the character set is properly configured to prevent
     // converter issues on HDF5 <=v1.12.0 when loading ASCII strings first.
     // See https://github.com/HDFGroup/hdf5/issues/544 for further information.
-    const auto dtype_cset = H5Tget_cset(dtype.getId());
-    if (H5Tget_class(element_type.getId()) == H5T_STRING && dtype_cset == H5T_CSET_ASCII) {
-        H5Tset_cset(element_type.getId(), dtype_cset);
+    if (H5Tget_class(element_type.getId()) == H5T_STRING &&
+        H5Tget_cset(dtype.getId()) == H5T_CSET_ASCII) {
+        H5Tset_cset(element_type.getId(), H5T_CSET_ASCII);
     }
     return element_type;
 }};
@@ -54,9 +54,8 @@ struct string_type_checker<char[FixedLen]> {
 inline static DataType getDataType(const DataType& element_type, const DataType& dtype) {
     DataType return_type = (dtype.isFixedLenStr()) ? AtomicType<char[FixedLen]>() : element_type;
     // TEMP. CHANGE: See string_type_checker<void> definition
-    const auto dtype_cset = H5Tget_cset(dtype.getId());
-    if (dtype_cset == H5T_CSET_ASCII) {
-        H5Tset_cset(return_type.getId(), dtype_cset);
+    if (H5Tget_cset(dtype.getId()) == H5T_CSET_ASCII) {
+        H5Tset_cset(return_type.getId(), H5T_CSET_ASCII);
     }
     return return_type;
 }};
@@ -68,9 +67,8 @@ inline static DataType getDataType(const DataType&, const DataType& dtype) {
         throw DataSetException("Can't output variable-length to fixed-length strings");
     // TEMP. CHANGE: See string_type_checker<void> definition
     DataType return_type = AtomicType<std::string>();
-    const auto dtype_cset = H5Tget_cset(dtype.getId());
-    if (dtype_cset == H5T_CSET_ASCII) {
-        H5Tset_cset(return_type.getId(), dtype_cset);
+    if (H5Tget_cset(dtype.getId()) == H5T_CSET_ASCII) {
+        H5Tset_cset(return_type.getId(), H5T_CSET_ASCII);
     }
     return return_type;
 }};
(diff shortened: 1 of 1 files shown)


mgeplf:
> Dependency versions: hdf5 1.10.7

Is this the version it's tested against, or are all versions tested? Or is this the (only?) version that exhibits `string converter issues when memory and dataset types do not match`?

sergiorg-hpc:
> > Dependency versions: hdf5 1.10.7
> 
> Is this the version it's tested against, or are all versions tested? Or is this the (only?) version that exhibits `string converter issues when memory and dataset types do not match`?

We tested it with the latest release version v1.12.0, and the issue reproduces there as well. Nico and I checked the source code and it is a bit messy in that part, with patches over the years and some legacy code from the 90s. We are bringing this matter to the next HDF5 meeting (Tuesday of next week) and I'm going to formally report the issue, with a test file and source code example.

sergiorg-hpc:
@mgeplf I just tested it with the latest commit available in the official repository, and the issue still reproduces.

sergiorg-hpc:
We attended the weekly Call the Doctor meetings from the HDF Group, and it has been confirmed that what we found is a bug on HDF5:

https://github.com/HDFGroup/hdf5/issues/544

Hence, for now, this patch could be the solution to avoid the issue until the official patch is released.

include/highfive/bits/H5ReadWrite_misc.hpp
```diff
@@ -32,27 +33,36 @@ struct BufferInfo {
...
 };
 
 template <>
 struct string_type_checker<void> {
-inline static DataType getDataType(const DataType& element_type, bool) {
-    return element_type;
+inline static DataType getDataType(const DataType& element_type, const DataType& dtype) {
+    // TEMP. CHANGE: Ensure that the character set is properly configured to prevent converter
+    // issues on HDF5 <=v1.12.0
+    if (H5Tget_class(element_type.getId()) == H5T_STRING) {
```
sergiorg-hpc, line 44: Just a comment, should we check whether the encoding is ASCII or UTF-8, and throw an exception otherwise? I wonder if someone could create a file with another encoding, and break other parts of the code with this change (i.e., we are no longer forcing UTF-8 as the memory type of the buffer).

matz-e: I would at a minimum link to your issue here, to better keep track of it.

sergiorg-hpc: I added the check for ASCII files, otherwise I wouldn't feel comfortable forcing a character set that is wider than 8-bits. It might or might not happen, but just in case this is a safer fix for now.

I also added the comments you suggested, @matz-e!

sergiorg-hpc:
I'll update the PR as soon as I can confirm that everything seems to work as usual.

sergiorg-hpc:
It seems that the current version is relatively simple and can hide the problem, without introducing additional issues (e.g., 16-bit encoded strings could be read with errors in the original proposal).

To evaluate the result, I used the following source code to test the fix:

```C++
    vector<std::string> buffer;
    HighFive::File file("/gpfs/bbp.cscs.ch/project/proj12/jenkins/cellular/circuit-1k/nodes.h5");
    for (int i = 1; i < argc; i++) {
        file.getDataSet(std::string(argv[i])).read(buffer);

        // Print the values read from the file
        printf(">> %s:\n", argv[i]);
        for (const auto &str : buffer)
            printf("'%s' ", str.c_str());
        printf("\n\n");
    }
```

By calling the executable with `./a.out /nodes/All/0/@library/region /nodes/All/0/@library/morphology` (i.e., reading first the ASCII-encoded `region` and then the UTF-8 encoded `morphology`), the results clearly differ:

**_Without_ the fix**
```text
HDF5-DIAG: Error detected in HDF5 (1.8.12) thread 0:
  #000: ../../src/H5Dio.c line 179 in H5Dread(): can't read data
    major: Dataset
    minor: Read failed
  #001: ../../src/H5Dio.c line 435 in H5D__read(): unable to set up type info
    major: Dataset
    minor: Unable to initialize object
  #002: ../../src/H5Dio.c line 936 in H5D__typeinfo_init(): unable to convert between src and dest datatype
    major: Dataset
    minor: Feature is unsupported
  #003: ../../src/H5T.c line 4532 in H5T_path_find(): no appropriate function for conversion path
    major: Datatype
    minor: Unable to initialize object
terminate called after throwing an instance of 'HighFive::DataSetException'
  what():  Error during HDF5 Read:  
Aborted
```

**_With_ the fix**
```text
>> /nodes/All/0/@library/region:
'L1' 'L2' 'L3' 'L23' 'L123' 'L456' 

>> /nodes/All/0/@library/morphology:
'sm090227a1-2_idC' 'dend-C280998A-P3_axon-sm110131a1-3_INT_idA' 'dend-rr110114C1_idA_axon-sm110131a1-3_INT_idA' 'dend-C280998A-P3_axon-C190898A-P3' ... ' 'tkb060509a2_ch3_mc_n_el_100x_1' 
```

pramodk:
@matz-e / @alkino / @ferdonline : could you take a look at this and we can finish this?

sergiorg-hpc:
> @matz-e / @alkino / @ferdonline : could you take a look at this and we can finish this?

That was my last change, my apologies. I just realized this morning that this could be simplified, so I pushed it :).

This is ready to be reviewed and merged, if no additional comments appear.

ferdonline (approved):
LGTM

PR #431, opened 2021-04-01
Add integration tests with Morphio

### LINKED ISSUE #426: Enhance CI to test additional flagship projects
To ensure APIs didn't change and most projects keep running our CI should run additional integration tests
E.g. MorphIO seems to fail as default constructors were disabled/made private.

ferdonline:
Add Integration tests with MorphIO

Fixes #426

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -30,8 +30,9 @@ jobs:
       run: sudo apt-get update && sudo apt-get install libboost-all-dev libhdf5-dev libeigen3-dev
 
     - name: Build
-      CMAKE_OPTIONS=(-DHIGHFIVE_EXAMPLES:BOOL=False -DHIGHFIVE_UNIT_TESTS:BOOL=False)
-      run: source $GITHUB_WORKSPACE/.github/build.sh
+      run: |
+        CMAKE_OPTIONS=(-DHIGHFIVE_EXAMPLES:BOOL=False -DHIGHFIVE_UNIT_TESTS:BOOL=False)
+        source $GITHUB_WORKSPACE/.github/build.sh
 
     - name: Test
       working-directory: ${{github.workspace}}/build
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,0 +1,41 @@
+name: HighFive_Integration_tests
+
+on:
+  push:
+    branches:
+      - ci_test
+      - release/**
+  pull_request:
+    branches:
+      - master
+      - release/**
+    paths-ignore:
+      - '**.md'
+      - '**.rst'
+      - 'doc/**'
+
+env:
+  BUILD_TYPE: RelWithDebInfo
+  INSTALL_DIR: ${{github.workspace}}/install
+
+jobs:
+
+  Linux_Compilers:
+    runs-on: ubuntu-18.04
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: "Install libraries"
+      run: sudo apt-get update && sudo apt-get install libboost-all-dev libhdf5-dev libeigen3-dev
+
+    - name: Build
+      run: source $GITHUB_WORKSPACE/.github/build.sh
+
+    - name: Test
+      working-directory: ${{github.workspace}}/build
+      run: |
+        make install
+        git clone https://github.com/BlueBrain/MorphIO.git --recursive
+        cd MorphIO
+        sh ci/cpp_test.sh "-DCMAKE_PREFIX_PATH=$INSTALL_DIR -DEXTERNAL_HIGHFIVE=True"
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -20,8 +20,8 @@ env:
 
 jobs:
 
-  Linux_Compilers:
-    runs-on: ubuntu-18.04
+  MorphIO:
+    runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.yml
```diff
@@ -0,0 +1,41 @@
...
+    runs-on: ubuntu-18.04
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: "Install libraries"
+      run: sudo apt-get update && sudo apt-get install libboost-all-dev libhdf5-dev libeigen3-dev
+
+    - name: Build
+      run: source $GITHUB_WORKSPACE/.github/build.sh
```
matz-e, line 33: This does not import anything from the child environment… so `sh` would seem a better fit.

matz-e (approved):
I'm nitpicking, but this is fine.

.github/workflows/integration.yml
ferdonline: for the other way around, we want to be free to use parent variables in the child environment, exported or not.

tdegeus:
Personally I think that it is up to MorphIO to check against HighFive and not the other way around. MorphIO's CI could be run with a cronjob, or potentially one could even use a hook.

ferdonline:
Objective here is to add an additional protection layer against merging breaking changes.
I have nothing against setting it up in MorphIO and use hooks, as long as runs on our PRs, but I find it more complex. Would you like to take the task?

tdegeus:
As an outsider I simply think that PRs of a library (HighFive) should be discussed with any user in mind. Breaking changes should only be made in major releases, and where possible a deprecation warning should be included for at least one release cycle. 

Now, having a CI run on every PR of HighFive to check if this or that library that uses HighFive might be broken is in my opinion too aggressive. Instead, in think that dependent libraries should simply CI against HighFive, giving them ample opportunity to adapt if need (hopefully as little as possible). And if they decide to delay the changes needed they can simply using HighFive's latest release (instead of the main branch). I don't think HighFive's development should be delayed because of some downstream library. In my opinion the latter is a bit the suggestion that is made by instating this check.

But again, just an outsider. Also both projects are BlueBrain. So up to you, just giving my perpective.

ferdonline:
> Now, having a CI run on every PR of HighFive to check if this or that library that uses HighFive might be broken is in my opinion too aggressive

Technically it's not aggressive, it's a 2min test (contrary to launch their whole CI). Conceptually it could be if we opened the way to do it with random "this or that" projects. But we carefully selected a project we manage and care about in this projects own testing interests.

tdegeus:
> > Now, having a CI run on every PR of HighFive to check if this or that library that uses HighFive might be broken is in my opinion too aggressive
> 
> Technically it's not aggressive, it's a 2min test (contrary to launch their whole CI). Conceptually it could be if we opened the way to do it with random "this or that" projects. But in our own testing interests, we selected a project we manage and care about.

Sure, I understand well that within one organisation one wants to take great care of key projects! So you should definitely go ahead, regardless of what I wrote ;)

That being said two clarifications:

* What I meant by aggressive, is that one (or at least I) tends to hesitate to merge if CI is failing, even if it is for a good reason. But it is true that as long as one takes into consideration that there could be reasons to merge even with failing CI I guess that all is fine. Indeed, one should also care about our planet, and try to reduce (long) CI runs to only functional. And in that respect you are probably right for checking only on PRs here.
* For those of use not dealing with MorphIO, possible failing PR runs here could be a bit hard to judge. But this point is mute, since both projects are BlueBrain projects. 

But of course, this is a bit of a useless discussion, because good PRs should never lead to failing MorphIO runs (if there are indeed no unforeseen bugs on that side, which is also quite unlikely for a library as HighFive).

pramodk:
Even though I am from BlueBrain and very much care about our core libraries, I think @tdegeus has raised some good questions. Here are some of my additional questions/comments without endorsing any specific viewpoint :) :
* If highfive master is introducing some backward incompatible change, which is fine in master branch, then it will fail CI. How to handle such circular dependency?
* If morphio CI starts failing, it would be confusing for external contributors to understand the semantics of morphio tests. e.g.  If it's highfive test then often it would be straightforward use of APIs. If it's morphio test then there will be some domain specific knowledge in the test that wouldn't be straightforward to understand? (even though it's testing highfive usage)
* This implicitly suggest that highfive contributors are supposed to download and test against morphio project. Should that be the case?
* As highfive has well defined API, isn't it "easy" to cover the morph-io use cases by additional unit tests in the highfive itself? (I know that we do have such integration CIs but I wonder if this is really necessary for library like highfive)

ferdonline:
Thanks Pramod for raising these points. Those are some of the reasons for integration tests to run async of changes, and even internally we have seen those problems in other CI's. 

Keeping as is would mean: (1) nobody can break API, not even in master. I don't see a big problem there, we be forced to deprecating first and give some time for flag project to adapt. And thus would push for faster release cycle.
 (2) If third party project starts to fail we will fail. This IMO is the critical point. My argument here is that, being internal to BBP we could react fast, but I agree it could take days.

Option 1: If you think it's it's worth I'll investigate a good way of doing it, probably via a third party project (integration tests) which run overnight if there were changes.
Option 2: We make this test to run on-demand / commits to release
Option 3: Live with The restrictions mentioned before.

Apart from that, we anyway should make more tests to cover all public API.

ferdonline:
PR #442 and the new `HighFive-testing` repo build a nive Integration test env and therefore supersede this PR

PR #434, opened 2021-04-06
Bring back default ctors with Deprecation notice

ferdonline:
**Description**

Default constructors are not recommended since it opens more
room for errors, and thus they had been removed in PR 417
However, before completely remove them, let them be well marked
deprecated with a nice compilation notice

**How to test this?**

tests added to `tests_high_five_base`. Use ctest

### DIFF
--- a/include/highfive/H5DataSet.hpp
+++ b/include/highfive/H5DataSet.hpp
@@ -90,8 +90,8 @@ class DataSet : public Object,
         return getSpace().getElementCount();
     }
 
-    // No empty datasets
-    DataSet() = delete;
+    H5_DEPRECATED("Default ctor creates unsafe uninitialized objects")
+    DataSet() = default;
 
   protected:
     using Object::Object;  // bring DataSet(hid_t)
--- a/include/highfive/H5DataSet.hpp
+++ b/include/highfive/H5DataSet.hpp
@@ -90,7 +90,7 @@ class DataSet : public Object,
         return getSpace().getElementCount();
     }
 
-    H5_DEPRECATED("Default ctor creates unsafe uninitialized objects")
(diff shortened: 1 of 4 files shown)


include/highfive/H5DataSet.hpp
```diff
@@ -90,8 +90,8 @@ class DataSet : public Object,
         return getSpace().getElementCount();
     }
 
-    // No empty datasets
-    DataSet() = delete;
+    H5_DEPRECATED("Default ctor creates unsafe uninitialized objects")
```
tdegeus, line 93:
```suggestion
    H5_DEPRECATED("Default constructor creates unsafe uninitialized objects")
```

include/highfive/H5Group.hpp
```diff
@@ -26,8 +26,8 @@ class Group : public Object,
   public:
     const static ObjectType type = ObjectType::Group;
 
-    // No empty groups
-    Group() = delete;
+    H5_DEPRECATED("Default ctor creates unsafe uninitialized objects")
```
tdegeus, line 29:
```suggestion
    H5_DEPRECATED("Default constructor creates unsafe uninitialized objects")
```

tdegeus (commented):
~~Honestly I'm not entirely sure why this is needed? Is it what I've seen passing to you want to get some more CI working?~~

Sorry, overlooked the reason. OK, then it seems fine

include/highfive/H5DataSet.hpp
ferdonline: Yep, relatively to the latest release we had removed API without deprecating, which is not nice (my bad). We only noticed in some internal projects. The added tests would fail without the patch.

tdegeus (approved):

Issue #435, opened 2021-04-09
SZIP support


JaWSnl:
While GZIP is nice, SZIP is performing better for my use case. 

Note however that the SZIP that's provided with HDF5 is (or at the very least used to be) commercially licensed for anything but the most basic use case. See https://support.hdfgroup.org/doc_resource/SZIP/Commercial_szip.html for more details.

However `libaec` provides an SZIP compatible library and that makes SZIP a very useful addition in my opinion. See https://gitlab.dkrz.de/k202009/libaec for more details.

A pull request will be provided shortly.

PR #436, opened 2021-04-09
Add SZIP support (#435)

JaWSnl:
HDF has default support for SZIP, but commercially licensed.
Using `libaec` SZIP can be used with a permissive license.
This commit allows SZIP to be defined as a property.

The good news is I added a test (which is mostly copy-paste of the shuffle-deflate test). The bad news is I was unable to run the test successfully on my machine (not sure why that is, however compiling succeeds and I've also tested it by using it in my own code). I advise you to run the test yourself. You might want to install https://www.dkrz.de/redmine/projects/aec/wiki/Downloads instead of using the built-in szip for licensing reasons, and there may be some cmake configuration needed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
       shell: bash -l {0}  # required to activate conda
       run: |
         sudo apt-get -qq update
-        sudo apt-get -qq install libboost-all-dev libhdf5-openmpi-dev libeigen3-dev
+        sudo apt-get -qq install libboost-all-dev libhdf5-openmpi-dev libeigen3-dev libsz2
         conda install -y -c conda-forge xtl xsimd xtensor
 
     - name: Build
@@ -67,7 +67,7 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: "Install libraries"
-      run: sudo apt-get update && sudo apt-get install libboost-all-dev libhdf5-dev libeigen3-dev
+      run: sudo apt-get update && sudo apt-get install libboost-all-dev libhdf5-dev libeigen3-dev libsz2
 
     - name: Build
       env: ${{matrix.env}}
--- a/include/highfive/H5PropertyList.hpp
+++ b/include/highfive/H5PropertyList.hpp
@@ -148,6 +148,21 @@ class Deflate {
     const unsigned _level;
 };
 
+class Szip {
+  public:
+    explicit Szip(unsigned options_mask = H5_SZIP_EC_OPTION_MASK, 
+                  unsigned pixels_per_block = H5_SZIP_MAX_PIXELS_PER_BLOCK)
+        : _options_mask(options_mask)
+        , _pixels_per_block(pixels_per_block)
+    {}
+
+    private:
+      friend DataSetCreateProps;
+      void apply(hid_t hid) const;
+      const unsigned _options_mask;
+      const unsigned _pixels_per_block;
+};
+
 class Shuffle {
   public:
     Shuffle() = default;
--- a/include/highfive/bits/H5PropertyList_misc.hpp
+++ b/include/highfive/bits/H5PropertyList_misc.hpp
@@ -107,6 +107,18 @@ inline void Deflate::apply(const hid_t hid) const {
     }
 }
 
+inline void Szip::apply(const hid_t hid) const {
+    if (!H5Zfilter_avail(H5Z_FILTER_SZIP)) {
+        HDF5ErrMapper::ToException<PropertyException>(
+            "Error setting szip property");
+    }
+
+    if (H5Pset_szip(hid, _options_mask, _pixels_per_block) < 0) {
+        HDF5ErrMapper::ToException<PropertyException>(
+            "Error setting szip property");
+    }
+}
+
 inline void Shuffle::apply(const hid_t hid) const {
     if (!H5Zfilter_avail(H5Z_FILTER_SHUFFLE)) {
         HDF5ErrMapper::ToException<PropertyException>(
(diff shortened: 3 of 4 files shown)


matz-e (commented):
> The bad news is I was unable to run the test successfully on my machine (not sure why that is, however compiling succeeds and I've also tested it by using it in my own code). I advise you to run the test yourself. 

Do you link your own code against the SZIP lib? We normally use Spack to build our software, and I see that HDF5 takes a path to an SZIP installation at build time: https://github.com/BlueBrain/spack/blob/develop/var/spack/repos/builtin/packages/hdf5/package.py#L258

JaWSnl:
I indeed link my own code against szip, in the form of libaec (also see my description in #436), which I compiled locally, and then more or less "simply replace the szip.lib that comes with hdf". I could elaborate if it helps. libaec is small so it could be an option to compile it in high five's build process.

I'm not familiar with spack, nor with travis-ci, chocolatey, conda, mamba. So I was very happy to see @alkino taking it upon himself to attempt to fix things in that regard.

The reason that tests are not working locally, I could investigate that a little longer. I'm working on windows by the way.

JaWSnl:
I've spent a short amount of time on building HighFive on Linux today, which worked pretty much straight away, and I was able to run the unit tests successfully.

Trying to get szip to work has taken the rest of my day, and has so far been unsuccessful. Building HDF5 was easy (a lot easier compared to windows, as always...). But getting HighFive to use the `cmake --install`-ed HDF5 version, has been much harder. After reading and searching in all the wrong places these two lines did the trick of using the right HDF5:
```
set(HDF5_ROOT "/usr/local/HDF_Group/HDF5/1.13.0")
set(HDF5_DIR  "/usr/local/HDF_Group/HDF5/1.13.0/share/cmake/hdf5")
```

Next step: szip. Either from libaec or from libsz2.

I actually didn't plan on spending this much time on it, but the journey has been educational so far, so no harm done. More to follow soon, hopefully.

JaWSnl:
Apparently I apparently haven't configured it correctly yet. I get the following cmake configure output when using my own version from HDF5:
```
-- Performing Test COMPILER_SUPPORTS_CXX14 - Success
/usr/local/HDF_Group/HDF5/1.13.0/h5cc
dir is /usr/local/HDF_Group/HDF5/1.13.0
/usr/bin/ld: cannot find -lhdf5
collect2: error: ld returned 1 exit status
-- HDF5 C compiler wrapper is unable to compile a minimal HDF5 program.
-- Found HDF5: /usr/local/HDF_Group/HDF5/1.13.0/lib/libhdf5_debug.so (found version "1.13.0")  
```
It looks for h5cc in the HDF5_ROOT, instead of in HDF5_ROOT/bin. Investigating...

JaWSnl:
Apparently, h5cc is actually a shell script, and it does this "..":
```
dir=$(
  cd -P -- "$(dirname -- "$prg")/.." && pwd -P
) || exit
```
Odd. However, the fact that this "wrapper is unable to compile a minimal HDF5 program" doesn't seem to be an issue I have to deal with now.

matz-e:
Thank you for looking into the compilation issues!

I've just played around locally on our cluster with this…

> I'm not familiar with spack, nor with travis-ci, chocolatey, conda, mamba

We use Spack a lot, as it allows us to build customized software with large dependency trees pretty fast on Linux clusters (at least in my experience). It's basically a fancy wrapper around `configure`, `cmake`, or any other build system du jour.

Anyways, I did the following:

* Install szip, and attempt to directly link in the `libszip.so` into the unit tests with our default HDF5 (which is not linked against SZIP itself). This failed badly with a bunch of errors.
* Reinstall HDF5 linking against the official SZIP. This still failed, but with a different error regarding data types. Removing `complex` from the tested types helped. I think it should probably be sufficient to test against `dataset_test_types`, and I'll push that. Unit tests pass for me locally now, without any additional modifications to CMake.

This leaves the question of how to test this in the CI / Ubuntu. The package database has only the sz shim based on your library available, but I did not see any HDF5 plugin itself.

JaWSnl:
Note that I haven't given up on this yet, but it may be a few days before I get around to it. If there's anything you can tell me that will help in getting the Windows build working and addresses this
> This leaves the question of how to test this in the CI / Ubuntu. The package database has only the sz shim based on your library available, but I did not see any HDF5 plugin itself.
then please let me know. From that sentence, I'm not sure what the package database is (Spack perhaps?) and could use some pointers on what steps to pursue.

matz-e:
> Note that I haven't given up on this yet, but it may be a few days before I get around to it. If there's anything you can tell me that will help in getting the Windows build working and addresses this

Sorry, I don't have any experience with building things in Windows.

> > This leaves the question of how to test this in the CI / Ubuntu. The package database has only the sz shim based on your library available, but I did not see any HDF5 plugin itself.
> 
> then please let me know. From that sentence, I'm not sure what the package database is (Spack perhaps?) and could use some pointers on what steps to pursue.

Apologies, I was a bit involved with other projects. With the package database, I meant the `apt`/`deb` repositories of Ubuntu. In our CI, we use the following:
* `sudo apt-get update && sudo apt-get install libboost-all-dev libhdf5-dev libeigen3-dev` on Linii
* `brew install boost hdf5 eigen` on Mac
* `conda install -y -c conda-forge boost-cpp hdf5 eigen xtl xsimd xtensor` on Windows
If HDF5 ships without SZIP codepaths by default, we may have to either build HDF5 somewhere in the CI or add this codepath untested (in which case there should be a test for an SZIP feature in HDF5 and conditionally enable/disable the code)

matz-e:
Alright, did not notice that most of the tests actually pass :facepalm:.

I've added a bit where the tests will expect an exception if SZIP is not present (currently only Windows)… IMO this should be fine, support is currently tested on more platforms. It would be nice to have more solid testing, but this will do it for me.

ferdonline (approved):
LGTM

matz-e:
Thank you @JaWSnl!

JaWSnl:
And thank you, @matz-e

PR #465, opened 2021-05-26
Remove old cmake files

alkino:
* All those options are handled correctly by CMake directly


CMakeLists.txt
```diff
@@ -45,21 +43,15 @@ mark_as_advanced(HIGHFIVE_USE_INSTALL_DEPS)
 # Check compiler cxx_std requirements
 # -----------------------------------
 
-if(NOT COMPILER_SUPPORTS_CXX11)
-  message(FATAL_ERROR "HighFive version >= 2.0 requires c++ standard >= c++11")
```
ferdonline, line 49: I think this is a very useful check/info for users. Imagine the user uses this in another project with `CMAKE_CXX_STANDARD 98`. It will try to compile and give errors out of the blue.

ferdonline (changes requested):

CMakeLists.txt
alkino: hm, I will look how others projects are doing it, but right now I was hopping the next few lines will handle it.

CMake/config/ReleaseDebugAutoFlags.cmake
```diff
@@ -31,19 +31,16 @@ endif()
...
+set(CMAKE_C_FLAGS_DEBUG  "${CMAKE_C_WARNING_ALL} ${CMAKE_C_WARNING_DEBUG}")
+set(CMAKE_C_FLAGS_RELWITHDEBINFO "${CMAKE_C_WARNING_ALL} ${CMAKE_C_WARNING_DEBUG}")
 
 
 
-set(CMAKE_CXX_FLAGS_RELEASE  "${CMAKE_CXX_WARNING_ALL} ${CMAKE_CXX_OPT_NORMAL}")
-set(CMAKE_CXX_FLAGS_DEBUG  "${CMAKE_CXX_DEBUGINFO_FLAGS} ${CMAKE_CXX_WARNING_ALL} ${CMAKE_CXX_WARNING_DEBUG} ${CMAKE_CXX_OPT_NONE} ${CMAKE_CXX_STACK_PROTECTION}")
-set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_DEBUGINFO_FLAGS} ${CMAKE_CXX_WARNING_ALL} ${CMAKE_C_WARNING_DEBUG} ${CMAKE_CXX_OPT_NORMAL}")
-set(CMAKE_CXX_FLAGS_FAST "${CMAKE_CXX_WARNING_ALL} ${CMAKE_CXX_OPT_FASTEST} ${CMAKE_CXX_LINK_TIME_OPT} ${CMAKE_CXX_GEN_NATIVE}")
+set(CMAKE_CXX_FLAGS_RELEASE  "${CMAKE_CXX_WARNING_ALL}")
```
ferdonline, line 44: `CMAKE_CXX_FLAGS_RELEASE` already come with some good defaults right? Maybe worth to just "add" our extra flags, but not overwrite the defaults

CMakeLists.txt
ferdonline: While I don't think there was anything bad with the previous code I can agree with simplifying IF we provide similar functionality. I believe the only counter case was exactly the one I mentioned (`CMAKE_CXX_STANDARD being set to 98`). Should we 
```
IF (CMAKE_CXX_STANDARD EQUAL 98)
    message(FATAL_ERROR...
```

ferdonline:
@alkino what do you think? For me we can get rid of the code if at least we do that check mentioned before, i.e.
```
IF (CMAKE_CXX_STANDARD EQUAL 98)
    message(FATAL_ERROR...
```

ferdonline (approved):

PR #467, opened 2021-05-27
Adding compound types example

### LINKED ISSUE #454: Support for CompoundType in Attributes?
(TLDR: See bottom paragraph)

I've been unable to find an example or unit test that puts compound types in attributes. Given the following example code:
```
HighFive::CompoundType create_compound_Size2D() {
    return {{ "width",  HighFive::AtomicType<double>{} },
            { "height", HighFive::AtomicType<double>{} }};
}
HIGHFIVE_REGISTER_TYPE(Size2D, create_compound_Size2D)

void WriteAttribute(HighFive::DataSet& h5DataSet, const Size2D& value)
{
    // What goes here?
}
```
What goes in the `WriteAttribute` function?

I've tried for example:
```
auto size2DType = create_compound_Size2D();
size2DType.commit(h5DataSet, "size2d_type");
h5DataSet.createAttribute(
    "pixelSize",
    HighFive::DataSpace(HighFive::DataSpace::datascape_scalar), size2DType )
            .write(value);
```
Which gives:
> error C2338: supported datatype should be an arithmetic value, a std::string or a container/array
As provided by `H5Converter_misc.hpp`.  

The same compiler error is given when I try:
```
create_compound_Size2D();
h5DataSet.createAttribute("pixelSize", value)
```
Which is how I can successfully use EnumTypes that I've registered in the same way.

So now I'm wondering: does HighFive currently support using CompoundType in an attribute? If not, can you provide me some pointers on how to implement it?

ferdonline:
**Description**

This example shows all the steps for using compund datatypes. Compound attribute writing is also tested here.

**Testing**

Examples are automatically picked up by our CI.
Also checked locally for validity of output
```
Size2D                   Type
    Location:  1:800
    Links:     1
    Type:      shared-1:800struct {
                   "width"            +0    native double
                   "height"           +8    native double
               } 16 bytes
dims                     Dataset {2/2}
    Location:  1:1280
    Links:     1
    Storage:   32 logical bytes, 32 allocated bytes, 100.00% utilization
    Type:      struct {
                   "width"            +0    native double
                   "height"           +8    native double
               } 16 bytes
group1                   Group
    Attribute: dims {2}
        Type:      struct {
                   "width"            +0    native double
                   "height"           +8    native double
               } 16 bytes
    Location:  1:1552
    Links:     1
```

Closes #454

### DIFF
--- a/src/examples/compound_types.cpp
+++ b/src/examples/compound_types.cpp
@@ -1,0 +1,44 @@
+/*
+ *  Copyright (c), 2021, Blue Brain Project, EPFL
+ *
+ *  Distributed under the Boost Software License, Version 1.0.
+ *    (See accompanying file LICENSE_1_0.txt or copy at
+ *          http://www.boost.org/LICENSE_1_0.txt)
+ *
+ */
+
+// Compound datatype test :: May 2021
+// //////////////////////////////////
+
+#include <highfive/H5File.hpp>
+
+
+typedef struct {
+    double width;
+    double height;
+} Size2D;
+
+
+HighFive::CompoundType create_compound_Size2D() {
+    return {{ "width",  HighFive::AtomicType<double>{} },
+            { "height", HighFive::AtomicType<double>{} }};
+}
+
+HIGHFIVE_REGISTER_TYPE(Size2D, create_compound_Size2D)
+
+int main() {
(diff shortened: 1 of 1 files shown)


ohm314 (approved):
LGTM

PR #468, opened 2021-06-01
Resolve an issue where padding of nested compound types were being calculated incorrectly (#461)

### LINKED ISSUE #461: Offset of nested compound types is incorrect
When nesting compound types, the offset is calculated incorrectly. 

**To Reproduce**
1. Check out https://github.com/JaWSnl/HighFive/commit/7b5ddc7daacc168bf5ea588d7fa24fc9aefb6b51
2. Run the test
3. The equality test fails because the size will be 24 bytes.
4. When opening the .h5-file in HDFView, the information is messed up (however, retrieving the data works fine because the writing and reading follow the same incorrect behavior). See screenshot below.

**Expected behavior**
The test succeeds because the size of the datatype is 16 bytes. The data looks right in HDFView.

**Desktop (please complete the following information):**
 - OS: Windows
 - Version: master

**Additional context**
After some good old-fashioned console log debugging, I believe the problem is caused in [H5DataType_misc.hpp:L255](https://github.com/JaWSnl/HighFive/blob/master/include/highfive/bits/H5DataType_misc.hpp#L255), as well as the `_H5_STRUCT_PADDING` declaration. The member_size is calculated using `H5Tget_size` but for a compound type containing another compound type, the size retrieved here is the size of the entire nested compound type (in the test example above, 12 bytes). 

Instead, the proper padding for a compound type should be the first natural member of the struct and not the entire struct. At least, that's my suspicion as I can't seem to find whether HDF5 has other rules on padding (e.g. at https://support.hdfgroup.org/HDF5/doc1.6/UG/11_Datatypes.html). However, implementing it in "raw" HDF5 seems to work just fine (see example below).

A solution may be to check whether the member is a compound type and if so, recursively find the first atomic type (e.g. int). However before I take the effort of implementing, I wanted to check with you whether my assumptions are correct and this is indeed a proper solution.

An example of how this would be implemented in raw HDF5:
```
    hid_t file = H5Fcreate ("compound-test-hdf5.h5", H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);

    hsize_t     dims[1] = {2};

    CSL2 input[2];
    input[0] = { 1, {1,1,1} };
    input[1] = { 2, {3,4,5} };

    hid_t csl1_memtype = H5Tcreate(H5T_COMPOUND, sizeof(CSL1));
    H5Tinsert (csl1_memtype, "m1", HOFFSET (CSL1, m1), H5T_NATIVE_INT);
    H5Tinsert (csl1_memtype, "m2", HOFFSET (CSL1, m2), H5T_NATIVE_INT);
    H5Tinsert (csl1_memtype, "m3", HOFFSET (CSL1, m3), H5T_NATIVE_INT);

    hid_t memtype = H5Tcreate (H5T_COMPOUND, sizeof (CSL2));
    H5Tinsert (memtype, "t1",   HOFFSET (CSL2, t1),   H5T_NATIVE_INT);
    H5Tinsert (memtype, "csl1", HOFFSET (CSL2, csl1), csl1_memtype);

    hid_t space = H5Screate_simple (1, dims, NULL);
    hid_t dset = H5Dcreate (file, "dataset1", memtype, space, H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT);
    H5Dwrite (dset, memtype, H5S_ALL, H5S_ALL, H5P_DEFAULT, input);

    H5Dclose (dset);
    H5Sclose (space);
    H5Fclose (file);
```
How the data looks in HDFView (HighFive on the left, HDF5 on the right):
Screenshot 2021-05-21 093653

On another note, I'm still uncertain whether stored compound types have to be created with the `H5T_NATIVE*` or with the lower level `H5T_STD_*` types for proper interchangeable usage on different platforms.

JaWSnl:
**Description**
See #461 for an extensive description of the problem. Note that given my limited experience with HDF5 and H5 the solution might not be optimal or even sufficient. Your expertise is required to verify that.

Fixes #461

**How to test this?**
A unit test was created

**Test System**
 - OS: Windows

### DIFF
--- a/include/highfive/bits/H5DataType_misc.hpp
+++ b/include/highfive/bits/H5DataType_misc.hpp
@@ -242,33 +242,61 @@ inline AtomicType<Reference>::AtomicType() {
     _hid = H5Tcopy(H5T_STD_REF_OBJ);
 }
 
+inline size_t find_first_atomic_member_size(hid_t hid)
+{
+    // Recursive exit condition
+    if (H5Tget_class(hid) != H5T_COMPOUND) {
+        return H5Tget_size(hid);
+    }
+
+    auto number_of_members = H5Tget_nmembers(hid);
+    if (number_of_members == -1) {
+        throw DataTypeException("Cannot get members of CompoundType with hid: " +
+                                std::to_string(hid));
+    }
+    if (number_of_members == 0) {
+        throw DataTypeException("No members defined for CompoundType with hid: " +
+                                std::to_string(hid));
+    }
+
+    auto member_type = H5Tget_member_type(hid, 0);
+    auto size = find_first_atomic_member_size(member_type);
+    H5Tclose(member_type);
+    return size;
+}
 
 // Calculate the padding required to align an element of a struct
 #define _H5_STRUCT_PADDING(current_size, member_size) (((member_size) - (current_size)) % (member_size))
 
 inline void CompoundType::create(size_t size) {
     if (size == 0) {
-        size_t current_size = 0, max_type_size = 0;
+        size_t current_size = 0, max_atomic_size = 0;
 
         // Do a first pass to find the total size of the compound datatype
         for (auto& member: members) {
             size_t member_size = H5Tget_size(member.base_type.getId());
+
             if (member_size == 0) {
                 throw DataTypeException("Cannot get size of DataType with hid: " +
                                         std::to_string(member.base_type.getId()));
             }
 
+            size_t first_atomic_size = find_first_atomic_member_size(member.base_type.getId());
+
             // Set the offset of this member within the struct according to the
-            // standard alignment rules
-            member.offset = current_size + _H5_STRUCT_PADDING(current_size, member_size);
+            // standard alignment rules. The c++ standard specifies that:
+            // > objects have an alignment requirement of which their size is a multiple
+            member.offset = current_size + _H5_STRUCT_PADDING(current_size, first_atomic_size);
 
             // Set the current size to the end of the new member
             current_size = member.offset + member_size;
 
-            max_type_size = std::max(max_type_size, member_size);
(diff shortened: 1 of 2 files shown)


include/highfive/bits/H5DataType_misc.hpp
```diff
@@ -242,33 +242,59 @@ inline AtomicType<Reference>::AtomicType() {
...
+        throw DataTypeException("Cannot get members of CompoundType with hid: " +
+                                std::to_string(hid));
+    }
+    if (number_of_members == 0) {
+        throw DataTypeException("No members defined for CompoundType with hid: " +
+                                std::to_string(hid));
+    }
+
+    auto member_type = H5Tget_member_type(hid, 0);
+    return find_first_atomic_member_size(member_type);
```
JaWSnl, line 263: Given the call to H5Tget_member_type one line higher, I think the return line should be replaced with the following:
```
auto size = find_first_atomic_member_size(member_type);
H5Tclose(member_type);
return size;
```

jorblancoa: I agree, this should be changed.

jorblancoa (changes requested):
LGTM!
The only thing is the small change you suggested yourself regarding closing the member type.

include/highfive/bits/H5DataType_misc.hpp
JaWSnl: The change has been force-pushed (so that history is a single clean commit. Is that desired, or should it have been a new commit?).

JaWSnl:
> 
> LGTM!
> The only thing is the small change you suggested yourself regarding closing the member type.

Done!

jorblancoa (approved):

PR #469, opened 2021-06-15
Compound from hid

maxnoe:
**Description**

This adds a Constructor to `DataType` and `CompoundType`, to be able to infer and inspect `CompoundType` from files.

**How to test this?**

Added a section to the unit tests below. Also ran the following program on the input file created with the python script below:

```cpp
#include <cstdint>
#include <iostream>
#include <string>
#include <vector>


int main() {
    HighFive::File file("compound.h5", HighFive::File::ReadOnly);

    HighFive::DataSet dataset = file.getDataSet("test");
    HighFive::DataType dtype = dataset.getDataType();
    HighFive::CompoundType type(dtype.getId());

    for (const auto& member: type.getMembers()) {
        std::cout << member.name << ' ' << member.base_type.string() << ' ' << member.offset << '\n';
    }
    std::cout << '\n';
    return 0;
}
```

output:
```
N Integer32 0
x Float64 4
hit Integer8 12
```

Python script used to create file:
```
import h5py
import numpy as np

data = np.array(
    [(1, 4.5, True), (10, 3.1415, False)],
    dtype=[('N', np.uint32), ('x', np.float64), ('hit', np.uint8)]
)



with h5py.File('compound.h5', 'w') as f:
    f.create_dataset('test', data=data)
```

**Test System**
 - OS: Manjaro
 - Compiler: gcc 11.1
 - Dependency versions: hdf5 1.12.0


alkino:
Thanks a lot, really interesting patch!

I will accept it for now, but wait for someone else to say its thoughts

alkino (dismissed):

maxnoe:
@alkino Could you greenlight the CI system so the tests can run?

include/highfive/H5DataType.hpp
```diff
@@ -142,6 +144,23 @@ class CompoundType : public DataType {
         create(size);
     }
 
+    inline CompoundType(const hid_t hid) {
+        _hid = hid;
```
maxnoe, line 148: I should probably add a check here if the `hid` is really pointing to a compound type.

ferdonline: Thanks. Very nice indeed! If you are feeling like doing it nicer and introduce compile-time checks, I'd suggest to create a wrapper type around hid_t, so `dtype.getId()` returns, e.g. `TypeId`, being the only type accepted by `CompoundType` constructor (since hid can be anything, inc datasets)
The check for a compound type, probably needs to be runtime.

maxnoe: That would be quite the breaking change, wouldn't it? Changing the return type of `getId`? It would also break compile time polymorphism for `H5Object`, right? 

That's one point I haven't yet fully understood, how HighFive is implementing this polymorphism.

include/highfive/H5DataType.hpp
```diff
@@ -83,6 +83,8 @@ class DataType : public Object {
     /// \brief Returns whether the type is a Reference
     bool isReference() const;
 
+    DataType() = default;
+    DataType(hid_t hid) : Object(hid) {};
```
ferdonline, line 87: I find this unsafe and not required for your use case.

include/highfive/H5DataType.hpp
ferdonline: I also don't like breaking changes, we would need to inherit from the currently returned type. Something we can do on a different PR.
Anyway what seems a better idea is to directly accept a `DataType` object instead of a generic hid.

ferdonline (changes requested):

include/highfive/H5DataType.hpp
alkino: done

maxnoe:
@alkino Thanks for taking care of the review comments.

alkino:
Yes I will discuss that with other developers

alkino:
In fact I fixed the problem, it does not work here, but I need to understand why.

ferdonline (approved):

Issue #455, opened 2021-05-07
More useful warning message


tdegeus:
I'm getting the following warning:

```
HighFive WARNING: data and hdf5 dataset have different types: Float64 -> Integer64
```

which is great! But it would be much more practical if the name of the dataset would be included in the message

PR #471, opened 2021-06-22
Adding path to type warning message

tdegeus:
Fixes https://github.com/BlueBrain/HighFive/issues/455

It would be good to test on this, but frankly I don't know how to.

### DIFF
--- a/include/highfive/bits/H5ReadWrite_misc.hpp
+++ b/include/highfive/bits/H5ReadWrite_misc.hpp
@@ -22,7 +22,8 @@ struct BufferInfo {
     using char_array_t = typename details::type_char_array<type_no_const>::type;
     static constexpr bool is_char_array = ! std::is_same<char_array_t, void>::value;
 
-    BufferInfo(const DataType& dtype);
+    template <class F>
+    BufferInfo(const DataType& dtype, F getName);
 
     // member data for info depending on the destination dataset type
     const bool is_fixed_len_string;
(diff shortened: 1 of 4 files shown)


include/highfive/bits/H5ReadWrite_misc.hpp
```diff
@@ -28,6 +28,7 @@ struct BufferInfo {
     const bool is_fixed_len_string;
     const size_t n_dimensions;
     const DataType data_type;
+    bool warning_thrown = false;
```
alkino, line 31: Why inside `BufferInfo`

tdegeus: It is there that the the warning is generated. I'm not sure what you have in mind?

matz-e: I was going to comment on this earlier… Maybe `BufferInfo::getDataType` could get a provenance `string` parameter that would appended to the warning?

This would save this member variable, and avoid having to do checks outside the call, even if passing a parameter is not that pretty.

tdegeus: (Because I don't think HighFive stores the name as member?)

matz-e: If overhead is a concern, I guess the name computing function can be passed in directly/via a lambda-callback?

tdegeus: With passing a lambda function I can very well live ! Please check the updated code

matz-e (approved):
LGTM

PR #481, opened 2021-10-13
Set the minimal support of HDF5 to 1.10

### LINKED ISSUE #368: hdf5-1.8 not supported?
Hi, thank you for this great library!

Using the default hdf5 centos7 package (which is version 1.8) I get an error using the `group.createAttribute` method. It works fine with hdf5 1.10 or 1.12.

In HighFive documentation it is not clear which hdf5 version is supported... but maybe you can just replace https://github.com/BlueBrain/HighFive/blob/40f7624b37e1386fe1ac1885beb474e9346311d6/CMake/HighFiveTargetDeps.cmake#L11 by `find_package(HDF5 1.10 REQUIRED)` to ask for *at least* the 1.10 version.

If you are interested in the `createAttribute` bug let me know I'll build a minimal example.

alkino:
Fix #368


matz-e (approved):

alkino:
Should wait to know if users use 1.8, and to look to the bug before trying to set minimal version to 1.10

PR #489, opened 2021-12-20
Add acknoledgments to README.md and include AUTHORS.txt

sergiorg-hpc:


AUTHORS.txt
```diff
@@ -0,0 +1,44 @@
...
+Marian Heil
+Mario Emmenlauer
+Mathieu Bernard
+Matthias Wolf
+Maximilian Nöthe
+Mightrider
+Mike DePalatis
+Mike Gevaert
+Nico Jahn
+Nicolas Cornu (maintainer)
```
sergiorg-hpc, line 27: I assume Nico is the maintainer, right?

jamesgking: Yes, I agree with that

AUTHORS.txt
```diff
@@ -0,0 +1,44 @@
...
+Omar Awile
+Pablo Toharia
+Philip Deegan
+Pramod Kumbhar
```
sergiorg-hpc, line 31: Jorge and Pramod have co-authored changes, but they are not listed as contributors in GitHub.

What happens with those contributors? Should we skip them from this list somehow?

jamesgking (approved):

AUTHORS.txt
jamesgking: Add them to the list. We use a '(contributor)' tag if a person has been involved in discussions, but in this case I assume they can be considered authors

sergiorg-hpc:
**Side note:** Should we also update each and every source code file to include the copyright header of the Blue Brain Project?

AUTHORS.txt
sergiorg-hpc: That is the rule I understood as well. But I wonder if we are missing many other authors...

pramodk: I don't remember contributing anything specific to this project. Co-author might be just because I might have committed some PR? I wouldn't mind just ignoring.

pramodk (approved):
LGTM

PR #500, opened 2022-01-24
Add more file and group property list modifiers 

matz-e:
Added property list modifiers for file creation:

* `FileVersionBounds`
* `MetadataBlockSize`

also added functions to retrieve these settings from the `File` objects.

Added property list modifiers for groups:

* `EstimatedLinkInfo`

also added a new function for group creation to be able to use this modifier, and to retrieve settings. This requires creating a file that has a lower version bound for HDF5 1.8.

### DIFF
--- a/CMake/HighFiveTargetDeps.cmake
+++ b/CMake/HighFiveTargetDeps.cmake
@@ -18,7 +18,7 @@ if(NOT TARGET libdeps)
   endif()
 
   target_include_directories(libdeps SYSTEM INTERFACE ${HDF5_INCLUDE_DIRS})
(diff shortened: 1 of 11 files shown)


matz-e:
> Can we also add wrappers for reading the version bounds?

Sure, I suppose. Although beyond verification of stuff, I don't see a use case for them so far.

```
std::pair<…> H5File::GetVersionBounds()
```
sound good?

matz-e:
> I guess this feature should be implement inside https://github.com/BlueBrain/HighFive/blob/master/include/highfive/bits/H5PropertyList_misc.hpp But I'm not sure ;)

I stuck it in with the file creation stuff, no? The MPI thing works sort of the same… Maybe I should move it over, though

Issue #472, opened 2021-06-25
Documentation all-round check / improve


ferdonline:
We deploy docs to https://bluebrain.github.io/HighFive/
However the documentation is mostly about the API doc strings, whose visual is not too appealing.

To make HighFive more accessible it would be nice to have a very good land page, potentially based on ReadTheDocs, featuring

- How to get / install / integrate in user projects
- Design principles
- Explanation of each main entity
- Walk-though examples (lots of)
- Advanced stuff? Properties / Parallel IO / Structures?

tdegeus:
May I in this context also suggest using a nicer style for the docs. Personally I have been using https://github.com/jothepro/doxygen-awesome-css

1uc:
The README and doxygen style aren't the biggest issues anymore. Closing as too broad and non-actionable.

PR #507, opened 2022-03-04
[docs] Relative filenames. Using doxygen-awesome 

1uc:
Improves the Doxygen generated documentation with the following features.

- [x] Adds the doxygen-awesome style sheet.
- [x] Makes filenames relative, e.g. in file dependency graphs.

Improves #472


doc/doxygen-awesome-css/download_stylesheet.sh
```diff
@@ -0,0 +1,10 @@
+#! /usr/bin/env bash
```
tdegeus, line 1: Why is this file here? Wouldn't it make sense to include all needed files here (that is at least what I do elsewhere)?

tdegeus (commented):
Awesome! I was tempted to do this already when I looked at the output in your other issue.

doc/doxygen-awesome-css/download_stylesheet.sh
1uc: All files needed are included in the repository. This file can be removed.

The hope is that this file documents which files need to be downloaded to update the styling. The script provides the answer, just one, and it also provides the URL where that one file can be found (hopefully).

Would you like me to remove it?

tdegeus: My suggestion would be to have a GH Action check if doxygen-awesome is outdated. It could run for example on all merges to the main branch. 
Ideally it could even open a PR if it founds things that are outdated, but I don't know if that is possible.

tdegeus: Just the first hit when searching : https://github.com/istein/gh-actions-open-pr/blob/main/.github/workflows/pr.yml 
It seems like we can modify this to do what we want

1uc: Nice idea, I'll improve the PR to incorporate this.

tdegeus: Amazing. Let me know if you need help!

1uc:
I've tested the workflow in a separate repository quite thoroughly. Is there a way of testing it in this repo without effecting others adversely?

.github/workflows/check_doxygen_awesome_version.yml
```diff
@@ -0,0 +1,46 @@
+on:
+  workflow_dispatch:
+
+jobs:
+  check-for-updates:
+    runs-on: ubuntu-20.04
```
tdegeus, line 6:
```suggestion
    runs-on: ubuntu-latest
```

tdegeus (changes requested):
Other than the minor comment to me it looks fine. 
No, I don't know if we can pre-test it. I think we will just have to try. I will try it elsewhere too, just to verify.

.github/workflows/check_doxygen_awesome_version.yml
```diff
@@ -0,0 +1,46 @@
+on:
```
tdegeus, line 1:
```suggestion
name: Auto-update doxygen-awesome

on:
```

tdegeus:
I check too, works like a charm! https://github.com/tdegeus/GooseFEM

tdegeus (approved):

1uc:
It worked :) I was not expecting it to trigger this soon. Very nice idea of yours. I'll look into turning it into an GH action. Thanks for all the feedback during review.

tdegeus:
Like a charm! Thanks again, this is a really nice feature. It would be really nice to generalise this!

PR #509, opened 2022-03-04
Remove cyclic dependencies surrounding `H5Easy.hpp`.

### LINKED ISSUE #508: Remove cyclic dependencies of includes.
When looking at the dependency graph of the `H5Easy.hpp` one sees several cyclic dependencies and it's not very obvious what needs what. By lifting the declaration in the namespace `H5Easy` into a separate file, the dependencies can be simplified and made acyclic.

This change might affect how template specializations are imported. Therefore, I like to ask if such a change would be welcome and worthwhile. Sketch of a solution to follow.

1uc:
Lift the declaration in the namespace `H5Easy` to the file `h5easy_bits/H5Easy_decl.hpp`. Therefore, other headers can include this instead of `H5Easy.hpp`, thereby breaking the cyclic dependency.

- [x] cylic dependencies removed.
- [x] improve isolation of optional dependencies.

Fixes #508

**How to test this?**
Compile and run the unit-tests. If there's interest in merging this change one would need to

- [ ] check all supported compilers.
- [ ] decide if including `h5easy_bits/H5Easy_xtensor.hpp` directly must also pull the specializations for the other optional dependencies.


1uc:
Let's close this for now, because it's not a pain point and if need be it can be reopened and revisited.

PR #512, opened 2022-03-12
[docs] Checking update doxygen-awesome on pushes on master

tdegeus:


.github/workflows/check_doxygen_awesome_version.yml
```diff
@@ -2,6 +2,9 @@ name: Auto-update doxygen-awesome
 
 on:
   workflow_dispatch:
+  push:
+    branches:
+    - master
```
1uc, line 7: I'd like to suggest triggering the workflow nightly, e.g. like this:

    schedule: 
      - cron: 0 2 * * *

Advantages: 
  - fires at a constant rate, independent of how active development is at that point.
  - only triggers once a day. If the the CI breaks, we have time to react.
  - wont be a condition for other PRs to succeed.

tdegeus: Good idea. But I think that checking one a week it sufficient.

tdegeus:
Another detail: should the action fail if the PR already exists, see https://github.com/tdegeus/GooseFEM/runs/5521406363?check_suite_focus=true

tdegeus:
@1uc maybe you are the right person to review this, such that we can merge?

tdegeus:
@1uc maybe you are the right person to review this, such that we can merge?

.github/workflows/check_doxygen_awesome_version.yml
```diff
@@ -2,6 +2,8 @@ name: Auto-update doxygen-awesome
 
 on:
   workflow_dispatch:
+  schedule:
+  - cron: 0 2 * * 1
```
1uc, line 6: This is good, crontab.guru says:

> At 02:00 on Monday.

.github/workflows/check_doxygen_awesome_version.yml
```diff
@@ -21,7 +23,7 @@ jobs:
           # NOTE: In a later runs of CI we will search for PR with this exact
           #       title. Only if no such PR exists will the script create a
           #       new PR.
-          PR_TITLE="Update doxygen-awesome to ${VERSION}"
+          PR_TITLE="[docs] Update doxygen-awesome to ${VERSION}"
```
1uc, line 26: This doesn't seem to cause issues in the test repo. Hence, it should be fine. It will also not generate a new PR for the current version since there's no difference it doesn't try to create a PR.

1uc (approved):
I've checked the issue you mention, and can't reproduce it. The workflow only runs once a week, so it shouldn't trigger, simply because the stylesheet is up-to-date before this workflow runs again.

1uc:
@tdegeus please merge if you agree w.r.t. to the issue of failing when run twice.

tdegeus:
Thanks. Concerning the 'issue' I had on GooseFEM: let's investigate further if it occurs again. Note though, it's a detail: the action in that case is not supposed to do anything, so the exit code is not of curcial importance.

PR #514, opened 2022-03-16
Refactor and move file related property list stuff

matz-e:
Follow-up to #500, moves the property list manipulation relating to
files into H5PropertyList.hpp, while also removing the template for MPI
file opening and moving to preprocessor conditionals.

IMO this is a bit cleaner and feels more natural when using the
modifier.


ohm314 (approved):
This looks good to me

PR #519, opened 2022-03-22
Fix "namespace definition is not allowed here"

ChrisDeGrendele:
I was getting the error "namespace definition is not allowed here" here when calling HIGHFIVE_REGISTER_TYPE. This fixes that error.

- [x ] Issue 1 fixed
- [ ] Issue 2 fixed
- [ ] Feature 1 added
- [ ] Feature 2 added

Fixes #(issue)

**How to test this?**

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce if there is no integration test added with this PR. Please also list any relevant details for your test configuration

```bash
cmake ..
make -j8
make test
```

**Test System**
 - OS: [e.g. Ubuntu 20.04]

### DIFF
--- a/include/highfive/H5DataType.hpp
+++ b/include/highfive/H5DataType.hpp
@@ -386,7 +386,7 @@ class FixedLenStringArray {
 /// \endcode
 #define HIGHFIVE_REGISTER_TYPE(type, function) \
     template<>                                 \
-    HighFive::DataType HighFive::create_datatype<type>() {         \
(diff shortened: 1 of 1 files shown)


matz-e (dismissed):
LGTM, thank you for the fix!

alkino (approved):

alkino:
An example of reproducing it will be good.
I think this kind of macro should be called outside of every context. Otherwise it is a bad practice.
let fix this however.

alkino (approved):

PR #527, opened 2022-03-28
Auto-load Catch2

tdegeus:
@matz-e @alkino


tdegeus:
Actually I don't think this works yet

CMakeLists.txt
```diff
@@ -85,11 +84,12 @@ if (NOT CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
 endif()
 
 if(HIGHFIVE_UNIT_TESTS)
-  if(HIGHFIVE_USE_EXTERNAL_CATCH2)
-    find_package(Catch2 REQUIRED)
-  elseif(EXISTS ${CMAKE_CURRENT_LIST_DIR}/deps/catch2/CMakeLists.txt)
+  find_package(Catch2)
+  if(EXISTS ${CMAKE_CURRENT_LIST_DIR}/deps/catch2/CMakeLists.txt)
     add_subdirectory(deps/catch2 EXCLUDE_FROM_ALL)
     list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/deps/catch2/contrib)
```
matz-e, line 90: One issue I have here is that I'm concerned about CMake mixing things up.

For example, we find Catch2, setting a bunch of stuff. Then, if the submodule is present, we include it and _append_ to the search path. Which means that `include(Catch)` would probably first look at the system Catch2, no?

It would seem cleaner to me to look for Catch2 only if the subdirectory is empty, printing a status message about the missing submodule and then using `REQUIRE`, maybe?

matz-e: FWIW, `find_library` seems to be in the CMake 3.1 we require, what was wrong with it?

tdegeus: I find `add_subdirectory(deps/catch2 EXCLUDE_FROM_ALL)` very un-Cmake, and that is probably where the issues start. Personally I would find it much simpler (and much more in the spirit of Cmake ;)) to remove the submodule all-together. We did not submodule Boost either and that did not pose any difficulties. Nowadays I do not consider Catch2 much less standard than Boost.

alkino: catch2 is packaged under debian / ubuntu / archlinux / fedora at least

tdegeus: and on conda-forge, and I imagine other package managers

matz-e: Well… so just `find_package(Catch2 REQUIRED)`?

Having stuff as submodules is annoying, but seems to be accepted practice in C++ projects. No complaints from me if we should ditch it again.

tdegeus: Yes! That would be much cleaner and much easier to support!

.github/workflows/gh-pages.yml
```diff
@@ -20,15 +20,11 @@ jobs:
     - name: Basic GitHub action setup
       uses: actions/checkout@v2
 
-    - name: Set conda environment "test"
-      uses: conda-incubator/setup-miniconda@v2
+    - name: Set conda environment
+      uses: mamba-org/provision-with-micromamba@main
```
matz-e, line 24: Why not `mamba` everywhere?

CMakeLists.txt
pramodk: I don't have strong opinions but to mention few aspects:

> We did not submodule Boost either and that did not pose any difficulties. 

I would say Boost is different beast altogether. As we depend on library components, (as we all know) one can't just submodule boost...

> catch2 is packaged under debian / ubuntu / archlinux / fedora at least
> and on conda-forge, and I imagine other package managers

Using system package managers is perfectly fine when we are doing development on our own system / desktop environments. When we move to clusters (where many of us work), the same doesn't apply. `Boost` availability via `module` has been quite universal for long time. I wouldn't say the same thing for `Catch2`. 

>Personally I would find it much simpler (and much more in the spirit of Cmake ;)) to remove the submodule all-together.

People involved in this discussion knows how to make catch2 available if needed. But IMO, we don't loose much by adding submodule for lightweight package is like Catch2. As @matz-e, it's just convenient, commonly used approach and one doesn't need to bother about the details.

So I think it's ok to use external packages but shouldn't get rid of submodule altogether.

.github/workflows/gh-pages.yml
tdegeus: Will do, much faster!

.gitmodules
```diff
@@ -1,6 +1,3 @@
 [submodule "deps/hpc-coding-conventions"]
```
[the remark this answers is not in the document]
tdegeus: I don't know what it does. It'd be happy to remove it!

CMakeLists.txt
```diff
@@ -24,7 +24,6 @@ set(USE_OPENCV OFF CACHE BOOL "Enable OpenCV testing")
 mark_as_advanced(USE_BOOST USE_EIGEN USE_XTENSOR)
 
 option(HIGHFIVE_USE_BOOST "Enable Boost Support" ${USE_BOOST})
-option(HIGHFIVE_USE_EXTERNAL_CATCH2 "Enable external Catch2 for unit tests" OFF)
```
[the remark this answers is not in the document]
tdegeus: I'm fine with this

tdegeus:
Do you know what is going on @matz-e @alkino ?

CMakeLists.txt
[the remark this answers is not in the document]
alkino: @pramodk and @ohm314 are against deleting the submodule. We should keep the previous system...

CMakeLists.txt
tdegeus: There is not something profoundly bad about carrying a submodule. However, my considerations are:

1. Have a submodule makes as responsible for keeping it up-to-date. Not carrying the submodule reduces work-load in that sense. 

2. I wonder how many people will actually run the unit-tests for highfive. We are responsible for that, and indeed we run quite a bit of CI. So I wonder how many people benefit. 

My proposition is on of : 
* Having the submodule, but not require its clone by default (so having a fallback that takes the system's)
* Having an option to specify the path to the user's download of Catch2 (instead of using the system's).

.gitmodules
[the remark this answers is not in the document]
matz-e: It provides a bunch of formatting hooks that are quite heavy and involved. We should rather set something up that is centered around `git-clang-format` (or similar)

matz-e:
> Do you know what is going on @matz-e @alkino ?

Mamba does not install the right HDF5. Pretty much don't like this - from the looks of the CI file, we rely on Ubuntu system packages, why do we mix and match package managements?
```
-- Found HDF5: /home/runner/micromamba/envs/myenv/lib/libhdf5.so (found version "1.12.1")  
CMake Warning at CMake/HighFiveTargetDeps.cmake:17 (message):
  Parallel HDF5 requested but libhdf5 doesnt support it
Call Stack (most recent call first):
  CMakeLists.txt:60 (include)
```

CMakeLists.txt
matz-e: > Having the submodule, but not require its clone by default (so having a fallback that takes the system's)

This can be done.

> Having an option to specify the path to the user's download of Catch2 (instead of using the system's).

Don't understand this. You can always install Catch2 independently and use `-DCMAKE_PREFIX_PATH=…`

.gitmodules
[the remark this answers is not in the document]
alkino: Done ;)

PR #531, opened 2022-03-31
Add an example for Eigen.

1uc:
This PR adds one example of writing an `Eigen::MatrixXd` to disk using `File::createDataset` directly.

It also adds to the build-system to exclude Eigen examples if H5 is being built without Eigen.

### DIFF
--- a/src/examples/eigen_matrix.cpp
+++ b/src/examples/eigen_matrix.cpp
@@ -37,7 +37,7 @@ int main(void) {
         // we create a new hdf5 file
         File file(FILE_NAME, File::ReadWrite | File::Create | File::Truncate);
 
-        // let's create our dataset of the size of the boost::multi_array
+        // let's create our dataset of the size of the eigen matrix
         file.createDataSet(DATASET_NAME, matrix);
 
     } catch (Exception& err) {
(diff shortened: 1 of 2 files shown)


src/examples/eigen_matrix.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include <highfive/H5File.hpp>
+
+using namespace HighFive;
+
+const std::string FILE_NAME("eigen_matrix_example.h5");
+const std::string DATASET_NAME("dset");
+const int nrows = 10;
+const int ncols = 3;
+
+// Create a 2D dataset 10x3 of double with boost multi array
```
alkino, line 23:
```suggestion
// Create a 2D dataset 10x3 of double with eigen matrix
```

src/examples/eigen_matrix.cpp
```diff
@@ -0,0 +1,51 @@
...
+        for (int i = 0; i < nrows; ++i) {
+            for (int j = 0; j < ncols; ++j) {
+                matrix(i, j) = double(j + i * 100);
+            }
+        }
+
+        // we create a new hdf5 file
+        File file(FILE_NAME, File::ReadWrite | File::Create | File::Truncate);
+
+        // let's create our dataset of the size of the boost::multi_array
```
alkino, line 40:
```suggestion
        // let's create our dataset of the size of the eigen matrix
```

1uc:
Correct, therefore, the copyright, also needs adjusting.

alkino (approved):

PR #537, opened 2022-04-01
Use `REQUIRE` in loops.

1uc:
In the unit test `CHECK` is used throughout. Even in loops to check that all the elements in two arrays are equal. When several tests fail, this can generate enough lines to overflow the scroll back buffer, e.g. for VS Code.

When checking that all elements of an array are equal it's suffices to print one failing number and then stop the test.

This PR switches to `REQUIRE` for all assertions happening in a loop. Assertions outside of loops remain `CHECK` since it can be convenient to see several failed assertions for a single test.


matz-e (approved):
Thank you for catching this.

PR #539, opened 2022-04-01
Clang format

alkino:


.clang-format
```diff
@@ -78,7 +78,7 @@ PenaltyExcessCharacter: 1000000
 PenaltyReturnTypeOnItsOwnLine: 60
 PointerAlignment: Left
 ReflowComments:  true
-SortIncludes:    true
```
matz-e, line 81: Does this correspond to our coding conventions?

matz-e (commented):
I like it in general!

Could we add a little note in the README as to which version of `clang-format` we are using?

.clang-format
alkino: The problem is that headers of highfive need to be in a fixed order. Several peoples tried to fix it but unsucessfully iirc.

include/highfive/H5Easy.hpp
```diff
@@ -161,14 +158,14 @@ class DumpOptions
     inline void set(T arg, Args... args);
 
     ///
-    /// \brief Set chunk-size. If the input is rank (size) zero, automatic chunking is enabled.
-    /// \param shape Chunk size along each dimension.
+    /// \brief Set chunk-size. If the input is rank (size) zero, automatic chunking is
+    /// enabled. \param shape Chunk size along each dimension.
```
alkino, line 162: Looks wrong, right?

.clang-format
matz-e: Sorry, meant the whole file 😅

alkino: This file was already there...
Should i copy one from hpc-coding-convention?

tristan0x: Formatting config should the closest possible to hpc-coding-conventions. Of course, any rule can be overridden, like the `SortIncludes`

include/highfive/H5Easy.hpp
tristan0x: yeah, NL needed

tristan0x:
I would ignore this revision from `git blame`, see https://michaelheap.com/git-ignore-rev/
1. Create a `.git-blame-ignore-revs` file with this revision
2. Explain in the development guide how to use this file

.clang-format
matz-e: > This file was already there... Should i copy one from hpc-coding-convention?

Yes please. Or re-add the submodule and link to the corresponding file.

matz-e (approved):
Good from my side

PR #551, opened 2022-04-06
`select(const HyperSlab&)` reads into 1D arrays, only.

1uc:
Since the dimensions of the array into which the selection will be read should be clear without knowing if the hyperslab happens to be regular, we can only support reading general hyperslabs into 1D arrays.

We therefore remove unused functions which should restore HDF5 1.8 compatibility.

The lost functionality can be recoverd by providing means of checking if a hyperslab is regular, and returning the regular hyperslab if it is. One would then add a further overload `select(const RegularHyperSlab &)` which reads the regular hyperslab into a packed array, similar to how selecting columns works. However, both of these are not supported in HDF5 1.8.

### DIFF
--- a/include/highfive/bits/H5Slice_traits.hpp
+++ b/include/highfive/bits/H5Slice_traits.hpp
@@ -49,15 +49,6 @@ class ElementSet {
     friend class SliceTraits;
 };
 
-inline bool isRegularHyperslab(const DataSpace& dataspace) {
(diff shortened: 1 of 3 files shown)


1uc:
The library has been built successfully against HDF5 v1.8.10.

alkino (approved):

Issue #504, opened 2022-02-08
EnumType<> inside a CompoundType causes segfault


mdimura:
I created and registered a trivial EnumType and a Compound type with this EnumType. When I try to create the dataset - I get a segfault.

```c++
#include <highfive/H5File.hpp>
#include <vector>
using namespace HighFive;

enum class State {
	OPEN,
	CLOSED,
};
EnumType<State> create_enum_t()
{
	return {{"open",State::OPEN},
		{"closed",State::CLOSED}};
}
HIGHFIVE_REGISTER_TYPE(State, create_enum_t);

struct Record
{
	double pi=3.14;
	State st=State::OPEN; //
};
CompoundType rec_t()
{
    return {{"pi",AtomicType<double>{}},
            {"st",EnumType<State>{}}
    };
}
HIGHFIVE_REGISTER_TYPE(Record, rec_t);

int main()
{
	const size_t numRec = 2;
	File f("test.h5", File::Truncate);
	auto dataset = f.createDataSet<Record>("records", DataSpace(numRec));
	std::vector<Record> recs(numRec);
	dataset.write(recs);
	return 0;
}
```

**Stacktrace**
```
H5T_detect_class 0x00007fd98158c4ac
H5D__create 0x00007fd9813ffdb8
<unknown> 0x00007fd98140c6ee
H5O_obj_create 0x00007fd9814f5d2a
<unknown> 0x00007fd9814bb68c
<unknown> 0x00007fd98148efbf
H5G_traverse 0x00007fd98148f4f6
<unknown> 0x00007fd9814b720f
H5L_link_object 0x00007fd9814bce82
H5D__create_named 0x00007fd9813ffc3c
H5VL__native_dataset_create 0x00007fd981626255
<unknown> 0x00007fd98160e99c
H5VL_dataset_create 0x00007fd9816140dc
H5Dcreate2 0x00007fd9813de261
HighFive::NodeTraits<HighFive::File>::createDataSet H5Node_traits_misc.hpp:44 <=========
HighFive::NodeTraits<HighFive::File>::createDataSet<Record> H5Node_traits_misc.hpp:64
main main.cpp:34
__libc_start_main 0x00007fd980df8b25
_start 0x000055936123d52e

```
Variables (stack content) at `H5Node_traits_misc.hpp:44`:

```
Signal = SIGSEGV (Segmentation fault)
[...]

PR #552, opened 2022-04-06
Improve EnumType API

alkino:
* block creation of empty Enum (hdf5 segfault with them)
* fix examples to use create_datatype instead of AtomicType (more idiomatic)


alkino:
Should fix #504

matz-e (approved):

PR #556, opened 2022-04-07
Minor fixes to build the poster code

tristan0x:
Minor changes

---
Also, there is this declaration of a C array declaration with a non-static size 
https://github.com/BlueBrain/HighFive/blob/a0ed166c4221558acc51b9bc9e86be29d4cab835/doc/poster/example1_hdf5.cpp#L9-L10
Looks like a C99 extension.

### DIFF
--- a/include/highfive/H5PropertyList.hpp
+++ b/include/highfive/H5PropertyList.hpp
@@ -228,6 +228,7 @@ class Chunking {
 
   private:
     friend DataSetCreateProps;
+    friend GroupCreateProps;
     void apply(hid_t hid) const;
     const std::vector<hsize_t> _dims;
 };
@@ -239,6 +240,7 @@ class Deflate {
 
   private:
     friend DataSetCreateProps;
+    friend GroupCreateProps;
     void apply(hid_t hid) const;
     const unsigned _level;
 };
--- a/doc/poster/example1_highfive.cpp
+++ b/doc/poster/example1_highfive.cpp
@@ -1,6 +1,6 @@
 #include <highfive/H5File.hpp>
 
-using namespace HighFive;
+using HighFive::File;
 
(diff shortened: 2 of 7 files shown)


include/highfive/H5PropertyList.hpp
```diff
@@ -228,6 +228,7 @@ class Chunking {
 
   private:
     friend DataSetCreateProps;
+    friend GroupCreateProps;
```
matz-e, line 231: Why this?

https://support.hdfgroup.org/HDF5/doc/RM/RM_H5P.html#Property-SetChunk

Says:
> hid_t plist | IN: Dataset creation property list                identifier.

And doesn't seem to talk about groups.

include/highfive/H5PropertyList.hpp
```diff
@@ -239,6 +240,7 @@ class Deflate {
...
```
matz-e, line 243: This OTOH, seems to be correct.

include/highfive/H5PropertyList.hpp
tristan0x: So instruction L15 is invalid @matz-e ?
https://github.com/BlueBrain/HighFive/blob/b18aa78cc53818212e1f75332036f8fba9bc13bd/doc/poster/example2.cpp#L13-L16

In that case, it would be nice to have a trap that redirects the compiler to a `static_assert(cond, "This property doesn't apply to groups")`

matz-e: I think so, and I agree!

ohm314: Yeah that was my mistake initially, but fixed it later.

ohm314 (approved):
looks good.

PR #558, opened 2022-04-07
Updated coverage job in CI to run on linux

ohm314:
**Description**

Our CI code coverage job has been running on macos and is quite slow. Switching to linux, which gives us more available runners and should also complete faster.


matz-e (approved):

matz-e (dismissed):
From 14m to 6m, thanks! 🎉 🌮

alkino (dismissed):

alkino (approved):

PR #561, opened 2022-04-10
Fix computation of PADDING

alkino:
Computation of padding takes size_t types that are unsigned.
If member_size is smaller than current_size, the compute is bad.

### DIFF
--- a/include/highfive/bits/H5DataType_misc.hpp
+++ b/include/highfive/bits/H5DataType_misc.hpp
@@ -270,7 +270,7 @@ inline size_t find_first_atomic_member_size(hid_t hid) {
 
 // Calculate the padding required to align an element of a struct
 #define _H5_STRUCT_PADDING(current_size, member_size) \
-    (((member_size) - (current_size)) % (member_size))
+    ((std::max(member_size, current_size) - std::min(member_size, current_size)) % (member_size))
 
 inline void CompoundType::create(size_t size) {
     if (size == 0) {


matz-e (dismissed):

include/highfive/bits/H5DataType_misc.hpp
```diff
@@ -270,7 +270,7 @@ inline size_t find_first_atomic_member_size(hid_t hid) {
 
 // Calculate the padding required to align an element of a struct
 #define _H5_STRUCT_PADDING(current_size, member_size) \
-    (((member_size) - (current_size)) % (member_size))
+    ((std::max(member_size, current_size) - std::min(member_size, current_size)) % (member_size))
```
ohm314, line 273: Good catch - the reason why this is correct and the original code wasn't, is that `%` is division remainder. So if  `(member_size) - (current_size)` is negative, then the result of the macro might be negative, messing up your padding.
(I'm mostly explaining that for my own benefit 😆 )

ohm314 (approved):
looks good, thanks @alkino !

include/highfive/bits/H5DataType_misc.hpp
matz-e: These are all `size_t`, seems like a real danger of some wrap-around, too.

matz-e (approved):

include/highfive/bits/H5DataType_misc.hpp
ohm314: you're right - but the wraparound will still break the original code - so this is a good fix.

PR #562, opened 2022-04-11
Add benchmarking programs

ferdonline:
**Description**

The current PR adds benchmarking programs so that we can compare HighFive versus hand-written hdf5 versions.

Results made the basis for the HiveFive poster presented at PASC 2022

**This PR**

 - Adds the benchmarking programs under `src/benchmarking/`
 - Updates CI to build and run the highfive benchmark
 - Add to CI the tests to check integration of the exported HighFive Cmake target in other projects (left behind in travis)

**Notes**
It was noticeable the effect of doing `vector.insert()` vs `std::copy(... std::back_inserter())` (the latter being slower) and the sensitivity of doing things "per row", namely doing checks or creating hyperslab objects.


benchmarks/Makefile
```diff
@@ -0,0 +1,14 @@
+CXX=g++
+HDF5_ROOT=/usr/local
```
matz-e, line 2: Can't this use CMake, too?

matz-e (commented):
IMO, having this at the top level is sort of cluttering. Would be better to have it in, i.e., `src/benchmarks`, as we already have `src/examples`.

benchmarks/Makefile
matz-e: So move poster `.cxx` files into `examples`, and add the compilation bits?

ohm314: I'd replace the old ones, though and then of course there needs to be some update in the doc generation. If you think this is a good idea, I can take care of it.

ferdonline: Thanks for the feedback. I see the benefits of recycling the benchmark demos to examples. My only two arguments against are:
 - Makefiles make it dead-easy to see and tune compilation flags. I was playing with optimization and inlining  levels, etc, without thinking "maybe cmake is trying to do smth smart".
 - Having a separate folder with the run_benchmarks script would help anybody to reproduce the results we got and eventually proceed from there.

Having that in mind I kind of would prefer just moving the benchmarks to src, but no strong feelings.

src/benchmarks/Makefile
```diff
@@ -0,0 +1,15 @@
+CXX=g++
+HDF5_ROOT=/usr/local
```
matz-e, line 2: This is highly system dependent. Also all user-modifiable variables here should be set with:
>  There is another assignment operator for variables, `?='. This is called a conditional variable assignment operator, because it only has an effect if the variable is not yet defined. This statement: 
(see https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_chapter/make_6.html)

I think this would allow to grab these variables from the environment easier.

benchmarks/Makefile
tristan0x: Sorry to jump on this only now but have you considered using the Google benchmark library? It is very straightforward to use. On one side, it requires a new library to link against but on the other side, it would wrap the 2 executables and the bash script in one single executable, and would not require caliper.

ferdonline: That sounds very interesting, will definitely read about it. Nevertheless, I would like to close this topic, maybe that's a nice improvement for a free-friday?

matz-e (commented):
And this is "simpler" than CMake?

If there's a Makefile, it should at least go the ancient way of doing things:
```
❯ pkg-config --libs --cflags hdf5
-I/usr/include/hdf5/mpich -I/usr/include/x86_64-linux-gnu/mpich -L/usr/lib/x86_64-linux-gnu/hdf5/mpich -lhdf5 -Wl,-Bsymbolic-functions -Wl,-z,relro -lmpich -lm -lbacktrace -lpthread -lrt
```

matz-e:
> And this is "simpler" than CMake?
> 
> If there's a Makefile, it should at least go the ancient way of doing things:
> 
> ```
> ❯ pkg-config --libs --cflags hdf5
> -I/usr/include/hdf5/mpich -I/usr/include/x86_64-linux-gnu/mpich -L/usr/lib/x86_64-linux-gnu/hdf5/mpich -lhdf5 -Wl,-Bsymbolic-functions -Wl,-z,relro -lmpich -lm -lbacktrace -lpthread -lrt
> ```

See also: https://stackoverflow.com/questions/28533059/how-to-use-pkg-config-in-make

ferdonline:
> And this is "simpler" than CMake?

Never said it's easier, just mentioned it's good to have full control of flags. And in this case that purpose seem lost too with all those flags...
Maybe you were using the parallel version?

ferdonline:
@matz-e Any further comment here?

src/benchmarks/README.md
```diff
@@ -0,0 +1,17 @@
+# Benchmarking
+
+This folder contains several baseline programs used to benchmark HighFive and assess
+its overhead wrt no-highfive programs.
+
+It features a straightforward Makefile whose flags can be easily tuned. By default
+compilation features -O2 -g and looks up hdf5 in /usr/local.
```
matz-e, line 7:
```suggestion
compilation features -O2 -g and finds HDF5 via `pkg-config`.
```

src/benchmarks/README.md
```diff
@@ -0,0 +1,17 @@
...
+Additionally, a `run_benchmarks.sh` script is provided to measure execution time and
+profile using hpctoolkit.
+
+## Compile
+
+Basically, run `make`. By default it compiles with `-g -O2` but it's configurable. e.g.
+
+```
+make HDF5_ROOT=/opt/hdf5-1.10/ COMPILE_OPTS="-g -O1"
```
matz-e, line 16:
```suggestion
make CXX=clang++ COMPILE_OPTS="-g -O1"
```

src/benchmarks/hdf5_bench.cpp
```diff
@@ -0,0 +1,59 @@
+#include "hdf5.h"
+#include <iostream>
+#include <stdexcept>
+#include <vector>
+
+#define NROWS      1000000  // 1M
+#define ROW_LENGTH 10
+
+const std::vector<std::vector<int>> data(NROWS, {1, 2, 3, 4, 5, 6, 7, 8, 9, 10});
```
matz-e, line 9: OT, but Eigen / `std::vector<std::array<…>>` would be more appropriate.

matz-e (approved):

PR #571, opened 2022-04-15
Make support of std::complex<T> generic

alkino:


matz-e (dismissed):

include/highfive/bits/H5DataType_misc.hpp
```diff
@@ -169,31 +169,14 @@ class AtomicType<FixedLenStringArray<StrLen>>: public DataType {
...
-            // h5py/numpy compatible datatype
-            H5Tinsert(_hid, "r", 0, H5T_NATIVE_DOUBLE);
-            H5Tinsert(_hid, "i", sizeof(double), H5T_NATIVE_DOUBLE);
-        };
-    } complexType;
-    _hid = H5Tcopy(complexType.getId());
-}
+template <typename T>
+class AtomicType<std::complex<T>>: public DataType {
+  public:
```
ferdonline, line 174: what about `static_assert(std::is_arithmetic<T>, )` ?

alkino: is_arithmetic is not enough, it should be is floating point.
But I think it is one more check for nothing because the user should not write std::complex of smth that is floating point. His program will be ill-formed.

ferdonline: It's a "good to have". Not a blocker

matz-e (approved):

matz-e (approved):

PR #586, opened 2022-05-27
Automagic serialization

alkino:
All types are supported


include/highfive/bits/H5Converter_misc.hpp
```diff
@@ -628,16 +636,31 @@ struct inspector<boost::numeric::ublas::matrix<T>> {
...
-class data_converter {
-    static Reader<T> get_readers(const std::vector<size_t> dims) {
-        size_t n_dim = inspector<T>::recursive_ndim;
-        if (n_dim < dims.size()) {
-            throw std::string("Error dataset is too larger for T");
-        } else if (n_dim > dims.size()) {
+struct data_converter {
+    static std::vector<size_t> try_reduce_dims(const std::vector<size_t> dims, size_t n) {
+        std::vector<size_t> dims_{};
+        for (auto it = dims.crbegin(); it != dims.crend() && dims_.size() != n; ++it) {
```
matz-e, line 642: With `dims_.size() != n` in the loop condition, it seems to me that the resulting `dims_` will always be `<= n`.

But below you then test `dims_.size() > n`, I don't see how that will ever be triggered?

alkino: If we try to fit {3, 3, 1} into 1D it will not works.
And the loop will stop after the last element.

matz-e: Do we have a test for that?

alkino: https://github.com/BlueBrain/HighFive/blob/master/tests/unit/tests_high_five_base.cpp#L1533-L1569

matz-e: Hm, tried to cover that condition in there, and got told off by https://github.com/BlueBrain/HighFive/blob/start_serialize/include/highfive/bits/H5Dataspace_misc.hpp#L145

I'm still wondering how to cover it, but I think that most of our use cases are kept in check, too.

matz-e: I'd still feel more comfortable with it like this
```suggestion
        for (auto it = dims.crbegin(); it != dims.crend(); ++it) {
```
but maybe I'm overthinking this in a wrong way

alkino: hm your patch looks wrong because we can go "too" far

include/highfive/bits/H5Attribute_misc.hpp
```diff
@@ -67,10 +63,25 @@ inline void Attribute::read(T& array) const {
...
+    }
+    return array;
+}
+
+template <typename T>
+inline void Attribute::read(T& array) const {
+    const DataSpace& mem_space = getMemSpace();
+    auto dims = mem_space.getDimensions();
+    details::inspector<T>::prepare(array, dims);
+    array = read<T>();
```
1uc, line 84: These lines look counter intuitive to me. Therefore, a few questions:
* If a user allocates an `array` of the correct shape and passes this to `Attribute::read(T& array)`. Will these lines cause a move? If yes, what happens if `T` only support copying?
* Can we restructure this in such a way that `T Attribute::read()` becomes syntactic sugar for `void Attribute::read(T&)`. Currently it's the other way around. In the simplest case it would be
```
T Attribute::read() const {
    T array;

    // Since the sizes aren't as needed, this
    // will trigger re-allocation of `array`.
    read(array);

    return array;
}

File: include/highfive/bits/H5Converter_misc.hpp [hunk 3]
```diff
@@ -1,419 +1,750 @@
...
+template <typename T>
+struct inspector<T*> {
+    using type = T*;
+    using value_type = T;
+    using base_type = typename inspector<value_type>::base_type;
+    using hdf5_type = typename inspector<value_type>::hdf5_type;
+
+    static constexpr size_t ndim = 1;
+    static constexpr size_t recursive_ndim = ndim + inspector<value_type>::recursive_ndim;
+    static constexpr bool is_trivially_copyable = true;
```
1uc, line 376: From the documentation I gather that this answers the question whether the inner type is trivially copyable? Don't you need the same thing as for `std::vector<T>`?

include/highfive/bits/H5Converter_misc.hpp
```diff
@@ -1,419 +1,750 @@
...
+    using type = unqualified_t<T>;
+    using base_type = unqualified_t<T>;
+    using hdf5_type = base_type;
 
-inline bool is_1D(const std::vector<size_t>& dims) {
-    return std::count_if(dims.begin(), dims.end(), [](size_t i) { return i > 1; }) < 2;
-}
+    static constexpr size_t ndim = 0;
+    static constexpr size_t recursive_ndim = ndim;
+    static constexpr bool is_trivially_copyable = true;
```
1uc, line 85: How do you know that `T` is trivially copyable?

alkino: it is the base inspector. I can try if T is trivially copyable, right.

include/highfive/bits/H5Converter_misc.hpp
alkino: We don't support recursivity with T*, I can try to add it

include/highfive/bits/H5Attribute_misc.hpp
alkino: done :)

include/highfive/bits/H5Converter_misc.hpp
```diff
@@ -594,6 +593,12 @@ struct inspector<boost::numeric::ublas::matrix<T>> {
     }
 
     static void prepare(type& val, const std::vector<size_t>& dims) {
+        if (dims.size() < 2) {
```
1uc, line 596: Hard coded expr:  `... < 2`.

1uc (approved):

PR #587, opened 2022-06-04
Support for half-precision (16-bit) floating-point datasets

ssbotelh:
**Description**

Although 16-bit float is not a native C++ type, it is widely used to save runtime and disk space in situations where high precision is not needed, most notably in deep-learning applications. Python's `h5py` trivially supports the creation of `float16` datasets (`type "<f2"`). This PR adds optional support for half-precision in HighFive.

**Code change summary**

- The main code change is in `include/highfive/bits/H5DataType_misc.hpp`, where the `AtomicType` constructor was specialized for `float16_t`
- The type list `numerical_test_types` was augmented by adding `float16_t` (if enabled)
- Added example code `create_dataset_half_float.cpp` in `src/examples`
- A few makefile changes

**How to test this?**

Download the (single-header) `Half` library from http://half.sourceforge.net/, and copy `half.hpp` into `/usr/local/include/`.
Then build and run unit tests:
 
```bash
cmake .. -DHIGHFIVE_USE_HALF_FLOAT=ON
make -j8
make test
```
You should see tests with `numerical_test_types - 13`, which corresponds to `float16_t`

**Test System**
 - OS: Ubuntu 18.04
 - Compiler: GNU g++ 10.3.0
 - Dependency versions: tested with HDF5 1.10.5, boost 1.65.1, and Half 2.2.0

### DIFF
--- a/include/highfive/bits/H5DataType_misc.hpp
+++ b/include/highfive/bits/H5DataType_misc.hpp
@@ -133,11 +133,13 @@ inline AtomicType<unsigned long long>::AtomicType() {
 using float16_t = half_float::half;
 
 template <>
-inline AtomicType<float16_t>::AtomicType()
-{
+inline AtomicType<float16_t>::AtomicType() {
     _hid = H5Tcopy(H5T_NATIVE_FLOAT);
+    // Sign position, exponent position, exponent size, mantissa position, mantissa size
     H5Tset_fields(_hid, 15, 10, 5, 0, 10);
+    // Total datatype size (in bytes)
     H5Tset_size(_hid, 2);
+    // Floating point exponent bias
     H5Tset_ebias(_hid, 15);
 }
 #endif
--- a/include/highfive/bits/H5DataType_misc.hpp
+++ b/include/highfive/bits/H5DataType_misc.hpp
@@ -16,6 +16,9 @@
 #include <H5Ppublic.h>
 #include <H5Tpublic.h>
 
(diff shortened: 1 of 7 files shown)


include/highfive/bits/H5DataType_misc.hpp
```diff
@@ -125,7 +128,20 @@ inline AtomicType<unsigned long long>::AtomicType() {
...
-// float, double and long double mapping
+// half-float, float, double and long double mapping
+#ifdef H5_USE_HALF_FLOAT
+using float16_t = half_float::half;
+
+template <>
+inline AtomicType<float16_t>::AtomicType()
+{
+    _hid = H5Tcopy(H5T_NATIVE_FLOAT);
+    H5Tset_fields(_hid, 15, 10, 5, 0, 10);
```
alkino, line 139: Can you add a comment here to be explicit?

alkino, line 139:
```suggestion
    sign position, exponent position, exponent size, mantissa position, mantissa size
    H5Tset_fields(_hid, 15, 10, 5, 0, 10);
```

ssbotelh: Added comments and fixed clang-format violations. Thanks!

matz-e:
> clang-format is version 12 here: https://github.com/BlueBrain/HighFive#code-formatting

#588 begs to differ.

alkino (approved):

PR #591, opened 2022-06-14
Add support of bool

alkino:
Support bool with vector and array, quite messy.


alkino:
it delete support of bool => hbool_t but seriously what is it and who use that?

alkino:
It is a nightmare because of `char[N]`...

PR #602, opened 2022-08-19
Sliced XTensor IO

### LINKED ISSUE #596: H5Easy does not support sliced load for xtensor type
Running this code returns a template resolution error.

```cpp
int main()
{
	const HighFive::File file = HighFive::File("example.h5", HighFive::File::ReadOnly);
	auto A = H5Easy::load<xt::xtensor<float, 2>>(file, "/example", {0, 1})
}
```

Error output:
```
H5Easy_public.hpp(157): error C3861: 'load_part': identifier not found
```

using vcpkg to install libs:
- xtensor: 0.24.0
- xtensor-io: 0.13.0#1
- highfive: 2.3#2

nathan-mxr:
**Description**

Implemented load_part and dump_extend for XTensor data types.

This requires a new public load API that accepts a vector of sizes as a parameter, to define how many elements should be read in each dimension.

Fixes #596 

**How to test this?**

Have added tests to tests_high_five_easy.cpp to exercise this feature:
- Loading a single value into an XTensor/XArray (no sizes argument)
- Loading a subset of a dataset into an XTensor/XArray (normal sizes argument)
- Load a subset that exceeds the bounds of a dataset into an XTensor/XArray (index + size > dataset dimension)
- Dumping an XTensor/XArray/XView to a non-zero index of a Dataset. 

```bash
cmake ..
make -j8
make test
```

**Test System**
 - OS: Ubuntu 22.04
 - Compiler: gcc 11.2.0
 - Dependency versions: hdf5 1.12

### DIFF
--- a/include/highfive/h5easy_bits/H5Easy_public.hpp
+++ b/include/highfive/h5easy_bits/H5Easy_public.hpp
@@ -136,6 +136,11 @@ inline DataSet dump(File& file,
 }
 
 template <class T>
+inline T load(const File& file, const std::string& path, const std::vector<size_t>& idx, const std::vector<size_t>& sizes) {
+    return detail::io_impl<T>::load_part(file, path, idx, sizes);
+}
+
+template <class T>
 inline T load(const File& file, const std::string& path, const std::vector<size_t>& idx) {
     return detail::io_impl<T>::load_part(file, path, idx);
 }
--- a/include/highfive/h5easy_bits/H5Easy_public.hpp
+++ b/include/highfive/h5easy_bits/H5Easy_public.hpp
@@ -136,7 +136,10 @@ inline DataSet dump(File& file,
 }
 
 template <class T>
-inline T load(const File& file, const std::string& path, const std::vector<size_t>& idx, const std::vector<size_t>& sizes) {
+inline T load(const File& file,
+              const std::string& path,
+              const std::vector<size_t>& idx,
+              const std::vector<size_t>& sizes) {
     return detail::io_impl<T>::load_part(file, path, idx, sizes);
 }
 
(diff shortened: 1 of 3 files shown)


tdegeus:
Thanks! Before diving into it, could you briefly discuss here what the API would like like? Along this lines should we consider something like the API for `xt::view` ?

nathan-mxr:
> Thanks! Before diving into it, could you briefly discuss here what the API would like like? Along this lines should we consider something like the API for `xt::view` ?

Hi Tom, the new H5_Easy API method for `load` can be found in the change to `include/highfive/h5easy_bits/H5Easy_public.hpp`. 

It is an extension of the previously existing prototype, which had arguments:
```
const File& file, 
const std::string& path, 
const std::vector<size_t>& idx
```
allowing only a single item of the dataset to be addressed.

The new API has arguments:
```
const File& file,  
const std::string& path, 
const std::vector<size_t>& idx,
const std::vector<size_t>& sizes
```
Where idx and sizes must have the same number of elements. idx indicates where the first element of the slice can be found, and sizes denotes how many elements in each dimension should be read from the dataset. This is similar to the H5 hyperslab selection arguments `offset` and `count`: https://bluebrain.github.io/HighFive/class_high_five_1_1_slice_traits.html#a6a76a20247d0e8f8629b81c17425d73c
Hyperslab selection also allows for "stride" and "block" vectors, that I have not implemented.

I think the implementation for `xt::view` may be slightly more complicated, as (if I understand correctly) `xt::view::data()` will return a pointer to the entire underlying data structure, so can't be trivially used with `write_raw()` and `read()`.

tdegeus (changes requested):
Sorry for my late review. In general I think this is very nice. Thanks a lot! What I really not like is the automatic resizing of the load slice (see above). For the rest, I had a worry, but I think that it is unjustified (see above). 

Then, I think that the test would be a lot cleaner without the automatic type-cast, which tests a feature of xtensor and not of highfive.

alkino:
As there is no update since 5 months, I close it. If someone want to fix / continue this PR, we can reopen it.

PR #618, opened 2022-10-14
Implement file space strategies.

1uc:
Implements support for the different file space strategies:
https://portal.hdfgroup.org/display/HDF5/H5P_SET_FILE_SPACE_STRATEGY
https://portal.hdfgroup.org/display/HDF5/H5P_SET_FILE_SPACE_PAGE_SIZE

including paged allocation with configurable page size. Since both are "file creation properties". This MR also includes a new ctor for `File` which enables creating files specifying both creation and access properties.

### DIFF
--- a/include/highfive/H5File.hpp
+++ b/include/highfive/H5File.hpp
@@ -85,6 +85,14 @@ class File: public Object, public NodeTraits<File>, public AnnotateTraits<File> 
     /// \brief Returns the HDF5 version compatibility bounds
     std::pair<H5F_libver_t, H5F_libver_t> getVersionBounds() const;
 
+#if H5_VERSION_GE(1, 10, 1)
+    /// \brief Returns the HDF5 file space strategy.
+    H5F_fspace_strategy_t getFileSpaceStrategy() const;
+
+    /// \brief Returns the page size, if paged allocation is used.
+    hsize_t getFileSpacePageSize() const;
+#endif
+
     ///
     /// \brief flush
     ///
@@ -94,6 +102,7 @@ class File: public Object, public NodeTraits<File>, public AnnotateTraits<File> 
 
   protected:
     hid_t getAccessPList() const;
+    hid_t getCreatePList() const;
 
   private:
     using Object::Object;
--- a/include/highfive/H5PropertyList.hpp
+++ b/include/highfive/H5PropertyList.hpp
@@ -193,7 +193,59 @@ class MetadataBlockSize {
     const hsize_t _size;
 };
 
+#if H5_VERSION_GE(1, 10, 1)
 ///
+/// \brief Configure the file space strategy.
+///
+/// See the upstream documentation of `H5Pget_file_space_strategy` for more details. Essentially,
+/// it enables configuring how space is allocate in the file.
+///
+class FileSpaceStrategy {
+  public:
+    ///
+    /// \brief Create a file space strategy property.
+    ///
+    /// \param strategy  The HDF5 free space strategy.
+    /// \param perist Should free space managers be persisted across file closing and reopening.
+    /// \param threshold The free-space manager wont track sections small than this threshold.
+    FileSpaceStrategy(H5F_fspace_strategy_t strategy, hbool_t persist, hsize_t threshold);
+
+  private:
+    friend FileCreateProps;
+
+    void apply(const hid_t list) const;
+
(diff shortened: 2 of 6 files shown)


include/highfive/H5File.hpp
```diff
@@ -55,6 +55,18 @@ class File: public Object, public NodeTraits<File>, public AnnotateTraits<File>
...
+    /// \param filename: filepath of the HDF5 file
+    /// \param openFlags: Open mode / flags ( ReadOnly, ReadWrite)
+    /// \param fileAccessProps: the file access properties
+    ///
+    /// Open or create a new HDF5 file
+    File(const std::string& filename,
+         unsigned openFlags,
+         const FileCreateProps& fileCreateProps,
+         const FileAccessProps& fileAccessProps = FileAccessProps::Default());
+
```
alkino, line 69: SHould you instead add a new Prop to the ctor above?

1uc: I'm worried about code such as:
```
auto fapl = FileAccessProps{};
File("foo.h5", File::ReadOnly, fapl);
```
if we simply change the signature to:
```
File(const std::string& filename,
     unsigned openFlags = ...,
     const FileCreateProps& fileCreateProps = FileCreateProps::Default(),
     const FileAccessProps& fileAccessProps = FileAccessProps::Default());
```
then that will break. OTOH, if we do:
```
File(const std::string& filename,
     unsigned openFlags = ...,
     const FileAccessProps& fileAccessProps = FileAccessProps::Default(),
     const FileCreateProps& fileCreateProps = FileCreateProps::Default());
```
then the order is different than the one for `H5Fcreate`.

1uc:
One thing to consider is if we want to start adding more and more methods to `File`, e.g. `File::getFileSpacePageSize()`. Instead one could pursue a strategy that allows reading/extracting values from property lists.

include/highfive/H5File.hpp
matz-e: `createDataSet` also has first create, then access properties. So this works for me the best in terms of consistency.

include/highfive/H5PropertyList.hpp
```diff
@@ -194,6 +194,70 @@ class MetadataBlockSize {
...
+///
+/// See the upstream documentation of `H5Pget_file_space_strategy` for more details. Essentially,
+/// it enables configuring how space is allocate in the file.
+///
+class FileSpaceStrategy {
+  public:
+    ///
+    /// \brief Create a file space strategy property.
+    ///
+    /// \param strategy  The HDF5 free space strategy.
```
matz-e, line 207:
```suggestion
    /// \param strategy The HDF5 free space strategy.
```
Whitespace, please.

include/highfive/bits/H5File_misc.hpp
```diff
@@ -98,14 +105,63 @@ inline hsize_t File::getMetadataBlockSize() const {
...
+    return strategy;
+}
+
+inline hsize_t File::getFileSpacePageSize() const {
+    auto fcpl = getCreatePList();
+    hsize_t page_size;
+
+    if (getFileSpaceStrategy() != H5F_FSPACE_STRATEGY_PAGE) {
+        HDF5ErrMapper::ToException<FileException>(
+            std::string("Cannot obtain page size, because not using paged allocation."));
```
matz-e, line 136:
```suggestion
            std::string("Cannot obtain page size as paged allocation is not used."));
```

matz-e (dismissed):

include/highfive/H5PropertyList.hpp
matz-e: Still there? Is this from formatting? The line does not align, either. I'm puzzled.

include/highfive/H5PropertyList.hpp
```diff
@@ -193,7 +193,59 @@ class MetadataBlockSize {
...
+    /// \param perist Should free space managers be persisted across file closing and reopening.
```
matz-e, line 209:
```suggestion
    /// \param persist Should free space managers be persisted across file closing and reopening.
```

include/highfive/bits/H5File_misc.hpp
matz-e: `because not using` is missing a subject and feels like a kludge.

matz-e (approved):

PR #622, opened 2022-10-27
Deprecate FileDriver(s)

matz-e:
The current FileDriver implementations just inherit from
FileAccessProps, and I think that with further additions to the file
access property lists, it is cleaner to just use FileAccessProps
directly.

### DIFF
--- a/src/examples/parallel_hdf5_write_dataset.cpp
+++ b/src/examples/parallel_hdf5_write_dataset.cpp
@@ -36,9 +36,11 @@ int main(int argc, char** argv) {
     using namespace HighFive;
     try {
         // open a new file with the MPI IO driver for parallel Read/Write
+        FileAccessProps faprops;
+        faprops.add(MPIOFileACCess{MPI_COMM_WORLD, MPI_INFO_NULL});
         File file(FILE_NAME,
                   File::ReadWrite | File::Create | File::Truncate,
-                  MPIOFileDriver(MPI_COMM_WORLD, MPI_INFO_NULL));
+                  faprops);
 
         // we define the size of our dataset to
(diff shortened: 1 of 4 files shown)


alkino (approved):
+1 this way to do is not straight forward following what is done for other properties!

roblatham00:
Could you also update `parallel_hdf5_write_dataset.cpp` to demonstrate the preferred method?  Sounds like with this change it would simply be
```
        // open a new file with the MPI IO driver for parallel Read/Write
          File file(FILE_NAME,
                    File::ReadWrite | File::Create | File::Truncate,
                    FileAccessProps(MPI_COMM_WORLD, MPI_INFO_NULL));

But there is no FileAccessProps constructor that takes a `comm` and `info`

**Comment by alkino** [comment]
@roblatham00: I updated the example so you can see.

File: src/examples/parallel_hdf5_write_dataset.cpp [hunk 1]
```diff
@@ -36,9 +36,11 @@ int main(int argc, char** argv) {
     using namespace HighFive;
     try {
         // open a new file with the MPI IO driver for parallel Read/Write
+        FileAccessProps faprops;
+        faprops.add(MPIOFileACCess{MPI_COMM_WORLD, MPI_INFO_NULL});
         File file(FILE_NAME,
                   File::ReadWrite | File::Create | File::Truncate,
-                  MPIOFileDriver(MPI_COMM_WORLD, MPI_INFO_NULL));
+                  faprops);
```
1uc: Maybe we can call them `access_props`? (or `fapl` like HDF5)

1uc (approved):

PR #621, opened 2022-10-26
Work in progress: collective data

roblatham00:
**Description**
Make it possible to request collective I/O for H5Dwrite and H5Dread.   Like HDF5 itself, users have to opt-in with an 'enable' call. 

I'm brand new to HighFive, and have developed a lot more C code than C++ code, so I would appreciate a review to make sure I'm doing this the HighFive way

fixes #112 

**How to test this?**

Also in this pull request are some updates to the parallel write test
- make the memory type match the file type
- request collective I/O
- 
 - Dependency versions: collective I/O has been in HDF5 "forever".  The "why wasn't this collective"? routine was introduced in hdf5-1.8.0

PR #623, opened 2022-10-27
Enable collective MPI IO.

1uc:
This PR adds the ability to pass a data transfer property list to `read` and `write` operations. This is needed to enable collective MPI-IO.

In essence this is #621

### DIFF
--- a/src/examples/parallel_hdf5_write_dataset.cpp
+++ b/src/examples/parallel_hdf5_write_dataset.cpp
@@ -49,14 +49,35 @@ int main(int argc, char** argv) {
 
         // Create the dataset
         DataSet dataset = file.createDataSet<double>(DATASET_NAME, DataSpace(dims));
-        dataset.enable_collective();
 
         // Each node want to write its own rank two time in
         // its associated row
-        double data[1][2] = {{mpi_rank*1.0, mpi_rank*1.0}};
+        double data[1][2] = {{mpi_rank * 1.0, mpi_rank * 2.0}};
+
+        auto xfer_props = DataTransferProps{};
+        xfer_props.add(UseCollectiveIO{});
 
         // write it to the associated mpi_rank
-        dataset.select({std::size_t(mpi_rank), 0}, {1, 2}).write(data);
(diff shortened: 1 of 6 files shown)


matz-e (dismissed):
Looks decent to me.

1uc:
@roblatham00 would this variation work for your applications?

src/examples/parallel_hdf5_write_dataset.cpp
```diff
@@ -52,10 +52,30 @@ int main(int argc, char** argv) {
...
+        auto err = H5Pget_mpio_no_collective_cause(xfer_props.getId(), &local_cause, &global_cause);
+        if (err < 0) {
+            throw std::runtime_error("Failed to check mpio_no_collective_cause.");
+        }
+        if (local_cause || global_cause) {
+            std::cout << "The operation wasn't collective: " << local_cause << " " << global_cause
+                      << std::endl;
+            throw std::runtime_error("IO wasn't collective.");
+        } else {
+            std::cout << "Success! The operation was collective.\n";
```
roblatham00, line 76: Thanks for adding an example of how to obtain the property list and query it.  In a parallel test, this output will clutter the screen so can we delete the "Success" and emit text only if an error?

roblatham00: I know this is an example, and clients can do whatever they want with highfive, but  I think throwing a runtime error is overkill for this condition.  The user requested collective i/o, HDF5 library was unable to perform I/o collectively (for whatever reason), but did perform I/O independently.

1uc: Good point. The intention behind throwing is that we run the examples a part of our CI. So what you've spotted is that I'm abusing the examples as a test that the request for collective I/O was forwarded correctly to HDF5.

It doesn't cover collective reads, nor all of the entry points. I'll add proper test coverage for this; which frees up the example to be purely instructional. Thank you!

matz-e (approved):

PR #624, opened 2022-10-28
Collective metadata

1uc:
This PR enables users to configure collective metadata reads and writes (at file level). The heavy lifting was done in #620; here we simply restructure the proposed solution.

In more detail, this PR adds:
* ability to request collective MPI-IO metadata reads
* ability to request collective MPI-IO metadata writes
* explains the caveats of collective metadata reads
* add collective MPI-IO metadata writes to `MPIOFileDriver`.


1uc:
I think it would be nice to have an example showing how to set everything up to use collectives, always. However this requires both this PR and #623. Therefore, we'll do this later.

include/highfive/bits/H5FileDriver_misc.hpp
```diff
@@ -15,6 +15,7 @@ namespace HighFive {
 #ifdef H5_HAVE_PARALLEL
 inline MPIOFileDriver::MPIOFileDriver(MPI_Comm comm, MPI_Info info) {
     add(MPIOFileAccess(comm, info));
+    add(MPIOCollectiveMetadataWrite(true));
```
matz-e, line 18: Looks good. Only question I have: do we want this by default? Given that changes to the file structure have to be collective, it seems sensible. Just wondering if there are other implications.

matz-e (approved):

include/highfive/bits/H5FileDriver_misc.hpp
1uc: Fair enough, we don't want to support that driver anymore. Therefore, changing it in any way can only break things.

roblatham00: I agree that collective metadata reads and writes should be opt-in, just as they are in HDF5 itself.   In my domain, I'm always behaving in a collective-friendly manner but I know that is not always the case with HDF5 users.

include/highfive/bits/H5PropertyList_misc.hpp
```diff
@@ -89,6 +89,31 @@ inline void RawPropertyList<T>::add(const F& funct, const Args&... args) {
 
 // Specific options to be added to Property Lists
 
+#ifdef H5_HAVE_PARALLEL
+
+void MPIOCollectiveMetadata::apply(const hid_t plist) const {
+    auto read = MPIOCollectiveMetadataRead{collective_};
+    auto write = MPIOCollectiveMetadataWrite{collective_};
+
```
roblatham00, line 97: sigh... there are probably some users out there somewhere that want collective write but independent read, or vice versa.

1uc: In which case, with property list `plist`, they would ask for collective reading:
```
plist.add(MPIOCollectiveMetadataRead{});
```
or collective writing:
```
plist.add(MPIOCollectiveMetadataWrite{});
```

The third option `MPIOCollectiveMetadata` reduces the typing effort and increases discoverability of the correct settings for the case where all operations are collective.

matz-e (approved):

PR #629, opened 2022-11-02
Add a function to get property list from object

alkino:
Usage
```c++

using namespace HighFive;

int main() {
    File file{"a.h5"};
    auto fapl = file.getAccessPropertyList();
    return 0;
}
```

### DIFF
--- a/include/highfive/H5PropertyList.hpp
+++ b/include/highfive/H5PropertyList.hpp
@@ -44,6 +44,7 @@ enum class PropertyType : int {
     LINK_ACCESS,
 };
 
+namespace details {
 template <typename T, typename U>
 T get_plist(const U& obj, hid_t (*f)(hid_t)) {
     auto hid = f(obj.getId());
@@ -54,6 +55,7 @@ T get_plist(const U& obj, hid_t (*f)(hid_t)) {
     t._hid = hid;
     return t;
 }
+}  // namespace details
 
 ///
 /// \brief Base Class for Property lists, providing global default
@@ -68,7 +70,7 @@ class PropertyListBase: public Object {
 
   private:
     template <typename T, typename U>
-    friend T get_plist(const U&, hid_t (*f)(hid_t));
+    friend T details::get_plist(const U&, hid_t (*f)(hid_t));
 };
 
 ///
--- a/include/highfive/H5PropertyList.hpp
+++ b/include/highfive/H5PropertyList.hpp
@@ -11,8 +11,6 @@
 
 #include <vector>
 
-#include <H5Apublic.h>
-#include <H5Gpublic.h>
 #include <H5Ppublic.h>
 
 // Required by MPIOFileAccess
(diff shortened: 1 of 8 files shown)


include/highfive/H5PropertyList.hpp
```diff
@@ -88,6 +99,14 @@ class PropertyList: public PropertyListBase {
     void _initializeIfNeeded();
 };
 
+const auto& getFileCreateList = [](const File& obj) {return get_plist<PropertyType::FILE_CREATE>(obj, H5Fget_create_plist);};
+const auto& getFileAccessList = [](const File& obj) {return get_plist<PropertyType::FILE_ACCESS>(obj, H5Fget_access_plist);};
+const auto& getDataSetCreateList = [](const DataSet& obj) {return get_plist<PropertyType::DATASET_CREATE>(obj, H5Dget_create_plist);};
+const auto& getDataSetAccessList = [](const DataSet& obj) {return get_plist<PropertyType::DATASET_ACCESS>(obj, H5Dget_access_plist);};
+const auto& getGroupCreateList = [](const Group& obj) {return get_plist<PropertyType::GROUP_CREATE>(obj, H5Gget_create_plist);};
+const auto& getDataTypeCreateList = [](const DataType& obj) {return get_plist<PropertyType::DATATYPE_CREATE>(obj, H5Tget_create_plist);};
+const auto& getAttributeCreateList = [](const Attribute& obj) {return get_plist<PropertyType::ATTRIBUTE_CREATE>(obj, H5Aget_create_plist);};
```
1uc: Maybe we can reduce the number of different names:
```
FileAccessProps getAccessList(const File& obj) {
    return get_plist<PropertyType::FILE_ACCESS>(obj, H5Fget_access_plist);
}

FileCreateProps getCreateList(const File& obj) {
    return get_plist<PropertyType::FILE_CREATE>(obj, H5Fget_create_plist);
}

DatasetAccessProps getAccessList(const Dataset& obj) {
    return get_plist<PropertyType::DATASET_ACCESS>(obj, H5Dget_access_plist);
}

// ...

```

matz-e (approved):

include/highfive/H5PropertyList.hpp
1uc: In terms of which case to use for free functions:
https://github.com/BlueBrain/HighFive/blob/master/include/highfive/H5DataType.hpp#L252

But this shows we've been doing both (see under Functions):
https://bluebrain.github.io/HighFive/namespace_high_five.html

include/highfive/H5PropertyList.hpp
```diff
@@ -64,9 +66,6 @@ class PropertyListBase: public Object {
         return plist;
     }
 
-  protected:
-    using Object::Object;
-
```
1uc: For my understanding: Removing these lines is just for sake of tidying up, right? Because (most) `Object::Object` are already protected, they wouldn't become public anyway.

include/highfive/H5PropertyList.hpp
```diff
@@ -64,9 +66,6 @@ class PropertyListBase: public Object {
...
   private:
     template <typename T, typename U>
     friend T get_plist(const U&, hid_t (*f)(hid_t));
```
1uc, line 71: What if you moved these lines to the `PropertyList`, would that enable `T{hid}`? (Maybe friendliness isn't inherited in old/some compilers?

1uc (dismissed):
Thank you! This is very useful.

matz-e (approved):
`Property` is dropped from the method names, as in `getCreatePropertyList` would be a wee bit cleaner. But I'm fine with how it is.

include/highfive/H5PropertyList.hpp
alkino: I don't want to put this function in the public API of property list.

1uc (approved):

1uc (approved):

PR #635, opened 2022-11-07
Enable setting the page buffer size.

1uc:
For page allocated/aggregated files, one can configure a page buffer. The page buffer is useful, when multiple datasets occupy one page, or when a lot of metadata is accessed.


1uc:
There's an issue that when I remove the `#ifndef` and compile against a parallel version of HDF5, then HDF5 print an error message:
```
7: HDF5-DIAG: Error detected in HDF5 (1.12.2) thread 0:
7:   #000: H5F.c line 620 in H5Fopen(): unable to open file
7:     major: File accessibility
7:     minor: Unable to open file
7:   #001: H5VLcallback.c line 3501 in H5VL_file_open(): failed to iterate over available VOL connector plugins
7:     major: Virtual Object Layer
7:     minor: Iteration failed
7:   #002: H5PLpath.c line 578 in H5PL__path_table_iterate(): can't iterate over plugins in plugin path '(null)'
7:     major: Plugin for dynamically loaded library
7:     minor: Iteration failed
7:   #003: H5PLpath.c line 620 in H5PL__path_table_iterate_process_path(): can't open directory: /usr/local/hdf5/lib/plugin
7:     major: Plugin for dynamically loaded library
7:     minor: Can't open directory or file
7:   #004: H5VLcallback.c line 3351 in H5VL__file_open(): open failed
7:     major: Virtual Object Layer
7:     minor: Can't open object
7:   #005: H5VLnative_file.c line 97 in H5VL__native_file_open(): unable to open file
7:     major: File accessibility
7:     minor: Unable to open file
7:   #006: H5Fint.c line 1950 in H5F_open(): page buffering is disabled for parallel
7:     major: File accessibility
7:     minor: Unable to open file
7: Filters: Test page buffer size
```

matz-e (approved):
Not happy about the parallelism restriction, but we can't do anything about that :(

pramodk:
> > There's an issue that when I remove the #ifndef and compile against a parallel version of HDF5, then HDF5 print an error message:

> Not happy about the parallelism restriction, but we can't do anything about that :(

A bit annoying limitation if we have to also link to a serial version. @1uc: wondering if it's worth asking on hdf5 forum if there is any workaround/alternatives.

1uc:
There's this:
https://github.com/HDFGroup/hdf5/blob/94119211a74ce966380d0013efd48c6effe01f10/src/H5Fint.c#L1922

Which smells like something to avoid dealing with an issue during debugging the feature (it's been there since page buffers were introduced). The RFC seems to suggest that it's safe see, 7.2 Parallel Regression Tests:
https://docs.hdfgroup.org/archive/support/HDF5/docNewFeatures/PageBuffering/RFC-Page_Buffering.pdf

1uc:
I've asked upstream:
https://forum.hdfgroup.org/t/page-buffer-with-phdf5/10435

Currently, using the page buffer isn't possible in the parallel version of HDF5. Hence, for now this is the best we can do.

PR #636, opened 2022-11-07
Check also datatype sizes for floats to avoid losing bits

ohm314:
**Description**

HDF5 treats floats and doubles as Float, one datatype class. so reading a double dataset into a float vector won't even cause a warning message. This PR extends DataTypeClass a little bit to allow comparing DataTypeClasses with bit operators and then treats the case specially where datatypes are floating types and also checks the type size, printing a warning if there is a mismatch.

**How to test this?**

Here's a minimal patch to one of the examples that tests this:

```
--- a/src/examples/create_dataset_double.cpp
+++ b/src/examples/create_dataset_double.cpp
@@ -6,6 +6,7 @@
  *          http://www.boost.org/LICENSE_1_0.txt)
  *
  */
+#include <array>
 #include <iostream>
 #include <string>
 #include <vector>
@@ -37,6 +38,15 @@ int main(void) {
         // write it
         dataset.write(data);

+        std::vector<std::array<float, 6>> lore(2);
+        DataSet readback = file.getDataSet(DATASET_NAME);
+        readback.read(lore);
+        for (const auto& arr: lore) {
+            for (const auto e: arr) {
+                std::cout << " " << e;
+            }
+            std::cout << std::endl;
+        }
     } catch (Exception& err) {
         // catch and print any HDF5 error
         std::cerr << err.what() << std::endl;
```

```bash
cmake .. -DHIGHFIVE_EXAMPLES=ON
make -j8
src/examples/create_dataset_double_bin
```

### DIFF
--- a/include/highfive/H5DataType.hpp
+++ b/include/highfive/H5DataType.hpp
@@ -24,18 +24,18 @@ namespace HighFive {
 /// \brief Enum of Fundamental data classes
 ///
 enum class DataTypeClass {
-    Time,
-    Integer,
-    Float,
-    String,
-    BitField,
-    Opaque,
-    Compound,
-    Reference,
-    Enum,
-    VarLen,
-    Array,
-    Invalid
+    Time = 1 << 1,
+    Integer = 1 << 2,
+    Float = 1 << 3,
+    String = 1 << 4,
+    BitField = 1 << 5,
+    Opaque = 1 << 6,
+    Compound = 1 << 7,
+    Reference = 1 << 8,
+    Enum = 1 << 9,
+    VarLen = 1 << 10,
+    Array = 1 << 11,
+    Invalid = 0
 };
 
 inline DataTypeClass operator|(DataTypeClass lhs, DataTypeClass rhs) {
--- a/include/highfive/H5DataType.hpp
+++ b/include/highfive/H5DataType.hpp
@@ -9,6 +9,7 @@
 #ifndef H5DATATYPE_HPP
 #define H5DATATYPE_HPP
 
+#include <type_traits>
 #include <vector>
 
 #include "H5Object.hpp"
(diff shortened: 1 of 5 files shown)


matz-e (dismissed):
Looks good _to me_

include/highfive/H5DataType.hpp
```diff
@@ -37,6 +38,16 @@ enum class DataTypeClass {
...
 };
 
+inline DataTypeClass operator|(DataTypeClass lhs, DataTypeClass rhs) {
+    using T = std::underlying_type<DataTypeClass>::type;
+    return static_cast<DataTypeClass>(static_cast<T>(lhs) | static_cast<T>(rhs));
+}
+
+inline DataTypeClass operator&(DataTypeClass lhs, DataTypeClass rhs) {
+    using T = std::underlying_type<DataTypeClass>::type;
+    return static_cast<DataTypeClass>(static_cast<T>(lhs) & static_cast<T>(rhs));
```
1uc, line 48: This looks like bitfield manipulations to me, but when I look at `DataTypeClass` it's just an `enum class` without giving the individual values a particular value. In a small reproducer I get that they're numbered through: `0, 1, 2, 3, ..., n`. Therefore, I'm not sure I understand theses lines of code correctly.

ohm314: This is just some convenient trick to be able to check multiple instances of this enum with a specific value. Like this allows you to more elegantly write this

```c++
if ((dtype.getClass() & data_type.getClass()) == DataTypeClass::Float) {
    // ...
}
```

instead of this

```c++
if ((dtype.getClass() == DataTypeClass::Float) and (data_type.getClass() == DataTypeClass::Float)) {
    // ...
}
```

if it were enums instead of enum classes, they would implicitly convert to ints and you coudl do this without defining bitwise operators, but for enum classes, one needs to add the operator. Full disclosure, the trick comes (of course, where else) from here: https://softwareengineering.stackexchange.com/a/204566

ohm314: Let me rephrase: ~elegantly~ briefly
since not everyone will find that elegant.

1uc: By assigning powers of two as values, the bit logic should workout.

1uc (approved):

PR #643, opened 2022-11-09
Allow finding HDF5 in config mode.

1uc:
This PR allows finding HDF5 when only `*-config.cmake` files are present, i.e. when the find module can't find HDF5, but in config mode the dependency could be found.


matz-e (approved):
May want to add this to changelog and bump stuff some more?

matz-e (approved):

PR #646, opened 2022-11-10
Test for several versions of libhdf5

alkino:


.github/workflows/ci.yml
```diff
@@ -74,6 +74,40 @@ jobs:
...
+      run: |
+        sudo apt-get -qq update
+        sudo apt-get -qq install ninja-build libsz2 zlib1g-dev
+
+    - name: Build HDF5
+      run: |
+        git clone https://github.com/HDFGroup/hdf5 --depth 1 --branch ${{ matrix.hdf5_version }}
+        mkdir -p hdf5/BUILD && cd hdf5/BUILD
+        cmake .. -GNinja -DCMAKE_INSTALL_PREFIX=$PWD/I -DHDF5_ENABLE_Z_LIB_SUPPORT=ON -DHDF5_ENABLE_SZIP_SUPPORT=ON -DHDF5_ENABLE_SZIP_ENCODING=ON -DUSE_LIBAEC=ON
+        ninja && ninja install
```
pramodk, line 99: Just an alternative (that we use in neuron) in case building HDF5 from source would become problematic/too slow: 1) we can keep platform fixed (e.g. ubuntu 22.04) 2) build hdf5 library with system compiler and store the binary artefacts somewhere 3) use those artefacts here in the CI

Downside is we will need to rebuild HDF5 artefacts but that we will need once in a while when we change base OS image (which we don't need to do often).

pramodk: Or, just using one of the binary distribution that they provide with centos7/centos8 platform:
image

.github/workflows/ci.yml
```diff
@@ -74,6 +74,40 @@ jobs:
...
+    - name: "Install libraries"
+      run: |
+        sudo apt-get -qq update
+        sudo apt-get -qq install ninja-build libsz2 zlib1g-dev libboost-all-dev
+
+    - name: Build HDF5
+      run: |
+        git clone https://github.com/HDFGroup/hdf5 --depth 1 --branch ${{ matrix.hdf5_version }}
+        mkdir -p hdf5/BUILD && cd hdf5/BUILD
+        cmake .. -GNinja -DCMAKE_INSTALL_PREFIX=$HOME/I -DHDF5_ENABLE_Z_LIB_SUPPORT=ON -DUSE_LIBAEC=ON
```
ohm314, line 98:
```suggestion
        cmake .. -GNinja -DCMAKE_INSTALL_PREFIX=$HOME/${{ matrix.hdf5_version }} -DHDF5_ENABLE_Z_LIB_SUPPORT=ON -DUSE_LIBAEC=ON
```

.github/workflows/ci.yml
```diff
@@ -74,6 +74,40 @@ jobs:
...
+        ninja && ninja install
+
+    - name: Build
+      run: |
+        mkdir BUILD && cd BUILD
+        cmake -GNinja -DHDF5_ROOT=$HOME/I ..
```
ohm314, line 104:
```suggestion
        cmake -GNinja -DHDF5_ROOT=$HOME/${{ matrix.hdf5_version }} ..
```

ohm314 (changes requested):
to make things more readable.

.github/workflows/ci.yml
alkino: done

1uc (approved):

ohm314 (approved):
lgtm

Issue #203, opened 2019-09-26
Writing/Reading from a in-memory file/buffer?


wuyuanyi135:
Hello,
I wonder whether HighFive supports loading or dumping the data from/into a in-memory buffer, instead of a file with given filename? 

Regards,
YW

wuyuanyi135:
I have found a solution in [so](https://stackoverflow.com/questions/43721534/create-hdf5-file-from-scratch-using-file-image-operations-memory-mapped-hdf5-fi).
First, create a new class derived 
``` cpp
#include <highfive/H5PropertyList.hpp>
#include <hdf5.h>
class h5_memory_access : public HighFive::FileAccessProps {
 public:
  h5_memory_access(): HighFive::FileAccessProps() {
    _initializeIfNeeded();
      auto result = H5Pset_fapl_core(getId(), 1 << 22, false);
      assert(result >= 0 && "H5Pset_fapl_core failed");
  }

  void get_buffer(std::vector<uint8_t>& buf) {
      ssize_t sz=H5Fget_file_image(getId(),nullptr ,0);
      H5Fget_file_image(getId(),buf.data(),sz);
  }
};
```

Then, 
```cpp
    const h5_memory_access &props = simex::h5_memory_access();
    HighFive::File file = HighFive::File("/tmp/whatever", H5F_ACC_TRUNC, props);
```

note that the file name would be arbitrary. `/tmp/whatever` would never be created.

Finally, the data bytes can be retrieved with `get_buffer` method.

ferdonline:
@wuyuanyi135 Thanks for sharing your solution. Besides creating a new derived class (which is the best way of doing it) I believe you could alternatively use `RawPropertyList`s.
```
RawPropertyList<PropertyType::FILE_ACCESS> lcpl;
lcpl.add(H5Pset_fapl_core, 1 << 22, false);
```

sighingnow:
To whom who need to reading a `HighFive::File` from memory buffer, the pull request https://github.com/BlueBrain/HighFive/pull/648 would make that possible:

```cpp
  hid_t file_id = H5LTopen_file_image(const_cast<void*>(data), size,
                                      H5LT_FILE_IMAGE_DONT_RELEASE);
  hid_t root_id = H5Gopen(file_id, "/", H5P_DEFAULT);
  HighFive::File file(root_id);
```

PR #648, opened 2022-11-19
Construct HDF5 file from existing hid_t

sighingnow:
**Description**

This pull request enables creating a `HighFive::File` from existing `hid_t`, to allow use patterns like reading HDF5 files from memory buffers:

```cpp
  hid_t file_id = H5LTopen_file_image(const_cast<void*>(data), size,
                                      H5LT_FILE_IMAGE_DONT_RELEASE);
  hid_t root_id = H5Gopen(file_id, "/", H5P_DEFAULT);
  HighFive::File file(root_id);
```

### DIFF
--- a/include/highfive/H5File.hpp
+++ b/include/highfive/H5File.hpp
@@ -70,6 +70,12 @@ class File: public Object, public NodeTraits<File>, public AnnotateTraits<File> 
          const FileAccessProps& fileAccessProps = FileAccessProps::Default());
 
     ///
+    /// \brief File creating a HighFive::File from and already opended HDF5 file.
+    ///
+    /// \param hid: existing hid, usually created by H5Gopen (follows H5LTopen_file_image)
+    explicit File(const hid_t hid);
+
+    ///
     /// \brief Return the name of the file
     ///
     const std::string& getName() const noexcept;
--- a/include/highfive/bits/H5File_misc.hpp
+++ b/include/highfive/bits/H5File_misc.hpp
@@ -43,7 +43,6 @@ inline File::File(const std::string& filename,
                   const FileAccessProps& fileAccessProps)
     : File(filename, openFlags, FileCreateProps::Default(), fileAccessProps) {}
 
-
 inline File::File(const std::string& filename,
                   unsigned openFlags,
                   const FileCreateProps& fileCreateProps,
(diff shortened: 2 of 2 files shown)


sighingnow:
See also the write-side discussion in https://github.com/BlueBrain/HighFive/issues/203

sighingnow:
All check passed. Can anyone help to get this pull request merged?

Thanks!

1uc:
Thank you for your pull request. What you're proposing implies strong changes to how HDF5 resources, such as the file ID can enter into the HighFive library. One of the questions raised by the code you post is: who's responsible for freeing the `file_id` and `root_id`. I'm guessing `file_id` is being leaked and `root_id` would be released by HighFive. Currently, these problems should not occur with HighFive, because it allocates and frees all HDF5 resources.

There's other ways of doing the same thing, e.g. through properties, see:
https://docs.hdfgroup.org/archive/support/HDF5/doc1.8/Advanced/FileImageOperations/HDF5FileImageOperations.pdf

While that method would fit much more neatly into HighFive, it's to quote the mentioned PDF:

> A potential problem with the H5Pget/set_file_image calls is the overhead of allocating and copying of large
file image buffers. The callback routines enable application programs to avoid this problem. However, the use
of these callbacks is complex and potentially hazardous [...]

sighingnow:
> One of the questions raised by the code you post is: who's responsible for freeing the file_id and root_id. I'm guessing file_id is being leaked and root_id would be released by HighFive. Currently, these problems should not occur with HighFive, because it allocates and frees all HDF5 resources.

Thanks for the comments. I admit that is a concern in current implementation. How about using the pointer `void *data` and size `size_t size` as the parameter and internally creating the `file_id` as well as `root_id`? That could be enough for correct lifecycle management.

I cannot see how such a functionality can be archieved using properties? In #203 I only see how to control the hdf5 core not write back to file. Could you please show some example about reading a HDF5 file from memory buffer?

Thanks!

1uc:
We've discussed alternatives. One idea that came up is to simply subclass `HighFive::File`. It's sketched out in #652. Please have a look and let us know if this satisfies your needs.

One big advantage is, that you can progress without waiting for and requiring a particular version of HighFive.

sighingnow:
> Please have a look and let us know if this satisfies your needs.

It does work for me. Thanks! I have left a minor comments on that thread.

sighingnow:
Closing as in favour of #652.

PR #652, opened 2022-11-23
Example of reading an in memory HDF5 file.

1uc:
As discussed internally, one way of adding functionality related to inmemory file would be to subclass `HighFive::File`. In the PR this option is sketched out as an example.

This could be an alternative to implementing #648

### DIFF
--- a/include/highfive/H5File.hpp
+++ b/include/highfive/H5File.hpp
@@ -111,6 +111,9 @@ class File: public Object, public NodeTraits<File>, public AnnotateTraits<File> 
         return details::get_plist<FileAccessProps>(*this, H5Fget_access_plist);
     }
 
+  protected:
+    File() = default;
+
   private:
     using Object::Object;
 
--- a/src/examples/CMakeLists.txt
+++ b/src/examples/CMakeLists.txt
@@ -32,6 +32,11 @@ function(compile_example exemple_source)
     add_executable(${example_name} ${exemple_source})
     target_link_libraries(${example_name} HighFive)
 
+    if(${example_name} MATCHES ".*hl_hdf5.*")
+        find_package(HDF5 REQUIRED C HL)
+        target_link_libraries(${example_name} ${HDF5_HL_LIBRARIES})
+    endif()
+
 endfunction()
 
 file(GLOB list_example "*.cpp")
--- a/src/examples/hl_hdf5_inmemory_files.cpp
+++ b/src/examples/hl_hdf5_inmemory_files.cpp
@@ -1,0 +1,68 @@
+/*
+ *  Copyright (c), 2022, Blue Brain Project
+ *
+ *  Distributed under the Boost Software License, Version 1.0.
+ *    (See accompanying file LICENSE_1_0.txt or copy at
+ *          http://www.boost.org/LICENSE_1_0.txt)
+ *
+ */
+#include <iostream>
+#include <vector>
+
+
+#include <highfive/H5File.hpp>
+#include <hdf5_hl.h>
+
+using namespace HighFive;
+
+class InMemoryFile: public HighFive::File {
+  public:
+    explicit InMemoryFile(std::vector<std::uint8_t> buffer)
+        : _buffer(std::move(buffer)) {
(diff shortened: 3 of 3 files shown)


include/highfive/H5File.hpp
```diff
@@ -111,6 +111,9 @@ class File: public Object, public NodeTraits<File>, public AnnotateTraits<File>
         return details::get_plist<FileAccessProps>(*this, H5Fget_access_plist);
     }
 
+  protected:
+    File() = default;
+
```
1uc, line 116: This protected default constructor is needed for MSVC which otherwise complains that there is no suitable default constructor.

src/examples/hl_hdf5_inmemory_files.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+using namespace HighFive;
+
+class InMemoryFile: public HighFive::File {
+  public:
+    explicit InMemoryFile(std::vector<std::uint8_t> buffer)
+        : _buffer(std::move(buffer)) {
+        _hid = H5LTopen_file_image(_buffer.data(),
+                                   sizeof(_buffer[0]) * _buffer.size(),
+                                   H5LT_FILE_IMAGE_DONT_RELEASE | H5LT_FILE_IMAGE_DONT_COPY);
```
sighingnow, line 24: I think you need use the returned `hid_t` by `H5Gopen(file_id, "/", H5P_DEFAULT);`, as mentioned in #648.

I'm not very familiar with libhdf5, but from my previous experiences, using `hid_t` returned by `H5LTopen_file_image()` as `_hid` won't work.

include/highfive/H5File.hpp
matz-e: Can that be a comment in the source?

src/examples/CMakeLists.txt
```diff
@@ -32,6 +32,11 @@ function(compile_example exemple_source)
     add_executable(${example_name} ${exemple_source})
     target_link_libraries(${example_name} HighFive)
 
+    if(${example_name} MATCHES ".*hl_hdf5.*")
+        find_package(HDF5 REQUIRED C HL)
+        target_link_libraries(${example_name} ${HDF5_HL_LIBRARIES})
+    endif()
```
matz-e, line 38: Do we execute these examples, too?

include/highfive/H5File.hpp
1uc: I'm not too convinced. My impression is that we don't comment on this type of thing:
* https://github.com/BlueBrain/HighFive/blob/master/include/highfive/H5Object.hpp#L71
* https://github.com/BlueBrain/HighFive/blob/master/include/highfive/H5Group.hpp#L31
* https://github.com/BlueBrain/HighFive/blob/master/include/highfive/H5DataSpace.hpp#L108

Instead we have CI that checks if it builds on various platforms. The comment is difficult to write in a precise way, because I don't have easy access to MS compilers other than through CI. Additionally, in a year it might work with MSVC, just not with the version we use in CI.

src/examples/CMakeLists.txt
1uc: I can't find anything pointing to us running the examples, no. I've run it locally ofc.

src/examples/hl_hdf5_inmemory_files.cpp
1uc: Sorry about the delay. The answer, and part of the reason for the delay, is I simply don't know. I'm not observing the same issue, the example worked as shown. But I'm not doing anything complex in the example.

At a later point we can try and improve the example by opening and closing a group; but for the time being I hope it addresses your immediate requirements.

The one difference I'm confident does matter is the `| H5LT_FILE_IMAGE_DONT_COPY` because without it, a copy is performed and the first `*_DONT_RELEASE` is ignored.

sighingnow: Thanks. Introduce the protect ctor and using a aubclass does work for me.

alkino (approved):

Issue #216, opened 2019-10-10
Add Chunking property automatically if this is resizable.


alkino:
We should look at what hdf5.py does and do the same.

ferdonline:
check here: https://github.com/h5py/h5py/blob/4c7477bbd26e18daefb1fe7d8d2a35c05a816901/h5py/_hl/filters.py#L202
h5py by default will automatically calculate in case chunking is needed.

tpongo-afk:
Is anyone working on this? I'm interested in the resolution.

alkino:
Do you mean you want to work on it? If yes, feel free!

No one is currently working on it. I can assign the ticket to you

tpongo-afk:
Yes, I would give it a try.

alkino:
thanks for the try, you can ask us question if needed

PR #679, opened 2023-01-26
Add chunking property if necessary and not set by user

tpongo-afk:
**Description**

Adds Chunking property if the data set is extendable, compressed or the shuffle filter is set and Chunking was not set by the user.
Chunk size guessing is based on the current h5py implementation (https://github.com/h5py/h5py/blob/0e6dbac0a01d85e8ce2d11f3a567b0526c325d09/h5py/_hl/filters.py#L338).

Related issue: #216 

**How to test this?**

Should I write test cases? Let me know. I'm new to that to be honest.
Anyways I tried a few combinations and they worked.

**Test System**
 - OS: Ubuntu 22.04
 - Compiler: g++ 11.3.0

### DIFF
--- a/include/highfive/bits/H5PropertyList_misc.hpp
+++ b/include/highfive/bits/H5PropertyList_misc.hpp
@@ -363,9 +363,10 @@ template <typename... Args>
 inline Chunking::Chunking(hsize_t item, Args... args)
     : Chunking(std::vector<hsize_t>{item, static_cast<hsize_t>(args)...}) {}
 
-std::vector<std::size_t> Chunking::guessChunkingSize(const std::vector<std::size_t>& dims,
-                                                     const std::vector<std::size_t>& max_dims,
-                                                     std::size_t typesize) {
+inline std::vector<std::size_t> Chunking::guessChunkingSize(
+    const std::vector<std::size_t>& dims,
+    const std::vector<std::size_t>& max_dims,
+    std::size_t typesize) {
     const std::size_t CHUNK_BASE = 16 * 1024;   // Multiplier by which chunks are adjusted
     const std::size_t CHUNK_MIN = 8 * 1024;     // Soft lower limit (8k)
     const std::size_t CHUNK_MAX = 1024 * 1024;  // Hard upper limit (1M)
--- a/include/highfive/bits/H5PropertyList_misc.hpp
+++ b/include/highfive/bits/H5PropertyList_misc.hpp
@@ -180,7 +180,9 @@ inline void Chunking::apply(const hid_t hid) const {
     }
 }
 
-std::vector<std::size_t> Chunking::guessChunkingSize(const std::vector<std::size_t>& dims, const std::vector<std::size_t>& max_dims, std::size_t typesize) {
+std::vector<std::size_t> Chunking::guessChunkingSize(const std::vector<std::size_t>& dims,
+                                                     const std::vector<std::size_t>& max_dims,
+                                                     std::size_t typesize) {
     const std::size_t CHUNK_BASE = 16 * 1024;   // Multiplier by which chunks are adjusted
     const std::size_t CHUNK_MIN = 8 * 1024;     // Soft lower limit (8k)
     const std::size_t CHUNK_MAX = 1024 * 1024;  // Hard upper limit (1M)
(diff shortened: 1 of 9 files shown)


alkino:
Thank you!
We are looking at this patch now and are discussing about API. We will come back to you.
I already made a little cleaning of the change to follow the rules and replace `product` by an already existing function doing the same thing.

tpongo-afk:
Thanks! I have to add that I found an issue in the meantime regarding the h5py implementation: https://github.com/h5py/h5py/issues/2029
In short this guessChunk function might produce large files in some cases for extendable datasets. My interest was actually in the compressed datasets having a default chunking, for that purpose, this code is fine.

alkino:
To be able to test the feature I start to split `_chunksIfNecessary` in atome.

alkino:
This is the final version

include/highfive/bits/H5PropertyList_misc.hpp
```diff
@@ -342,10 +353,69 @@ inline const std::vector<hsize_t>& Chunking::getDimensions() const noexcept {
...
+            chunk_size < CHUNK_MAX) {
+            break;
+        }
+
+        if (details::compute_total_size(chunkingDims) == 1) {
+            break;  // Element size larger than CHUNK_MAX
+        }
+
+        chunkingDims[idx % chunkingDims.size()] = static_cast<std::size_t>(
+            std::ceil(static_cast<double>(chunkingDims[idx % chunkingDims.size()]) / 2.));
```
1uc, line 412: The converse of wide table would be very narrow tables, e.g. as they occur in FVM simulations. In computation fluid dynamics (depending on the equation), but commonly there's `2`, ..., `8` conserved variables. Hence, one could easily have a shape `(nt, nx, ny, nz, 5)`, with moderately large `nx`, `ny`, `nz`; and unlimited `nt`. We might end up dividing the `5` by `2`, resulting in a chunk shape of `(ct, cx, cy, cz, 3)`. To me this is reason to be skeptical because: a) the chunk size is not a power of 2; and b) we'll need to chunks to fetch the conserved variables.

The power of two issue I feel is awkward. Note, that it occurs quite generally. Whenever one of `dims[i]` isn't a power of two, the chunk size will likely not be either.

Data is read efficiently from disk and RAM in contiguous chunks of a power of 2 size. In page allocated files, the chunks won't fill the page, leaving a gap at the end. This gap can either be filled by small (chunks of) datasets, causing fragmentation of the file; or left empty. These chunks might not fit very nicely into caches.

The size inflation issue is maybe less of a concern. For `(n, 3, 3)` is 1.7x larger; for `(n, 3, 3, 3, 3)` is 3.1x (maybe that's too special of a case). In the context of compression, it seems unfortunate to start off with a 1.7x larger dataset. Yes, there's a good chance that all the extra space will be compressed very efficiently.

I'm not fully sure how max-dims works; and I need to check this. The concern is that someone would pad their dims to get powers of two, e.g. `dims = (n, 3)` with `max_dims=(n, 4)`. This algorithm would often create chunks of shape `(n, 3)`. Which has the same two issues, but this time it feels like the user specifically did something to get powers of two; but the choice of chunk undoes that work. Furthermore, the inflation aspect is more pronounced this time, it can be reasonably close to `2**d` where `d` is the number of padded dimensions. Fortunately, I think the increase in size would only happen if one of the "padded" dimensions is extended.

include/highfive/bits/H5Utils.hpp
```diff
@@ -45,6 +46,27 @@ inline std::vector<std::size_t> to_vector_size_t(const std::vector<std::size_t>&
     return vec;
 }
 
+// converter function for size_t -> hsize_t when size_t != hsize_t
+template <typename Size>
+inline std::vector<hsize_t> to_vector_hsize_t(const std::vector<Size>& vec) {
```
1uc, line 51: This function exists in similar form here:
https://github.com/BlueBrain/HighFive/blob/7acd5bab14258f52f584e250c47358ed1477276a/include/highfive/bits/H5Slice_traits.hpp#L64

1uc (changes requested):
I'm aware this is the solution implemented by h5py; and therefore, it has been in production and used a lot. Hence, it likely doesn't work as bad as I make it sound in this review.

Therefore, is there any way we could find out what motivated the h5py devs to pick this solution?

I think large parts of the PR are copy-pasted from h5py. Hence, there's licensing concerns.

include/highfive/bits/H5PropertyList_misc.hpp
```diff
@@ -342,10 +353,69 @@ inline const std::vector<hsize_t>& Chunking::getDimensions() const noexcept {
...
+
+    std::size_t dset_size = details::compute_total_size(chunkingDims) * typesize;
+    double target_size = CHUNK_BASE *
+                         std::exp2(std::log10(static_cast<double>(dset_size) / (1024. * 1024.)));
+
+    if (target_size > CHUNK_MAX) {
+        target_size = CHUNK_MAX;
+    } else if (target_size < CHUNK_MIN) {
+        target_size = CHUNK_MIN;
+    }
```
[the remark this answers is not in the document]
alkino: c++17

PR #681, opened 2023-02-04
Add a track order property

### LINKED ISSUE #670: Track creation order for the root group
I am having HighFive to write the results of an animation I am simulating. In my hierarchy, each frame is a direct children of the root group, so I have something like this

```
/0/
/1/
/2/
...
/10/
...
```

If I just let the default ordering, I would get the group `/10` before the group `/2`. Therefore, I want to track the insertion order. 

I am using this simple example to do some testing (and some variations on it)

```c++
struct A {
    void apply(hid_t hid) const
    {
        std::cout << "Adding insertion order property";
        auto status = H5Pset_link_creation_order(hid, H5P_CRT_ORDER_TRACKED | H5P_CRT_ORDER_INDEXED);
        if (status < 0) {
            std::cout << "error";
        }
    }
};

int main()
{
    HighFive::File f("test.hdf5", HighFive::File::Truncate);

    A aProp;
    auto rootG = f.getGroup("/");
    auto rpcl = rootG.getCreatePropertyList();
    rpcl.add(aProp);

    auto fCreateProps = HighFive::GroupCreateProps();
    fCreateProps.add(aProp);
    auto b = rootG.createGroup("b", fCreateProps);
    rootG.createGroup("a", fCreateProps);

    b.createGroup("d", fCreateProps);
    b.createGroup("c", fCreateProps);
}

```

I have had some success for groups that are not direct children of the root (in the example, `d` and `c`). When I run that code, in python, using h5py I get the following

```python
>>> [print(i) for i in f["/"]]                                                                                                                                
a                                                                                                                                                             
b                                                                                                                                                             
[None, None]

>>> [print(i) for i in f["/b"]]                                                                                                                               
d                                                                                                                                                            
c                                                                                                                                                             
[None, None]
```

which clearly shows that under the group `b` the ordering is being track, but not under the root group. 

Is there anyway I can add the property to the root group, or even better, to the file itself so all the groups inherit from it?

jjcasmar:
**Description**`

This PR adds a PropertyList to enable `H5P_CRT_ORDER_TRACKED | H5P_CRT_ORDER_INDEXED`, which allows to track the order in which the groups are created and to later iterate on them following that order. 

Closes #670

### DIFF
--- a/include/highfive/H5PropertyList.hpp
+++ b/include/highfive/H5PropertyList.hpp
@@ -498,6 +498,10 @@ class UseCollectiveIO {
 };
 #endif
 
+struct TrackOrderProperty {
+    void apply(hid_t hid) const;
+};
+
 }  // namespace HighFive
 
 #include "bits/H5PropertyList_misc.hpp"
--- a/include/highfive/bits/H5PropertyList_misc.hpp
+++ b/include/highfive/bits/H5PropertyList_misc.hpp
@@ -216,4 +216,11 @@ inline void UseCollectiveIO::apply(const hid_t hid) const {
 }
 #endif
 
+void de::hdf::TrackOrder::apply(hid_t hid) const {
+    if (H5Pset_link_creation_order(hid, H5P_CRT_ORDER_TRACKED | H5P_CRT_ORDER_INDEXED)) {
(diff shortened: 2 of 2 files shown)


alkino:
Sorry but this is overwritten by #683

PR #683, opened 2023-02-04
Add a Property for LinkCreationOrder

### LINKED ISSUE #680: Add property: link_creation_order
This property should be added.

https://docs.hdfgroup.org/hdf5/develop/group___g_c_p_l.html#ga24817b5c9553df3872de57c20bf11512

See #670

alkino:
Should fix #680


include/highfive/H5PropertyList.hpp
```diff
@@ -498,6 +498,32 @@ class UseCollectiveIO {
 };
 #endif
 
+enum CreationOrder {
+    Tracked = H5P_CRT_ORDER_TRACKED,
+    Indexed = H5P_CRT_ORDER_INDEXED,
+};
```
matz-e, line 504: Do we encapsulate other values like this? I didn't do this for the version bounds, wondering if this will incur more maintenance later?

At a minimum, seems like the `enum` should be inside `LinkCreationOrder`?

alkino: Yes, for example: https://github.com/BlueBrain/HighFive/blob/master/include/highfive/bits/H5DataType_misc.hpp#L36-L38

let's discuss ;)

I think, this is more convenient for user (to hide HDF5 ^^)

matz-e: Should we encapsulate more? Like the version bounds?

Anyways, I'd be fine as long as it's contained somewhere in a class, so we don't have a free `Tracked` without context :D

alkino: I agree with the class. I just wait to know if I should delete it, before moving it.

include/highfive/H5PropertyList.hpp
```diff
@@ -498,6 +498,32 @@ class UseCollectiveIO {
...
+///
+class LinkCreationOrder {
+  public:
+    ///
+    /// \brief Create the property
+    /// \param flags Should be a composition of HighFive::CreationOrder.
+    ///
+    explicit LinkCreationOrder(unsigned flags)
+        : _flags(flags) {}
+
```
1uc, line 519: We can now also read properties. Please those features to this property.

1uc (approved):

PR #684, opened 2023-02-05
Add access to values of properties

alkino:
API can be discussed

### DIFF
--- a/include/highfive/H5PropertyList.hpp
+++ b/include/highfive/H5PropertyList.hpp
@@ -138,17 +138,11 @@ class RawPropertyList: public PropertyList<T> {
 ///
 class MPIOFileAccess {
   public:
-    MPIOFileAccess(MPI_Comm comm, MPI_Info info)
-        : _comm(comm)
-        , _info(info) {}
+    MPIOFileAccess(MPI_Comm comm, MPI_Info info);
 
   private:
     friend FileAccessProps;
-    void apply(const hid_t list) const {
-        if (H5Pset_fapl_mpio(list, _comm, _info) < 0) {
-            HDF5ErrMapper::ToException<FileException>("Unable to set-up MPIO Driver configuration");
-        }
-    }
+    void apply(const hid_t list) const;
 
     MPI_Comm _comm;
     MPI_Info _info;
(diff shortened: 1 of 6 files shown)


alkino:
I wonder if we should return tuple instead of having one getter for each value.

Should we do a static function instead?

```c++
static std::tuple<hsize_t, hsize_t> fromDataTransferProps(dxpl);
```

include/highfive/H5PropertyList.hpp
```diff
@@ -244,15 +259,29 @@ class FileVersionBounds {
         : _low(low)
         , _high(high) {}
 
+    FileVersionBounds(const FileAccessProps& fapl) {
+        if (H5Pget_libver_bounds(fapl.getId(), &_low, &_high) < 0) {
+            HDF5ErrMapper::ToException<PropertyException>("Unable to access file version bounds");
+        }
+    }
+
+    H5F_libver_t get_low() const {
```
matz-e, line 268: I think here we could produce a `std::pair`

alkino: I did ;)

include/highfive/bits/H5PropertyList_misc.hpp
```diff
@@ -394,19 +394,71 @@ inline void AllocationTime::apply(hid_t dcpl) const {
 AllocationTime::AllocationTime(H5D_alloc_time_t alloc_time)
     : _alloc_time(alloc_time) {}
 
+AllocationTime::AllocationTime(const DataSetCreateProps& dcpl) {
+    if (H5Pget_alloc_time(dcpl.getId(), &_alloc_time) < 0) {
+        HDF5ErrMapper::ToException<PropertyException>("Error setting allocation time");
```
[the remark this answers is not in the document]
1uc: Thank you!

include/highfive/bits/H5PropertyList_misc.hpp
```diff
@@ -189,31 +387,136 @@ inline void Shuffle::apply(const hid_t hid) const {
...
+inline uint32_t MpioNoCollectiveCause::getLocalCause() const {
+    return _local_cause;
+}
+
+inline uint32_t MpioNoCollectiveCause::getGlobalCause() const {
+    return _global_cause;
+}
+
+inline std::pair<uint32_t, uint32_t> MpioNoCollectiveCause::getCause() const {
+    return {_local_cause, _global_cause};
```
1uc, line 518: Isn't this a C++17-feature? If so why is it getting past CI?

1uc (approved):

Issue #685, opened 2023-02-09
Boost_NO_BOOST_CMAKE TRUE


aminiussi:
**Describe the bug**
Forcing **Boost_NO_BOOST_CMAKE** to **TRUE** might prevent build whe nested as a submodule of a project using different Boost discovery settings.

**To Reproduce**

This is tricky, as I only get the issue on a HPE Cray system (which is not easy to get). I got the problem by cloning a project using my own fork of HighFive (because of [another issue](https://github.com/BlueBrain/HighFive/issues/180#issue-457350409)) with a boost-cmake enabled installation of boost.

**Expected behavior**

In `CMake/HighFiveTargetDeps.cmake` we find:
```
    set(Boost_NO_BOOST_CMAKE TRUE)  # Consistency
```
I am not sure to understand the consistency issue, but why not rely on the user settings, which would prevent conflicting with the possible super project ?

Basicaly, I would expect the HighProject to find the system and serialization library the same way the enclosing project finds them and let me decide if I want to use boost-caml=ke or not.

**Stacktrace**
```
CMake Error at /lus/work/CT4/oca7233/alainm/fargOCA/cmake-3.25.0-cce-15.0.0-7fqa/share/cmake-3.25/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Boost (missing: system serialization) (found version
  "1.80.0")
Call Stack (most recent call first):
  /lus/work/CT4/oca7233/alainm/fargOCA/cmake-3.25.0-cce-15.0.0-7fqa/share/cmake-3.25/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
  /lus/work/CT4/oca7233/alainm/fargOCA/cmake-3.25.0-cce-15.0.0-7fqa/share/cmake-3.25/Modules/FindBoost.cmake:2376 (find_package_handle_standard_args)
  external/HighFive/CMake/HighFiveTargetDeps.cmake:26 (find_package)
  external/HighFive/CMakeLists.txt:63 (include)
```

**Desktop (please complete the following information):**
 - OS: Red Hat Enterprise Linux release 8.6 (Ootpa)

**Additional context**
```
$ CC --version
Cray clang version 15.0.0  (324a8e7de6a18594c06a0ee5d8c0eda2109c6ac6)
Target: x86_64-unknown-linux-gnu
Thread model: posix
[...]

PR #687, opened 2023-02-10
Allow users to set `Boost_NO_BOOST_CMAKE`.

1uc:
Forcing `Boost_NO_BOOST_CMAKE=On` was reported to cause build failure in:
https://github.com/BlueBrain/HighFive/issues/685

This commit keeps the current default, but respects the users choice.


1uc:
Looking over the various CMake code, I feel it's in need of a general overhaul. Therefore, I prefer the solution that tries to avoid potentially breaking build systems twice in a row.

alkino (approved):

PR #698, opened 2023-03-03
Fix : Add support for std::byte as Dataset type for c++17 and above

### LINKED ISSUE #480: Add support for std::byte as Dataset type
Creating a dataset with underlying type of `std::byte` currently triggers a static assertion ("Type not supported"). This makes sense as `std::byte` is implemented as an enum rather than a scalar/numeric type. However, CPP Core Guidelines encourages the use of `std::byte` as the preferred way to refer to generic, non-character data. Does it make sense for `HighFive` to support this?

Minimal example of the issue:
```c++
#include <cstddef>
#include <vector>

using namespace HighFive;

int main()
{
    std::vector<std::byte> buffer(1024, std::byte(0xCD));
    File file("database.h5", File::ReadWrite | File::Create | File::Truncate);
    file.createDataSet<std::byte>("frame", DataSpace::From(buffer)).write(buffer);
}
```

WeinaJi:
**Description**
Add  support for std::byte as Dataset type when using c++17 and above 
- [x] Issue 1 fixed
- [x] Feature 1 added

Fixes #480 

**How to test this?**

```bash
cmake -DCMAKE_CXX_STANDARD=17 ..
make -j8
make test
```

**Test System**
 - OS: [e.g. Ubuntu 20.04]

### DIFF
--- a/tests/unit/test_all_types.cpp
+++ b/tests/unit/test_all_types.cpp
@@ -179,3 +179,36 @@ TEMPLATE_TEST_CASE("Scalar in std::vector<std::array>", "[Types]", bool, std::st
         CHECK(value.size() == 5);
     }
 }
+
+#if HIGHFIVE_CXX_STD >= 17
+TEMPLATE_TEST_CASE("Scalar in std::vector<std::byte>", "[Types]", std::byte) {
+    const std::string FILE_NAME("Test_vector_byte.h5");
+    const std::string DATASET_NAME("dset");
+    std::vector<TestType> t1(5, std::byte(0xCD));
+
+    {
+        // Create a new file using the default property lists.
+        File file(FILE_NAME, File::ReadWrite | File::Create | File::Truncate);
+
+        // Create the dataset
+        DataSet dataset = file.createDataSet(
+            DATASET_NAME,
+            {5},
+            create_datatype<typename details::inspector<std::vector<TestType>>::base_type>());
+
+        // Write into the initial part of the dataset
+        dataset.write(t1);
+    }
+
+    // read it back
+    {
+        File file(FILE_NAME, File::ReadOnly);
+
+        std::vector<TestType> value(5, std::byte(0xCD));
+        DataSet dataset = file.getDataSet("/" + DATASET_NAME);
+        dataset.read(value);
+        CHECK(t1 == value);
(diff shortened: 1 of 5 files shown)


alkino (approved):

tests/unit/test_all_types.cpp
```diff
@@ -179,3 +179,34 @@ TEMPLATE_TEST_CASE("Scalar in std::vector<std::array>", "[Types]", bool, std::st
...
+    const std::string DATASET_NAME("dset");
+    TestType t1(5, std::byte(0xCD));
+
+    {
+        // Create a new file using the default property lists.
+        File file(FILE_NAME, File::ReadWrite | File::Create | File::Truncate);
+
+        // Create the dataset
+        DataSet dataset = file.createDataSet(
+            DATASET_NAME, {5}, create_datatype<typename details::inspector<TestType>::base_type>());
```
1uc, line 195: Here a `create_datatype<std::byte>` might be preferable, because it would be closer to what users are expected to write. The concern is that a bug in `details::inspector<TestType>::base_type` might hide other bugs, e.g. the code would work with `create_type<int>` but not `create_datatype<std::byte>`.

tests/unit/test_all_types.cpp
```diff
@@ -179,3 +179,34 @@ TEMPLATE_TEST_CASE("Scalar in std::vector<std::array>", "[Types]", bool, std::st
...
+
+        // Write into the initial part of the dataset
+        dataset.write(t1);
```
1uc, line 198: Despite the above tests not doing it, it might be nice to also check:
```
        auto dataset2 = file.createDataSet(DATASET_NAME2, t1);
```


We could leave this for when we refactor the whole file to be more DRY.

1uc (approved):
Many thanks!

tests/unit/test_all_types.cpp
WeinaJi: Thanks for the comment, I've made the change.

tests/unit/test_all_types.cpp
WeinaJi: ok, I leave this for the refactoring later.

1uc (approved):

PR #701, opened 2023-03-15
Improve Cmake detection quietness, and adhere to standard procedure

### LINKED ISSUE #593: Installed package is noisy when found by CMake
This issue is more annoying than a problem, but should also be easy to fix.

In CMake, you can give the QUIET option to find_package() to suppress output. We use that to keep our CMake build free from informative, but (for our developers largely) useless messages.

HighFive is one of the few exceptions, it always prints the line with "(Re)Detecting Highfive dependencies". The generic way would be to evaluate "<packagename>_FIND_QUIETLY", which should tell you if the user requested a quiet finding.

ferdonline:
**Description**

The `packageConfig.cmake` was not respecting QUIET request. Further, it seems that we were not calling `check_required_components` which is pretty standard to validate component request.

Fixes #593

**How to test this?**

Tested locally, launching `test_project_integration.sh`

```bash
cd HighFive/tests
bash test_project_integration.sh
```

### DIFF
--- a/CMake/HighFiveConfig.cmake.in
+++ b/CMake/HighFiveConfig.cmake.in
@@ -15,6 +15,8 @@ if(TARGET HighFive)
     return()
 endif()
 
+@PACKAGE_INIT@
+
 # Get HighFive targets
 include("${CMAKE_CURRENT_LIST_DIR}/HighFiveTargets.cmake")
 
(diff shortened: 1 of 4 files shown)


1uc:
I don't have a strong opinion about this PR. However, I've just spent over 3 days trying to understand what HighFive CMake does; and how it interacts with Spack. Therefore, I'd prefer if this wasn't part of `2.7.0`.

Generally, I think we need to burn down our CMake and start over with something closer to #257. For that reason I'm not sure it's worth merging it. There's a nice trick in here that we can definitely use.

alkino (approved):

PR #702, opened 2023-03-15
Fix and test empty corner cases.

1uc:
The integration tests show that we have issues when dealing with empty datasets, both when reading and writing them. This commit adds testing to cover reading and writing empty datasets using `std::vector`, `Eigen::*` and `boost::multi_array` for datasets and attributes. Furthermore, it fixes the discovered bugs.

### DIFF
--- a/include/highfive/bits/H5Attribute_misc.hpp
+++ b/include/highfive/bits/H5Attribute_misc.hpp
@@ -73,6 +73,15 @@ inline void Attribute::read(T& array) const {
         throw DataSpaceException(ss.str());
     }
     auto dims = mem_space.getDimensions();
+
+    if (mem_space.getElementCount() == 0) {
+        auto effective_dims = details::squeezeDimensions(dims,
+                                                         details::inspector<T>::recursive_ndim);
+
+        details::inspector<T>::prepare(array, effective_dims);
+        return;
+    }
+
     auto r = details::data_converter::get_reader<T>(dims, array);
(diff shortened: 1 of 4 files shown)


1uc:
These are the result for the integration tests (green):
https://github.com/BlueBrain/HighFive-testing/actions/runs/4427635111/jobs/7765526499#step:6:35

1uc:
The issues to consider are:
* Even when loading an empty dataset, since we perform allocation, we need to ensure that the returned object is appropriately empty.
* For multi-dimensional arrays, there's a difference between `(n, m, 0)` and `(0, m, l)`, e.g. for nested `std::vector`s the former is a vector with `n` vectors, each containing `m` empty vectors. The latter is just an empty vector.
* We might need to consider that the shape of `std::vector<std::array<T, 2>>` is always `(n, 2)`. Therefore, even if `n == 0` the next dimension should be deduced as `2`. Currently, we unable to get the shape of compile-time constant dimensions correct, in general. We return `1` instead.
* We currently don't check that in `write_raw` the buffer isn't the `nullptr`, nor do we turn it into a no-op. Instead we rely on HDF5 doing the appropriate checks, e.g. if a positive number of elements are to be read into the `nullptr`.

ferdonline:
I have checked around and LGTM. It's a pity we have dome duplication in `Dataset::write` and `Slice::write`, but improving that is another story.

ferdonline (approved):

PR #706, opened 2023-03-16
Remove outdated badge.

1uc:
This PR proposes to remove the codecov CI because it's been highly unreliable recently. I'd like to avoid getting used to failing CI. Therefore, I think it's best if we simply remove it for the time being.

Next this also removes a badge for static analysis which ran last in 2017.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,6 @@
 # HighFive - HDF5 header-only C++ Library
 
 [![Build Status](https://travis-ci.org/BlueBrain/HighFive.svg?branch=master)](https://travis-ci.org/BlueBrain/HighFive)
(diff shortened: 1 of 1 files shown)


.github/workflows/coverage.yml
```diff
@@ -22,45 +22,3 @@ on:
...
-
-jobs:
-  # Job producing code coverage report
-  # ==================================
-  Code_coverage:
-    # When ubuntu version will be upgrade, you should look if xtensor is now available
-    runs-on: ubuntu-20.04
-
-    steps:
-    - name: "Install libraries"
```
ohm314, line 34: Maybe instead of removing it completely let's just disable it by adding: `if: ${{ false }}` (found [here](https://code-maven.com/slides/github-ci/disable-single-github-action-job).

1uc:
This should have been fixed via #708

alexsavulescu (approved):

PR #710, opened 2023-03-24
Allow HIGHFIVE_USE_INSTALL_DEPS only at install time

matz-e:
Remove the "flexibility" of finding dependencies at use time. This
simplifies our CMake code greatly.

### DIFF
--- a/CMake/HighFiveConfigFixed.cmake.in
+++ b/CMake/HighFiveConfigFixed.cmake.in
@@ -1,0 +1,4 @@
+@PACKAGE_INIT@
+
+include("${CMAKE_CURRENT_LIST_DIR}/HighFiveTargets.cmake")
+check_required_components("@PROJECT_NAME@")
--- a/CMake/HighFiveConfigFlexible.cmake.in
+++ b/CMake/HighFiveConfigFlexible.cmake.in
@@ -1,26 +1,21 @@
-function(copy_interface_properties target source)
-  foreach(prop
-          INTERFACE_COMPILE_DEFINITIONS
-          INTERFACE_COMPILE_FEATURES
-          INTERFACE_COMPILE_OPTIONS
-          INTERFACE_INCLUDE_DIRECTORIES
-          INTERFACE_LINK_LIBRARIES
-          INTERFACE_SOURCES
-          INTERFACE_SYSTEM_INCLUDE_DIRECTORIES)
-    set_property(TARGET ${target} APPEND PROPERTY ${prop} $<TARGET_PROPERTY:${source},${prop}>)
-  endforeach()
-endfunction()
-
 if(TARGET HighFive)
     return()
 endif()
 
-# Get HighFive targets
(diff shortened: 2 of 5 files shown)


tdegeus:
I've made this suggestion already before, but I still think it is worth to consider. I would propose to have a few targets: 
```
HighFive::include (headers only)
HighFive::HighFive (headers + HDF5)
HighFive::Boost (enable Boost, add relevant deps to target)
HighFive::xtensor (,, xtensor ,,,)
...
```
(In addition to the current target that remains the same for backward compatibility, but that is constructed by `HighFive::HighFive + HighFive::Boost`)

That way the user could do e.g.
```cmake
find_package(HighFive required)
target_link_libraries(mytarget HighFive::HighFive HighFive::xtensor)
```
or even 
```cmake
find_package(HighFive required)
find_package(hdf5 required)
target_link_libraries(mytarget HighFive::include ...my_custom_hdf5_target)
```
It would have the same simplification and in addition give a lot of flexibility all while cleaning up the user's `CMakeLists.txt`

1uc:
The context of this Draft PR is figuring out an internal build failure. It's not to rewrite or restructure our CMake code.

Reworking our CMake code is needed. I'm not sure we need all the targets you propose. One perk of header-only is, we don't compile anything; and we also don't have what CMake calls "private" dependencies, i.e. if a user needs HighFive with XTensor support, they must have XTensor objects. Hence they've set everything to be able to include and link with XTensor, or Eigen, or Boost, or ...

Hence I hope we could get to a point where:
```
#include <highfive/highfive.hpp>
#include <highfive/eigen.hpp>
```
is all that's needed to get Eigen support, i.e. a single "library" supports all optional dependencies simply by `#include`ing the "addons" that are needed (if and only if they're required).

I like the `HighFive::include` suggestion especially if it would allow us to turn off any HDF5 detection related code.

However, all of this is a topic for a different PR, maybe the one you made, or a different one.

1uc:
Let's not do any of this; and save changes for when we're willing to actually fix our CMake.

matz-e:
As @1uc says… we need a revamp. And this does not solve the core issue that `HDF5_LIBRARIES` sometimes points at targets, other times at libraries.

@tdegeus I've had a look at your old PR again, and I think I misunderstood some of the concepts in there before. I'm sorry! In general, I think it seems to be a better approach (@1uc helped convince me), one thing I'm wondering: Rather than create all kinds of different targets, would it be worth considering using CMake components?

matz-e:
I agree on the target part. I mentally tuned out that HighFive still used the older variables :see_no_evil:

Regarding the components, it seems that they could be used to make the targets available. As in:
```
find_package(highfive)  # no components in HighFive
```
and exposing `highfive::xtensor` would always have to do a `find_dependency(xtensor)`, but with targets that could be made optional, and kind of avoid all these badly namespaced global variables?

```
find_package(highfive COMPONENTS hdf5 xtensor)
```
would basically be equivalent to set `USE_XTENSOR=ON USE_BOOST=OFF`. I would have to see more about how all this normally is implemented, though.

PR #715, opened 2023-04-03
Fix version string

ischoegl:
**Description**

The version number generated from `H5Version.hpp.in` is not a string, but a corrupted floating point number

- [x] Fix version string in `H5Version.hpp`
- [x] Remove obsolete `VERSION` file referencing 2.2.2
- [x] Added an import of `H5Version.hpp` to `tests_high_five_base.cpp`

Without the quotes, including `H5Version.hpp` triggers errors similar to
```
error: invalid suffix '.0' on floating constant
```

**How to test this?**

```bash
cmake .. 
make
make test
```

**Test System**
 - OS: macOS Ventura 13.3
 - Compiler: Apple clang version 14.0.3
 - Dependency versions: hdf5 1.12


1uc:
Thank you for preparing an MR, we'll look into it and fix what looks like build-system issues.

1uc:
Probably this is missing the equivalent of `${CMAKE_CURRENT_BINARY_DIR}/include`:
```
target_include_directories(libheaders INTERFACE
  "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>"
  "$<INSTALL_INTERFACE:include>")
```
https://github.com/BlueBrain/HighFive/blob/559bd8565f001984013a0c7a215ef19503855884/CMake/HighFiveTargetExport.cmake#L5

1uc:
Merging this into `1uc/fix-version-string` to be able to fix the uncovered build issues.

Issue #716, opened 2023-04-03
Logging uses enum `DEBUG` that is a commonly used C/C++ define


emmenlau:
**Describe the bug**
The logging implemented in https://github.com/BlueBrain/HighFive/pull/690 is very nice. But there is a problem with one of the enum values:
https://github.com/BlueBrain/HighFive/blob/0fc71b66883b405cacb281ce79e05f6fd1cb7f3a/include/highfive/H5Utility.hpp#L49

The enum holds a value `DEBUG`. However this keyword conflicts with compiler defines `DEBUG`. And this define is very very common, i.e. on Microsoft Windows platforms where `_DEBUG` and `DEBUG` are quite common defines for debug builds.

Actually, all the log level names may be prone to name conflicts, since also `INFO`, `WARN` and `ERROR` are quite common names. Would it be possible to change the names and add a prefix or suffix, to make them more unique? Something like:
```c++
enum class LogSeverity {
    LVL_DEBUG = HIGHFIVE_LOG_LEVEL_DEBUG,
    LVL_INFO = HIGHFIVE_LOG_LEVEL_INFO,
    LVL_WARN = HIGHFIVE_LOG_LEVEL_WARN,
    LVL_ERROR = HIGHFIVE_LOG_LEVEL_ERROR
};
```

1uc:
Right! Wouldn't lower casing them work, since macros are expected to be all caps?

PR #717, opened 2023-04-03
Avoid name collision with macros in LogSeverity.

1uc:
See  #716.


matz-e (approved):

emmenlau:
lgtm, thanks for the quick response!

PR #719, opened 2023-04-03
Fix `HIGHFIVE_VERSION`.

1uc:
This fixes two issues:
* `HIGHFIVE_VERSION` should be a string literal. (Used to be something invalid.)
* The build system didn't include the `*_BINARY_DIR/include` for local builds. The generated file was installed.

See #715 for the initial contribution and more information.


1uc:
Double check that the file did get installed:
```
-- Installing: [...]/include/highfive/H5Version.hpp
```

matz-e (approved):

Issue #721, opened 2023-04-03
Can not build against HighFive on MSVC 2022 v17.5 due to `details::Boolean::TRUE`


emmenlau:
**Describe the bug**

I fail to build against current latest released HighFive v2.7.0 on Windows using MSVC 2022 v17.5, due to a compiler error:
```
[build] In file included from C:\data\Debug\include\highfive/H5DataSet.hpp:14:
[build] In file included from C:\data\Debug\include\highfive/H5DataType.hpp:412:
[build] C:\data\Debug\include\highfive/bits/H5DataType_misc.hpp(210,76): error: expected unqualified-id
[build]     return {{"FALSE", details::Boolean::FALSE}, {"TRUE", details::Boolean::TRUE}};
[build]                                                                            ^
[build] C:\Program Files (x86)\Windows Kits\10\\include\10.0.22621.0\\shared\minwindef.h(75,29): note: expanded from macro 'TRUE'
[build] #define TRUE                1
[build]                             ^
```

It seems Windows brings the define `TRUE` in header `minwindef.h`, which breaks the boolean definition of HighFive. Would it be possible to kindly rename these booleans?

I've tested that `True` and `False` would work instead.

emmenlau:
Thanks a lot to you, for the quick response!

Sadly, we also don't have anything that helps test for these kind of problems in a bit more concise way. I was just "lucky" to hit these two problems.

The only thing I can recommend is to have tests that include `Windows.h`, without defining any of the "reducer-macros" like `NOMINMAX` or `WIN32_LEAN_AND_MEAN`. Including the Windows header is sufficient to get a gazillion defines, class and function names, mostly in global namespace.

emmenlau:
Thanks to you for the great support!!

alkino:
Fixed by #722

PR #722, opened 2023-04-03
Avoid name collision with macros `TRUE` and `FALSE`.

1uc:
See #721.


1uc:
While `details::Boolean::HighFiveFalse` seems excessive, `False` looks to much like the Python name for `true`. We're free to rename it as it's not part of the API, so the only concern is to avoid macros interfering.

matz-e (approved):

PR #723, opened 2023-04-03
Include `Windows.h` to uncover more defects on Windows.

1uc:
The purpose is to make our CI more sensitive to name clashes of macros with tokens we've used in HighFive. The idea is that by including `Windows.h` we should be pulling in many standard macros. Hence we should be able to detect the defect.


1uc:
This requires #725 and #722 to pass.

matz-e (approved):

Issue #724, opened 2023-04-03
Build issue on certain MSVC in Cantera.


1uc:
The CI of https://github.com/Cantera uncovered a build issue for certain versions of MSVC. Some of the discussion started in #713. 

The link to the failing CI is:
https://github.com/Cantera/cantera/actions/runs/4588940048/jobs/8103492434

It reports a
```
c:\miniconda3\envs\test\library\include\highfive\bits/H5Path_traits_misc.hpp(31): error C2248: 'HighFive::Object::Object': cannot access protected member declared in class 'HighFive::Object'
...
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.16.27023\include\filesystem(2392): note: see reference to class template instantiation 'std::chrono::time_point<std::filesystem::_File_time_clock,std::filesystem::_File_time_clock::duration>' being compiled
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.16.27023\include\type_traits(616): note: see reference to class template instantiation 'std::basic_string_view<wchar_t,std::char_traits<wchar_t>>' being compiled
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.16.27023\include\xstring(2124): note: see reference to class template instantiation 'std::is_convertible<const _StringViewIsh &,std::basic_string_view<wchar_t,std::char_traits<wchar_t>>>' being compiled
```
Which smell like SFINAE to check if something HighFive related can be converted to a string. While doing so it fails because the ctor of `HighFive::Object` is protected. To be investigated further.

ischoegl:
Thanks, @1uc ... trying to debug a couple of other quirks, I updated/fixed our build scripts (unfortunately we're using a different build system - `SCons` rather than `cmake`) and I now can follow installation details a little better. The offending issue is now bypassed, as our installation now falls back to an earlier git submodule 2.6.2. 

The latest runner we're using does have a system installation of 2.7.0 from conda, which produces the same output as above for MSVC 14.1:
```
c:\miniconda3\envs\test\library\include\highfive\bits/H5Path_traits_misc.hpp(31): error C2248: 'HighFive::Object::Object': cannot access protected member declared in class 'HighFive::Object'
c:\miniconda3\envs\test\library\include\highfive\bits/H5Object_misc.hpp(27): note: see declaration of 'HighFive::Object::Object'
[...]

PR #727, opened 2023-04-04
Fix friends on windows.

1uc:
Reintroduces friend declarations, which are the most likely cause of the regression seen in #724. Since these declaration cause issues on older versions of GCC, they're behind the include guard `HIGHFIVE_HAS_FRIENDS=1`. For HighFive code the friend declarations are not needed.

### DIFF
--- a/CMake/HighFiveTargetDeps.cmake
+++ b/CMake/HighFiveTargetDeps.cmake
@@ -14,6 +14,10 @@ if(NOT TARGET libdeps)
     target_compile_definitions(libdeps INTERFACE -D_GLIBCXX_ASSERTIONS)
   endif()
 
+  if(HIGHFIVE_HAS_FRIENDS)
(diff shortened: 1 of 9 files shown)


1uc:
We've tried to build our examples with the toolset `v141`, i.e. `cmake -T v141` which should be the version that's failing in Cantera's CI. The issue is, it doesn't cause the build to fail, even without the fix in this PR. However, I'm not sure I've accurately reproduced the same combination of compiler and toolset.

@ischoegl If you have time, could you please try if this commit fixes the issues?

ischoegl:
@1uc … thank you, I appreciate it! While getting our CI to pull from that branch is a little tricky, I should be able to check on this tomorrow with a windows machine. Comparing configurations is not straightforward as we’re using SCons, rather than cmake. The CI is set up for VisualStudio 2022 (14.3), with toolset set to 14.1.

1uc:
Closed in favour of #728.

PR #728, opened 2023-04-04
Avoid MSVC compiler bug.

1uc:
The purpose of this MR is to get older versions of MSVC to compile HighFive. The hope is it'll reproduce the issue in #724. This fixes required are also included in this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -260,8 +260,13 @@ jobs:
 
   # Job testing in Windows
   # ======================
-  Windows:
-    runs-on: Windows-latest
+  Windows-2022:
+    runs-on:  ${{matrix.os}}
+    strategy:
+      matrix:
+        os: [ "windows-2022"]
+        vs-toolset: [ "v141", "v143" ]
+        cxxstd: ["14", "17"]
     steps:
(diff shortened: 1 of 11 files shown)


1uc:
We can compare two versions: `v141` and `v143`. They report the same version of visual studio:
```
-- Building for: Visual Studio 17 2022
```
but different compiler versions:
```
-- The CXX compiler identification is MSVC 19.16.27049.0
-- The CXX compiler identification is MSVC 19.34.31943.0
```

Checking: https://blog.knatten.org/2022/08/26/microsoft-c-versions-explained/
Suggests that those might be the compilers corresponding to the toolsets 141 and 143, respectively.

Comparing to path information found in Cantera CI we see:
```
Compiling with MSVC version 14.3
Compiling with MSVC toolset 14.1
```
https://github.com/Cantera/cantera/actions/runs/4588940048/jobs/8103492434#step:5:12

Which is probably consistent since the knatten's blog suggests that Visual Studio 2022 has MSVC versions `14.30` or `14.31`.

1uc:
Should be merged after #729.

pramodk (approved):
LGTM

Issue #532, opened 2022-03-31
Writing `Eigen::MatrixXd` as column major.


1uc:
I've created a PR for adding an example which writes an `Eigen::MatrixXd`, see #531. It's supposed to create a matrix
```
  0    1    2
100  101  102
...
900  901  902
```
This matrix is then written to file. When dumping the resulting file I get
```
HDF5 "eigen_matrix_example.h5" {
GROUP "/" {
   DATASET "dset" {
      DATATYPE  H5T_IEEE_F64LE
      DATASPACE  SIMPLE { ( 10, 3 ) / ( 10, 3 ) }
      DATA {
      (0,0):   0,   100, 200,
      (1,0): 300, 400, 500,
      (2,0): 600, 700, 800,
      (3,0): 900,   1, 101,
      (4,0): 201, 301, 401,
      (5,0): 501, 601, 701,
      (6,0): 801, 901,   2,
      (7,0): 102, 202, 302,
      (8,0): 402, 502, 602,
      (9,0): 702, 802, 902
      }
   }
}
}
```
Which is consistent with HDF5 assuming data is row-major (C-style) and the actual Eigen matrix being column-major (Fortran/BLAS-style). I couldn't immediately find any code that transposes the matrix before writing it. 

Note that a write-read cycle using HighFive returns, as expected, the same matrix.

Can someone confirm that I've set up the example properly and maybe comment on the desired behaviour?

tdegeus:
Thanks for pointing this out!
I would say that this is not desired. As for as I know HDF5 in row-major and the data should be presented as such. Writing column-major and row-major and then reinterpreting it falsely to correct for it does not make any sense to me. What if you were to read it in with let's say `h5py`? Then everything would be mixed up!

What I do in H5Easy is to cast to row-major before writing https://github.com/BlueBrain/HighFive/blob/34bbc607182b8eab3f5dee539e428cfe403542f6/include/highfive/h5easy_bits/H5Easy_Eigen.hpp#L89 I think that the core should do this too

matz-e:
Casting it like that seems like a good idea (element of least surprise)! Should this be unified / lifted to "core" to avoid duplication? And would it be wise to make it an option to throw an error / assert if a type conversion happens?

Thinking of bigger matrices that one wouldn't really want to store in memory multiple times.

tdegeus:
[...]

PR #731, opened 2023-04-19
Remove Eigen functionality for matrices.

1uc:
HighFive hasn't been respecting that Eigen uses Fortran order, as a result it's been writing `Eigen::Matrix` to file as if the data were C order. It's been consistently reading datasets back into `Eigen::Matrix` has if it already was in Fortran order.

This error is symmetric, i.e., if one writes an `Eigen::Matrix` to disk via core HighFive and then reads that dataset back into an `Eigen::Matrix` the outcome is correct (which has been check in CI).

More information can be found at:
https://github.com/BlueBrain/HighFive/issues/532


include/highfive/bits/H5Converter_misc.hpp
```diff
@@ -625,7 +625,18 @@ struct inspector<Eigen::Matrix<T, M, N>> {
     static constexpr bool is_trivially_copyable = std::is_trivially_copyable<value_type>::value &&
                                                   inspector<value_type>::is_trivially_copyable;
 
+
+    static void assert_not_buggy(Eigen::Index nrows, Eigen::Index ncols) {
+        if (nrows > 1 && ncols > 1) {
+            throw std::runtime_error(
+                "HighFive has been broken for Eigen::Matrix. Please check "
+                "https://github.com/BlueBrain/HighFive/issues/532.");
```
tdegeus, line 633:
```suggestion
                "HighFive has been broken for Eigen::Matrix. "
                "Please use H5Easy or open a PR. See "
                "https://github.com/BlueBrain/HighFive/issues/532.");
```

src/examples/eigen_matrix.cpp
```diff
@@ -1,50 +0,0 @@
...
-const std::string FILE_NAME("eigen_matrix_example.h5");
-const std::string DATASET_NAME("dset");
-const int nrows = 10;
-const int ncols = 3;
-
-// Create a 2D dataset 10x3 of double with eigen matrix
-// and write it to a file
-int main(void) {
-    try {
-        Eigen::MatrixXd matrix(nrows, ncols);
```
tdegeus, line 28: In stead of removing, you could use a row-major example for the moment. We can then revert one someone had the time to fix HighFive

doc/poster/example_eigen.cpp
```diff
@@ -1,35 +0,0 @@
...
-
-using namespace HighFive;
-
-void data_io(void) {
-    const std::string DATASET_NAME("dset");
-    const int nrows = 10;
-    const int ncols = 3;
-
-    try {
-        Eigen::MatrixXd mat(nrows, ncols);
```
tdegeus, line 16: See other example

doc/poster/example_eigen.cpp
```diff
@@ -1,35 +0,0 @@
...
-            }
-        }
-
-        File file("eigen_mat.h5", File::ReadWrite | File::Create | File::Truncate);
-
-        DataSet dset = file.createDataSet(DATASET_NAME, mat);
-        dset.write(mat);
-
-        Eigen::MatrixXd result;
-        dset.read(result);
```
tdegeus, line 30: It would always be good to add an assertion

include/highfive/bits/H5Converter_misc.hpp
1uc: It's unfortunately not that simple. Just switching to H5Easy isn't going to fix the issue of existing files. The point of breaking this is to make people check the ticket so that they understand how they're affected by this issue.

src/examples/eigen_matrix.cpp
1uc: No, `git revert` works quite well. I don't want there to be anything suggesting that reading or writing `Eigen::Matrix` from core HighFive is sensible.

include/highfive/bits/H5Converter_misc.hpp
tdegeus: You could say "Please use H5Easy to write NEW data. Note that old data may be 'corrupted'"

1uc: I'm not sure I understand what you want to achieve by suggestion H5Easy as an alternative, it's not a drop in replacement (Vectors serialized from core are 2D; while Easy strips the singleton dimension).

I would like the message to read that they need to go read a full length piece of text; and then figure out how this affects them.

The point is to end the indecision.

tdegeus: It's fine, you can keep what you had.

1uc:
@tdegeus Thank you for  your time and comments!

PR #733, opened 2023-04-20
Fix thread-safety statement.

1uc:


1uc:
@pramodk after merging the PR yesterday, I noticed that while the new sentence sounds good. It's likely not true. A user has reported that HighFive isn't thread-safe and provided code comparing HDF5 and HighFive. Given that some HighFive object modify themselves, I thinks plausible that HighFive isn't thread-safe even when using the thread-safe HDF5 library.

pramodk:
Ah ok! Just to have a reference in this ticket, could you add a link to relevant issue / comment?

1uc:
I think I've seen others, but this one feel credible: https://github.com/BlueBrain/HighFive/discussions/675.

pramodk (approved):

PR #735, opened 2023-04-26
Add wrapper for commonly used 'H5T*' functions.

1uc:
This add wrappers that perform the required error checking for commonly used `H5T*` functions. The wrappers are intentionally kept internal.

PR #736, opened 2023-04-26
Make codecov CI more permissive.

1uc:
Code coverage in HighFive should not determine the success or failure of CI. It might be useful to notice if CI essentially doesn't cover anything anymore. Hence the threshold of 50%.

The setting 'fail_ci_if_error' will make the CI fail if there are issues uploading results. Hence, this will just make the CI fail spuriously, e.g. if codecov servers are down. The default is false, we should stick with it.


1uc:
The concrete issue we want to avoid is #735 where CI fails due -0.01% converage.

alkino (approved):

PR #740, opened 2023-05-01
Allow user-specified `mem_space` for hyperslabs.

1uc:
This commit allows users to specify the memory dataspace. Previously, we forced the memspace to be 1D, because in general a hyperslab selection isn't a multi-dimensional block anymore.

We now allow users to specify the memory dataspace, which must be simple; but can be multi-dimensional.


pramodk (approved):
Looks straightforward change!

PR #744, opened 2023-05-05
Flexible `std::string`s.

1uc:
This PR outlines how one could read and write `std::string` and containers of `std::string` into/from fixed or variable length HDF5 strings.


1uc:
Things to discuss:
* Currently the new code defaults to ASCII while old code default to UFT8. Probably, it would make more sense to default to UFT8 everywhere.
* There's no default for the padding of fixed-length strings. The issue is that the natural choice null-terminated could easily lead to writing strings to disk that claim they're null-terminated, but they're not. The fact that they're not null-terminated isn't the issue, we know their length. The problem is only that they claim that they are; when they're not.

matz-e (approved):
Had a glance over this, looks good.

alkino (approved):

PR #753, opened 2023-05-12
Try last version 1.14.1

alkino:


alkino:
The bug seems to be https://github.com/HDFGroup/hdf5/issues/1632

tests/unit/tests_high_five_base.cpp
```diff
@@ -291,7 +291,7 @@ TEST_CASE("Test group properties") {
     File file(file_name, File::Truncate, fapl);
 
     GroupCreateProps props;
-    props.add(EstimatedLinkInfo(1000, 500));
+    props.add(EstimatedLinkInfo(10, 60));
```
matz-e, line 294: We're giving up on this?

1uc: I reread the all the messages. One of the very helpful things @matz-e wrote was that this is in debug mode. If we look closely what fails is an `HDassert`. However, it turns out that this is just a debugging help, i.e.,
https://github.com/HDFGroup/hdf5/blob/577c192518598c7e2945683655feffcdbdf5a91b/src/H5private.h#L601

I've updated the CI to build Release versions of HDF5; and the issue went away.

1uc (approved):
Thank you!

tests/unit/tests_high_five_base.cpp
matz-e: After some deliberation, I think as long as upstream is dragging their feet, we should insist on this. So I'm fine with the change, luckily we don't depend on this feature 😅

matz-e (approved):

Issue #638, opened 2022-11-07
Build coverage for pHDF5


1uc:
For sequential HDF5 we check that HDF5 can be build and linked properly. One of the important errors this find are missing `inline`. However, in #637 these checks failed to uncover such a case. It would be pleasant to close this gap, since the issues is likely to reoccur.

alkino:
Today in https://github.com/BlueBrain/HighFive-testing only serial tests are run.
We could duplicate compilation to add parallel tests.

PR #760, opened 2023-05-26
Also run project integration CI for pHDF5.

1uc:
This aim is to prevent build failures, e.g. missing `inline`, when using a parallel version of HDF5. In HighFive there's some code that's only compiled with the parallel version of HDF5. See #638.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -192,15 +192,32 @@ jobs:
   # =====================================================
   CMake_Project:
     runs-on: ubuntu-20.04
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


1uc:
This:
https://github.com/BlueBrain/HighFive/actions/runs/5089009161/jobs/9146119240?pr=760

confirms that the parallel version is now also compiled and picks up MPI.

1uc:
This:
https://github.com/BlueBrain/HighFive/actions/runs/5089009161/jobs/9146119240?pr=760

confirms that the parallel version is now also compiled and picks up MPI.

pramodk (approved):

PR #761, opened 2023-05-26
Make ci.yml DRYer.

1uc:
We use `include:` to run all combinations of Windows, even if they cant' exactly be described by a matrix.


alkino (approved):
I checked that CI was running what it told us

PR #762, opened 2023-05-26
added ability to create hard links

Quark-X10:
**Description**

- added a method in NodeTraits to create hard links

**How to test this?**

```bash
cmake ..
make -j8
make test
```

**Test System**
 - OS: Ubuntu 22.04.2 LTS (WSL)
 - Compiler: gcc 11.3.0

### DIFF
--- a/include/highfive/bits/H5Node_traits_misc.hpp
+++ b/include/highfive/bits/H5Node_traits_misc.hpp
@@ -358,6 +358,29 @@ inline void NodeTraits<Derivate>::createExternalLink(const std::string& link_nam
     }
 }
 
+template <typename Derivate>
+template <typename T, typename>
+inline void NodeTraits<Derivate>::createHardLink(const std::string& link_name,
+                                                 const T& target_obj,
+                                                 LinkCreateProps linkCreateProps,
+                                                 const LinkAccessProps& linkAccessProps,
+                                                 const bool parents) {
+    static_assert(!std::is_same<T, Attribute>::value,
+                    "hdf5 doesn't support hard links to Attributes");
+    if (parents) {
+        linkCreateProps.add(CreateIntermediateGroup{});
+    }
+    auto status = H5Lcreate_hard(target_obj.getId(),
+                                ".",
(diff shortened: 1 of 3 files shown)


alkino:
Should fix #357 ?

Quark-X10:
I think it should fix #357
I let a typo slip into my PR, sorry for that. Actually this is my first PR ever.

1uc:
>  Actually this is my first PR ever.

In that case: Well done! It includes a sensible test which is very good. The only things I'd like to change is the description, since I never know which way links go. We'll fix the formatting issue. Clang-format is a bit particular about using exactly the correct version or otherwise the output differs.

Quark-X10:
Thanks !
> We'll fix the formatting issue.

I committed the formatting change before seeing your comment sorry.

> The only things I'd like to change is the description, since I never know which way links go. 

I agree with you, I'm always confused about links direction, if "target object" and "link name" are not clear enough, feel free to use a better wording !

include/highfive/bits/H5Node_traits_misc.hpp
```diff
@@ -358,6 +358,29 @@ inline void NodeTraits<Derivate>::createExternalLink(const std::string& link_nam
     }
 }
 
+template <typename Derivate>
+template <typename T, typename>
+inline void NodeTraits<Derivate>::createHardLink(const std::string& link_name,
+                                                 const T& target_obj,
+                                                 LinkCreateProps linkCreateProps,
```
alkino, line 365:
```suggestion
                                                 const LinkCreateProps& linkCreateProps,
```

Quark-X10: As in createSoftLink and createExternalLink, if parents is true, the link create properties have to be modified.
I tried to maintain consistent API with these 2 methods.

alkino: thank you for explanation :)

1uc:
Thank you, this looks good. I'll merge it into a branch `hard-links`, add a few additional tests. Mostly to check if I understand it correctly.

PR #764, opened 2023-05-26
File size utilities

Quark-X10:
**Description**

Two methods are added to the File interface : one to get the file disk size through H5Fget_filesize and the currently tracked unused space through H5Fget_freespace.
These can be useful when using a persistant unused space tracking strategy to decide when to trigger repacks after data multiple addition and deletion of data.

Unit tests have been written, however it may be irrelevant since the disk size may vary depending on the hdf5 library version.

**How to test this?**

```bash
cmake ..
make -j8
make test
```

**Test System**
 - OS: Ubuntu 22.04.2 LTS (WSL)
 - Compiler: gcc 11.3.0

### DIFF
--- a/tests/unit/tests_high_five_base.cpp
+++ b/tests/unit/tests_high_five_base.cpp
@@ -444,8 +444,8 @@ TEST_CASE("FileSizeAndUnusedSpace") {
     }
 
     {
-	FileCreateProps fcp;
-	fcp.add(FileSpaceStrategy(H5F_FSPACE_STRATEGY_FSM_AGGR, true, 0));
+        FileCreateProps fcp;
+        fcp.add(FileSpaceStrategy(H5F_FSPACE_STRATEGY_FSM_AGGR, true, 0));
         File file(file_name_tracked, File::ReadWrite | File::Create | File::Truncate, fcp);
         auto dset = file.createDataSet(ds_path, data);
     }
@@ -473,7 +473,6 @@ TEST_CASE("FileSizeAndUnusedSpace") {
         CHECK(file.getDiskSize() == tracked_file_size);
         CHECK(file.getUnusedSpace() == tracked_unused_space);
     }
-
 }
 #endif
 


include/highfive/H5File.hpp
```diff
@@ -110,6 +110,12 @@ class File: public Object, public NodeTraits<File>, public AnnotateTraits<File>
         return details::get_plist<FileAccessProps>(*this, H5Fget_access_plist);
     }
 
+    /// \brief Get the disk size of this file in bytes
+    size_t getDiskSize() const;
```
1uc, line 114: This is a wrapper for the function `H5Fget_filesize`. Could we use `getFileSize` instead?

tests/unit/tests_high_five_base.cpp
```diff
@@ -426,6 +426,56 @@ TEST_CASE("Test groups and datasets") {
...
+    const std::string ds_path("/dataset");
+    const std::vector<int> data{13, 24, 36};
+    const size_t untracked_file_size = 2048;
+    const size_t untracked_unused_space = 600;
+    const size_t tracked_file_size = 1158;
+    const size_t tracked_unused_space = 103;
+
+
+    {
+        File file(file_name_untracked, File::ReadWrite | File::Create | File::Truncate);
```
1uc, line 442: FYI this can be reduced down to:
```
File file(file_name_untracked, File::Truncate);
```
It's consistent with the rest of the file so there's no reason to change it.

tests/unit/tests_high_five_base.cpp
```diff
@@ -426,6 +426,56 @@ TEST_CASE("Test groups and datasets") {
     }
 }
 
+#if H5_VERSION_GE(1, 10, 1)
+TEST_CASE("FileSizeAndUnusedSpace") {
```
1uc, line 430: Please split the test in two. One for the untracked file and one for the tracked variant.

tests/unit/tests_high_five_base.cpp
```diff
@@ -426,6 +426,56 @@ TEST_CASE("Test groups and datasets") {
     }
 }
 
+#if H5_VERSION_GE(1, 10, 1)
```
1uc, line 429: The reason it's needed it so set the file space strategy. The newly wrapped functions were introduced in HDF5 1.6.x and don't need guards.

tests/unit/tests_high_five_base.cpp
```diff
@@ -426,6 +426,56 @@ TEST_CASE("Test groups and datasets") {
...
+    {
+        File file(file_name_tracked, File::ReadWrite);
+        file.unlink(ds_path);
+    }
+
+    {
+        File file(file_name_tracked, File::ReadWrite);
+        CHECK(file.getDiskSize() == tracked_file_size);
+        CHECK(file.getUnusedSpace() == tracked_unused_space);
+    }
```
1uc, line 475: As you already said, these will fail. Since we're not testing if HDF5 is correct, it might be enough to check that:
* the file size is `>0`.
* the free space is ` == 0` (when applicable).
* the free space is ` > 0` and ` < file_size` (when applicable).

1uc (commented):
Very nice, thank you! There's a few things I'd suggest changing. If you can take care of them that would be appreciated, if not we can do it.

1uc:
Thank you for the PR, I'll merge it into a feature branch and apply the changes I mentioned.

PR #780, opened 2023-06-27
Include `H5Version.hpp`.

1uc:


1uc:
The point is that `H5Version.hpp` is a generated files and resides in a different directory from all other files. Hence, this checks if that directory is included in the CMake target `HighFive`.

pramodk (approved):

PR #781, opened 2023-06-29
Run examples during CI.

1uc:


.github/workflows/ci.yml
```diff
@@ -107,13 +107,20 @@ jobs:
...
     - name: Build
       run: |
-        mkdir BUILD && cd BUILD
-        cmake -GNinja -DHDF5_ROOT=$HOME/${{ matrix.hdf5_version }} ..
-        ninja
+        CMAKE_OPTIONS=(
+          -GNinja
+          -DHDF5_ROOT=$HOME/${{ matrix.hdf5_version }}
+        )
+        source $GITHUB_WORKSPACE/.github/build.sh
```
matz-e, line 114: Why is this sourced? Does it set some environment variables?

1uc: All other locations source it... I think mostly because it sets `HIGHFIVE_BUILD` which would be the "proper" way to know the build directory, i.e. it should probably be `ctest --test-dir ${HIGHFIVE_BUILD} ... `; but we don't use and instead do:
```
    - name: Test
      working-directory: ${{github.workspace}}/build
      run: ctest --output-on-failure -C $BUILD_TYPE
```

matz-e: Hm, little iffy if that variable isn't used later. I keep having to tell people to not execute shell scripts with `.` or `source` and then wondering why their terminals close (due to `set -e`) or having a messed up environment...

I'll approve for consistency.

matz-e (approved):

PR #788, opened 2023-06-30
Cleanup of CI code.

1uc:
The `source` isn't needed, is slightly unexpected and more error prone than just running the script.


1uc:
Removing the `sources` makes various pipelines fail.

PR #794, opened 2023-07-10
Move `StringPadding` to own file.

1uc:


1uc:
This avoids a vicious include cycle in later parts of #744.

PR #795, opened 2023-07-10
Introduce '{Shallow,Deep}CopyBuffer'.

1uc:
This commit introduces the internal notion of buffers. We already need to copy non-contiguous data into a contiguous buffer. This commit simply makes this idea more explicit.

The advantage is that each class can focus on its purpose and thereby simplify its code.

### DIFF
--- a/include/highfive/bits/H5Converter_misc.hpp
+++ b/include/highfive/bits/H5Converter_misc.hpp
@@ -8,91 +8,132 @@
  */
 #pragma once
 
+#include <type_traits>
+
 #include "H5Inspector_misc.hpp"
+#include "../H5DataType.hpp"
 
 namespace HighFive {
 namespace details {
 
-template <typename T>
-struct Writer {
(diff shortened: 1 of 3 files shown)


alkino (approved):
really nice addition.

include/highfive/bits/H5Converter_misc.hpp
```diff
@@ -8,91 +8,132 @@
...
+        inspector<T>::serialize(val, this->get_pointer());
     }
+};
 
+
+template <typename T, typename Enable = void>
+struct Reader;
+
+template <typename T>
+struct Reader<T, typename enable_shallow_copy<T>::type>: ShallowCopyBuffer<T, false> {
```
alkino, line 103:
```suggestion
struct Reader<T, typename enable_shallow_copy<T>::type>: public ShallowCopyBuffer<T, false> {
```

alkino (dismissed):

alkino (approved):

PR #796, opened 2023-07-10
Read datatype

Quark-X10:
**Description**

HighFive provides the ability to commit datatypes to files with CompoundType::commit for example but does not provide the ability to open the previously commited datatypes.
This PR enhance the Node Traits to get the datatype stored in that node.

**How to test this?**

```bash
cmake ..
make -j8
make test
```

**Test System**
-OS: Ubuntu 22.04.2 LTS (WSL)
-Compiler: gcc 11.3.0
-Dependency versions: hdf5 1.10.7

### DIFF
--- a/tests/unit/tests_high_five_base.cpp
+++ b/tests/unit/tests_high_five_base.cpp
@@ -2870,16 +2870,22 @@ TEST_CASE("HighFiveEnum") {
 TEST_CASE("HighFiveReadType") {
     const std::string file_name("readtype_test.h5");
     const std::string datatype_name1("my_type");
+    const std::string datatype_name2("position");
 
     File file(file_name, File::ReadWrite | File::Create | File::Truncate);
 
-    auto t3 = AtomicType<int>();
     CompoundType t1 = create_compound_csl1();
     t1.commit(file, datatype_name1);
 
     CompoundType t2 = file.getDataType(datatype_name1);
 
+    auto t3 = create_enum_position();
+    t3.commit(file, datatype_name2);
+
+    DataType t4 = file.getDataType(datatype_name2);
+
     CHECK(t2 == t1);
+    CHECK(t4 == t3);
 }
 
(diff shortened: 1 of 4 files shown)


1uc:
Thank you, this looks interesting and I'll have a closer look later. In the meantime:

```
    template <typename Derivate>
    friend class NodeTraits;
```
is unfortunately a something that's a bit plagues by compiler errors, some compilers requires this others insist that it's not present. Please see,
https://github.com/BlueBrain/HighFive/blob/master/include/highfive/H5Selection.hpp#L61-L64

for example and if CI complains, please adjust accordingly.

Quark-X10:
Thank you for your quick reply.
I thought the protection was legacy since this type of declaration is unprotected elsewhere like here: https://github.com/BlueBrain/HighFive/blob/master/include/highfive/H5DataSet.hpp#L112-L113
CI does not seems to complain about that but i can correct if needed.
Moreover, I have some trouble to understand why the CI is complaining about a memory leak.

tests/unit/tests_high_five_base.cpp
```diff
@@ -2867,6 +2867,27 @@ TEST_CASE("HighFiveEnum") {
...
+    const std::string file_name("readtype_test.h5");
+    const std::string datatype_name1("my_type");
+    const std::string datatype_name2("position");
+
+    File file(file_name, File::ReadWrite | File::Create | File::Truncate);
+
+    CompoundType t1 = create_compound_csl1();
+    t1.commit(file, datatype_name1);
+
+    CompoundType t2 = file.getDataType(datatype_name1);
```
alkino, line 2880: I don't know yet why, but this line create the leak.

alkino:
So the leak is not your responsibility (I think it was only not tested), and this is here: https://github.com/BlueBrain/HighFive/blob/master/include/highfive/H5DataType.hpp#L242
We have to free that.

1uc (approved):
Nice work!

PR #798, opened 2023-07-11
Free return of H5Tget_member_name

alkino:
Tests have been run with and without fix.

### DIFF
--- a/include/highfive/H5DataType.hpp
+++ b/include/highfive/H5DataType.hpp
@@ -244,7 +244,9 @@ class CompoundType: public DataType {
             hid_t member_hid = H5Tget_member_type(_hid, i);
             DataType member_type{member_hid};
             members.emplace_back(std::string(name), member_type, offset);
-            H5free_memory(name);
+            if (H5free_memory(name) < 0) {
+                throw DataTypeException("Could not free names from the compound datatype");
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


include/highfive/H5DataType.hpp
```diff
@@ -239,11 +239,12 @@ class CompoundType: public DataType {
         size_t n_members = static_cast<size_t>(result);
         members.reserve(n_members);
         for (unsigned i = 0; i < n_members; i++) {
-            const char* name = H5Tget_member_name(_hid, i);
+            char* name = H5Tget_member_name(_hid, i);
             size_t offset = H5Tget_member_offset(_hid, i);
             hid_t member_hid = H5Tget_member_type(_hid, i);
             DataType member_type{member_hid};
-            members.emplace_back(name, member_type, offset);
+            members.emplace_back(std::string(name), member_type, offset);
```
alkino, line 246: It was already transform to std::string and so copied. But I prefer to show it, in the same scope.
Any better way?

include/highfive/H5DataType.hpp
```diff
@@ -239,11 +239,12 @@ class CompoundType: public DataType {
...
+            H5free_memory(name);
```
1uc, line 247: Please add the usual `if( ... < 0) { throw ... }` boilerplate for `H5free_memory`. See,
https://docs.hdfgroup.org/hdf5/v1_12/group___h5.html#ga71872bf6445cba956da86d4762b662cf

1uc (approved):

PR #799, opened 2023-07-11
Add doc for HighFive::DataSpace

alkino:


.github/workflows/gh-pages.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       with:
         submodules: 'recursive'
 
-    - uses: mamba-org/provision-with-micromamba@main
+    - uses: mamba-org/setup-micromamba@v1
```
1uc, line 27: This has already been merged, you need to rebase.

alkino: I don't think so: https://github.com/BlueBrain/HighFive/blob/master/.github/workflows/gh-pages.yml#L27

alkino: Only ci.yml has been modified this is gh-pages.yml (for the doc)

1uc: In that case: Thank you! =)

1uc (approved):

PR #805, opened 2023-07-13
Distinguish between buffer pointer and iterator.

1uc:
There's a notion of pointer to the first byte of the buffer. This is what one passes to HDF5, e.g. in `H5Dread`. There's also the concept of a iterator which is used for serialization.

Currently, both ideas are represented by an `hdf5_type*`. However, the StringBuffer will need to return proxy objects. This commits splits the notion. Conveniently, we can rename `get_pointer` to the more in-style `getPointer` and hence the compiler checks that we've considered all occurrences of `get_pointer`.

### DIFF
--- a/include/highfive/bits/H5Converter_misc.hpp
+++ b/include/highfive/bits/H5Converter_misc.hpp
@@ -22,7 +22,6 @@ struct enable_shallow_copy: public std::enable_if<inspector<T>::is_trivially_cop
 template <class T, class V = void>
 struct enable_deep_copy: public std::enable_if<!inspector<T>::is_trivially_copyable, V> {};
 
-
 template <typename T, bool IsReadOnly>
 struct ShallowCopyBuffer {
     using type = unqualified_t<T>;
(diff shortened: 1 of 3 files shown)


include/highfive/bits/H5Converter_misc.hpp
```diff
@@ -57,14 +60,22 @@ struct DeepCopyBuffer {
...
-    hdf5_type const* get_pointer() const {
+    hdf5_type const* getPointer() const {
         return buffer.data();
     }
 
+    hdf5_type* begin() {
+        return getPointer();
+    }
+
+    hdf5_type const* begin() const {
```
matz-e, line 75: Here we get a `const* begin`, but not above? What about `cbegin`?

1uc: It's the same as with say `std::vector`, if the object is const, then so is the iterator. See,
https://en.cppreference.com/w/cpp/container/vector/begin

As for `cbegin`, this is all inside `detail` and we wouldn't need it. Hence, I don't feel it's worth the effort, because it's internal.

matz-e: OK

alkino (approved):

PR #807, opened 2023-07-16
suppress unused warning

luohancfd:
**Description**

Please include a summary of the change and which issue is fixed or which feature is added.

- [ ] Warning of unused variables

**How to test this?**

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce if there is no integration test added with this PR. Please also list any relevant details for your test configuration

```bash
cmake ..
make -j8
make test
```

**Test System**
 - OS: [e.g. Ubuntu 20.04]

### DIFF
--- a/include/highfive/bits/H5Converter_misc.hpp
+++ b/include/highfive/bits/H5Converter_misc.hpp
@@ -13,6 +13,14 @@
 #include "H5Inspector_misc.hpp"
 #include "../H5DataType.hpp"
 
+#ifndef __unused
+#ifdef _MSC_VER
+#define __highfive_unused 
+#else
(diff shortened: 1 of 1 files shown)


1uc:
Thank you for reporting these, I'd prefer suppressing them by simply not giving them a name for the following reasons:

* decent chance it'll work on every (relevant) compiler.
* avoids compiler specific magic.
* avoid the `__highfive_*` which I think is a reserved identifier (`__` anywere).
* closer to what the rest of the codebase does.

I suggest we close in favour of: #808.

Thank again! We might need to look into adding CI which specifically checks for this (`-Werror` during CI only).

luohancfd:
sounds good! closing the issue for now

Issue #810, opened 2023-07-20
Improve productivity with C++20 concepts?


antonysigma:
I like the meta-programming feature built into the HighFive library. Thank you!

I am wondering if I can contribute C++20 code to make the compiler error more readable for average users?

For example, I want the compiler to report the missing `struct Deflate::apply()` function directly, instead of printing 100+ lines to failed recursive template instantiations.

```c++
examples/direct_chunk_write.cpp:98:11: error: no matching member function for call to 'add'
    props.add(Deflate{});
    ~~~~~~^~~
subprojects/highfive/include/highfive/bits/../H5PropertyList.hpp:87:10: note: candidate template ignored: constraints not satisfied [with P = (anonymous namespace)::Jpegls]
    void add(const P& property);
         ^
subprojects/highfive/include/highfive/bits/../H5PropertyList.hpp:86:15: note: because '(anonymous namespace)::Deflate' does not satisfy 'PropertyInterface'
    template <PropertyInterface P>
              ^
subprojects/highfive/include/highfive/bits/../H5PropertyList.hpp:65:9: note: because 'p.apply(hid)' would be invalid: no member named 'apply' in '(anonymous namespace)::Jpegls'
    { p.apply(hid) };
```

Here's the example C++20 concepts code:

```diff
diff --git a/include/highfive/H5PropertyList.hpp b/include/highfive/H5PropertyList.hpp
index cc4bf90..c40a848 100644
--- a/include/highfive/H5PropertyList.hpp
+++ b/include/highfive/H5PropertyList.hpp
@@ -57,6 +57,18 @@ class PropertyListBase: public Object {
     }
 };
 
+#if __cplusplus >= 202002L
+template<typename P>
+concept PropertyInterface = requires(P p, const hid_t hid)
+{
+    // The struct/class P must contain a public member function that accepts
+    // hid_t as the argument.
+    { p.apply(hid) };
+};
+
+#else
+#define PropertyInterface typename
+#endif
 
 ///
 /// \brief HDF5 property Lists
@@ -75,7 +87,7 @@ class PropertyList: public PropertyListBase {
     /// A property is an object which is expected to have a method with the
     /// following signature void apply(hid_t hid) const
     ///
-    template <typename P>
+    template <PropertyInterface P>
     void add(const P& property);
 
     ///
diff --git a/include/highfive/bits/H5PropertyList_misc.hpp b/include/highfive/bits/H5PropertyList_misc.hpp
[...]

PR #811, opened 2023-07-20
Constrain the Property object at compile-time

antonysigma:
**Description**

The property struct/class requires a public member function signature `apply(hid_t) const`. Describe the constrain in the C++20 concepts syntax. The primarily goal is to make compiler errors more readable for average users. Average users are those who do not understand SFINAE.

Resolves feature #810 .

**How to test this?**

It is a compile-time type checking. If the production code or test code compiles, the code is valid.

**Test System**

 - OS: Ubuntu 20.04
 - Compiler: clang 12.0 with C++20 and/or C++11
 - Dependency versions: Irrelevant.

### DIFF
--- a/include/highfive/H5PropertyList.hpp
+++ b/include/highfive/H5PropertyList.hpp
@@ -133,6 +133,26 @@ class PropertyListBase: public Object {
     friend T details::get_plist(const U&, hid_t (*f)(hid_t));
 };
 
+/// \interface PropertyInterface
+/// \brief HDF5 file property object
+///
+/// A property is an object which is expected to have a method with the
+/// following signature `void apply(hid_t hid) const`
+///
+/// \sa Instructions to document C++20 concepts with Doxygen: https://github.com/doxygen/doxygen/issues/2732#issuecomment-509629967
+///
+/// \cond
+#if __cplusplus >= 202002L
+template <typename P>
+concept PropertyInterface = requires(P p, const hid_t hid) {
+    {p.apply(hid)};
+};
+
+#else
+#define PropertyInterface typename
+#endif
+/// \endcond
+
 ///
 /// \brief HDF5 property Lists
 ///
@@ -149,8 +169,8 @@ class PropertyList: public PropertyListBase {
     /// Add a property to this property list.
     /// A property is an object which is expected to have a method with the
     /// following signature void apply(hid_t hid) const
-    ///
-    template <typename P>
+    /// \tparam PropertyInterface
+    template <PropertyInterface P>
     void add(const P& property);
 
     ///
@@ -438,6 +458,7 @@ class PageBufferSize {
 #endif
 
 /// \brief Set hints as to how many links to expect and their average length
+/// \implements PropertyInterface
 ///
 class EstimatedLinkInfo {
   public:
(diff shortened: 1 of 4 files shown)


antonysigma:
Hi HighFive project owners @alkino . I noticed that the PR passes all CI validations except `clang-format`. Shall I revise and resubmit the PR? I know how to run `clang-format` locally. Just wanna be certain about who does what, before I proceed.

1uc:
Thank you for the nice PR. Yes, please feel free to fix the formatting; I think the version of `clang-format` you need is `14.0.0`. Otherwise the required diff to appease CI is:

```
 diff --git a/include/highfive/H5PropertyList.hpp b/include/highfive/H5PropertyList.hpp
index af149a9..241c988 100644
--- a/include/highfive/H5PropertyList.hpp
+++ b/include/highfive/H5PropertyList.hpp
@@ -143,10 +143,9 @@ class PropertyListBase: public Object {
 ///
 /// \cond
 #if __cplusplus >= 202002L
-template<typename P>
-concept PropertyInterface = requires(P p, const hid_t hid)
-{
-    { p.apply(hid) };
+template <typename P>
+concept PropertyInterface = requires(P p, const hid_t hid) {
+    {p.apply(hid)};
 };
```

I've not had time today to compile it and see how it works myself. Though I'm looking forward to having a reason to look at concepts.

include/highfive/H5PropertyList.hpp
```diff
@@ -133,6 +133,27 @@ class PropertyListBase: public Object {
...
+/// \interface PropertyInterface
+/// \brief HDF5 file property object
+///
+/// A property is an object which is expected to have a method with the
+/// following signature `void apply(hid_t hid) const`
+///
+/// \sa Instructions to document C++20 concepts with Doxygen: https://github.com/doxygen/doxygen/issues/2732#issuecomment-509629967
+///
+/// \cond
+#if __cplusplus >= 202002L
```
1uc, line 145: @antonysigma  I would like there to be a way of forcing the concepts to be off. I'm worried a user will find a compiler that claims to be `202002L` but doesn't support concepts; or has a compiler bug. If this happens they would be unable to use HighFive, because a nice, but non-essential feature is broken.

If you like please add a CMake variable `HIGHFIVE_HAS_CONCEPTS` which defaults to `On`. Then add code to instruct the compiler to add `-DHIGHFIVE_HAS_CONCEPTS={0,1}` has appropriate. Then, the if-condition above would be
```
#if HIGHFIVE_HAS_CONCEPTS && __cplusplus >= 202002L
```

antonysigma: Done. I am not familiar to CMake >12.0 though. Feel free to revise the changes where it make sense.

1uc: Thanks! I slightly changed it and tested if it works. One might need `-fconcepts-diagnostics-depth=2` to get meaningful compiler output.

antonysigma: Great! Let me know if you need anything else. I will standby.

1uc: I'd like to prepare a PR which modifies CI to also compile HighFive as C++20 code. Otherwise, the concepts aren't tested during CI. Then we can rebase this PR onto the other one; and everything is tested nicely again. 

Thank you for this nice PR.

PR #813, opened 2023-07-27
Compile HighFive as C++20 project.

1uc:
This PR adds tests which check that HighFive compiles as a C++20 project, e.g. #811 will add concepts which were introduced in C++20.


matz-e (approved):
Thanks!

Issue #769, opened 2023-05-30
Investigate GCC -Warray-bounds warnings.


1uc:
When compiling the unit-tests on GCC 13.1.1 I observe:

```
In static member function ‘static _Up* std::__copy_move<_IsMove, true, std::random_access_iterator_tag>::__copy_m(_Tp*, _Tp*, _Up*) [with _Tp = long unsigned int; _Up = long unsigned int; bool _IsMove = false]’,
    inlined from ‘_OI std::__copy_move_a2(_II, _II, _OI) [with bool _IsMove = false; _II = long unsigned int*; _OI = long unsigned int*]’ at /usr/include/c++/13.1.1/bits/stl_algobase.h:506:30,
    inlined from ‘_OI std::__copy_move_a1(_II, _II, _OI) [with bool _IsMove = false; _II = long unsigned int*; _OI = long unsigned int*]’ at /usr/include/c++/13.1.1/bits/stl_algobase.h:533:42,
    inlined from ‘_OI std::__copy_move_a(_II, _II, _OI) [with bool _IsMove = false; _II = __gnu_cxx::__normal_iterator<long unsigned int*, vector<long unsigned int> >; _OI = __gnu_cxx::__normal_iterator<long unsigned int*, vector<long unsigned int> >]’ at /usr/include/c++/13.1.1/bits/stl_algobase.h:540:31,
    inlined from ‘_OI std::copy(_II, _II, _OI) [with _II = __gnu_cxx::__normal_iterator<long unsigned int*, vector<long unsigned int> >; _OI = __gnu_cxx::__normal_iterator<long unsigned int*, vector<long unsigned int> >]’ at /usr/include/c++/13.1.1/bits/stl_algobase.h:633:7,
    inlined from ‘static std::vector<long unsigned int> HighFive::details::inspector<std::vector<Size> >::getDimensions(const type&) [with T = std::array<std::__cxx11::basic_string<char>, 6>]’ at .../HighFive/include/highfive/bits/H5Inspector_misc.hpp:369:22,
    inlined from ‘static size_t HighFive::details::inspector<std::vector<Size> >::getSizeVal(const type&) [with T = std::array<std::__cxx11::basic_string<char>, 6>]’ at .../HighFive/include/highfive/bits/H5Inspector_misc.hpp:375:34,
[...]

PR #814, opened 2023-07-28
Avoid spurious GCC 13 array-bounds warnings.

1uc:
Prior to this commit GCC 13 would emit numerous warnings related to `-Warray-bounds`. There are numerous similar have been reported, see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56456

This PR replaces the `std::copy` with the equivalent loop.


1uc:
Addresses #769.

alkino (approved):

PR #817, opened 2023-08-03
Only add warning flags for tests and examples.

1uc:
The CMake code uses an interface target `HighFiveWarnings` which collects all warnings we want when building the examples and tests. However, we no longer set these warnings for users of HighFive, since they might have different preferences w.r.t. which flags should be set.


tests/unit/CMakeLists.txt
```diff
@@ -8,7 +9,7 @@ endif()
 ## Base tests
 foreach(test_name tests_high_five_base tests_high_five_multi_dims tests_high_five_easy test_all_types)
   add_executable(${test_name} "${test_name}.cpp")
-  target_link_libraries(${test_name} HighFive Catch2::Catch2WithMain)
+  target_link_libraries(${test_name} HighFive HighFiveWarnings Catch2::Catch2WithMain)
```
alkino, line 12: Maybe we should use the alias syntax: `HighFive::Warnings`, `HighFive::HighFive`, ...

alkino (approved):

tests/unit/CMakeLists.txt
1uc: That would look neat. We can look into that for `v3.0.0`.

PR #818, opened 2023-08-03
Add and use `<highfive/highfive.hpp>`.

1uc:
The header `<highfive/highfive.hpp>` is the top-level header for everything HighFive. It's recommended to use it instead of the other headers, unless compilation time is a concern. However, currently `<highfive/H5File.hpp>` includes almost all of HighFive. Hence, the savings are likely very modest, if any.


1uc:
This would require that we include `H5Version.hpp` in the regular source tree.

matz-e (approved):

alkino (approved):

PR #819, opened 2023-08-04
Include `H5Version.hpp` in the sources.

1uc:
Generated files require users to "build" and "install" HighFive, for the sole purpose of having files generated via `configure_file` available. The only file that generated in HighFive is the version file.

This PR proposes to automatically generate the file but also commit it to the repo. The advantage of doing so, is that users can clone the repo, set the appropriate value for `-I` and they're done (if they don't need any addons).

In order to check that the version number in `CMakeLists.txt` and `H5Version.hpp` are consistent we configure HighFive during CI and check that the file hasn't changed.

### DIFF
--- a/.github/workflows/version_file.yml
+++ b/.github/workflows/version_file.yml
@@ -1,0 +1,36 @@
+name: HighFive Check Version File
+
+on:
+  push:
+    branches:
+      - ci_test
+      - release/**
+  pull_request:
+    branches:
+      - master
+      - release/**
+
+jobs:
+  CheckVersion:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: true
+
+    - name: "Install libraries"
+      run: |
+        sudo apt-get -qq update
+        sudo apt-get -qq install libhdf5-dev ninja-build
+
+    - name: Build
+      run: |
+        # Will trigger `configure_file` for H5Version.hpp.
+        cmake -DHIGHFIVE_USE_BOOST=Off -B build .
+
+    - name: Test
+      run: |
+        # Check that the file hasn't changed, i.e. was updated
+        # after changing the version number.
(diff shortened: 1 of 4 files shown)


1uc:
Here the new CI ran with an outdated `H5Version.hpp`:
https://github.com/BlueBrain/HighFive/actions/runs/5760522600/job/15616631667

.github/workflows/version_file.yml
```diff
@@ -0,0 +1,36 @@
...
+    - name: Build
+      run: |
+        # Will trigger `configure_file` for H5Version.hpp.
+        cmake -DHIGHFIVE_USE_BOOST=Off -B build .
+
+    - name: Test
+      run: |
+        # Check that the file hasn't changed, i.e. was updated
+        # after changing the version number.
+        ! git status | grep include/highfive/H5Version.hpp
```
matz-e, line 36: Would be nice if we could ensure that the version is also greater or equal to the latest tag.

1uc: I think understand the motivation. It would be nice to have something that checks we didn't forget to bump the version number.

Our current workflow is to:
1. Create a commit which updates the change log and bumps the version.
2. Create a PR.
3. Merge the PR.
4. Tag the appropriate commit on `master`.

Therefore a naive solution has two issues:
* There's a brief period where the version number is strictly larger than the latest tag. Which requires as you say, that we check for greater or equal.
* Due to the greater or equal it'll initially succeed even if we forget to bump the version number. Only after we tag a commit, will it complain.

Instead it would be nicer to have a CD pipeline to create a release, it will change the version in `CMakeLists.txt` and tag consistently.

matz-e: > Instead it would be nicer to have a CD pipeline to create a release, it will change the version in CMakeLists.txt and tag consistently.

Yes, agreed. We can put this in a virtual backlog, let's get this in first.

matz-e (approved):

PR #823, opened 2023-09-09
Optionally Use Static Linking to HDF5

HunterBelanger:
I use HighFive and HDF5 in an HPC application. Usually, I try to link all of my dependencies statically to avoid problems on clusters. This PR adds a new CMake option, `HIGHFIVE_STATIC_HDF5`, which will then statically link to HDF5.  The new option is used to set `HDF5_USE_STATIC_LIBRARIES` before calling find_package for HDF5.


1uc (approved):

1uc:
@HunterBelanger We're reviewing and rewriting our CMake. I'd like to remove the variable `HIGHFIVE_STATIC_HDF5`. Instead users would use the HDF5 variables.

The idea is that:
```
set(HDF5_USE_STATIC_LIBRARIES On)

target_link_libraries(foo PUBLIC HighFive)
```
should cause HighFive to link to the static HDF5 libraries. Essentially, because it will largely equivalent to:
```
set(HDF5_USE_STATIC_LIBRARIES On)
find_package(HDF5 REQUIRED)
target_link_libraries(foo PUBLIC HDF5::HDF5)
target_include_directory(foo PUBLIC ${HIGHFIVE_INCLUDE_DIR})
```

Do you remember if you tried using the HDF5 variables directly and if yes why it didn't work? It's very possible that some of HighFive's CMake code was preventing this from working correctly.

Please let us know if it's obvious to you that using the HDF5 CMake variables has a flaw.

HunterBelanger:
@1uc, you are right, this does seem to work and is likely a better solution ! Feel free to remove this option if needed, and I will be on the lookout so I can modify my CMake files.

1uc:
Thank you! If the change happens it'll be in 3.0.0.

Issue #825, opened 2023-09-14
Tests appear to fail randomly


penguinpee:
**Describe the bug**

Building HighFive 2.7.1 for Fedora it appears tests are failing randomly.

**To Reproduce**

This happens in the Fedora build environment. The log files are available online:

1. https://download.copr.fedorainfracloud.org/results/gui1ty/neuro-sig/fedora-39-x86_64/06402644-highfive/builder-live.log.gz
```
The following tests FAILED:
	123 - HighFiveRenameRelative (Failed)
Errors while running CTest
```
2. https://download.copr.fedorainfracloud.org/results/gui1ty/neuro-sig/fedora-39-x86_64/06402644-highfive/builder-live.log.gz
```
The following tests FAILED:
	281 - Scalar in std::vector - std::vector<bool> (Failed)
	282 - Scalar in std::vector - std::vector<std::string> (Failed)
	285 - Scalar in std::array - bool (Failed)
Errors while running CTest
```

There's no difference between these builds except the time they took place. The command to run these tests:
```
/usr/bin/ctest --test-dir redhat-linux-build --output-on-failure --force-new-ctest-process -j2 -VV
```

Some of the errors are assertion errors, others appear to be some I/O issue. I've seen error messages like:
```
282: /builddir/build/BUILD/HighFive-2.7.1/tests/unit/test_all_types.cpp:54: FAILED:
282: due to unexpected exception with message:
282:   Unable to create file Test_vector.h5 (Virtual File Layer) Unable to lock file
```
frequently for different tests.

3. https://download.copr.fedorainfracloud.org/results/gui1ty/neuro-sig/fedora-39-x86_64/06402694-highfive/builder-live.log.gz

Third time lucky. This build succeeded having all tests pass.

**Expected behavior**

Tests to succeed or fail consistently.

**Stacktrace**

n/a

**Desktop (please complete the following information):**

 - OS: Fedora 38|39|rawhide
 - Version: HighFive 2.7.1

**Additional context**

Do I need to prep the environment in some way to make the tests run in a consistent manner? E.g. do tests need to be run sequentially?

1uc:
The issue might be that on one process HDF5 locks the file, then the other process wants to lock the same file while it's already locked and fails.

Given the rather generic `Test_vector.h5` and the fact that we to my knowledge don't run tests in parallel, this would be my first guess.

Would you mind running `ctest` with `-j1` (or without `-j1`)?

penguinpee:
[...]

PR #826, opened 2023-09-14
Use unique filenames in tests.

1uc:
Since HDF5 locks a file, running tests in parallel will fail if the file names aren't unique.


1uc:
Fixes #825.

matz-e (approved):

PR #857, opened 2023-11-21
Fail when reference counts are inconsistent.

1uc:
In HighFive we have two types of invalid HIDs. The first category consists of:
* `H5I_INVALID_HID` which HighFive uses to indicate that the object is "empty",
* `H5P_DEFAULT` (aka `0`) which is to "optimize" the default property.

The second category consists of anything else that `H5Iis_valid` reports as invalid, e.g. identifiers that have been closed, etc.

For the first category we must not increment or decrement the reference counters. However, for the second category, we'd like to get an error, because we're expecting either `H5I_INVALID_HID`, `H5P_DEFAULT` or a valid HID, but are given something else.

### DIFF
--- a/include/highfive/bits/H5DataType_misc.hpp
+++ b/include/highfive/bits/H5DataType_misc.hpp
@@ -131,7 +131,7 @@ inline StringType DataType::asStringType() const {
         throw DataTypeException("Invalid conversion to StringType.");
     }
 
-    if (isValid() && H5Iinc_ref(_hid) < 0) {
+    if (_hid > 0 && H5Iinc_ref(_hid) < 0) {
         throw ObjectException("Reference counter increase failure");
     }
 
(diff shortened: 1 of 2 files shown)


1uc:
It can uncover bugs in user code that stem from something like:
```
HighFive::File* file = malloc(sizeof(HighFive::File));
// Let's make it explicit that the allocated space might not be
// zeroed out:
memset(file, 'f');

*file = File("foo.h5");
```
because the `_hid` in `file` (with quite high probability) is an invalid, positive HID the assignment used to not decrement the junk HID. However, the new version will error out. The bug is already present in such user code, because if they hit a valid HID by chance, then the reference counts will be off.

1uc:
Reverting this back to draft to think more carefully about how to have a balance between not crashing, but still exposing existing bugs in user programs, see https://github.com/BlueBrain/HighFive/pull/857#issuecomment-1822257159

1uc:
Since HDF5 will print an error message, that's essentially what we'd do as well. (We can't throw from the dtor for example.)

PR #858, opened 2023-11-21
Check `ctest` output for HDF5-DIAG.

1uc:
HDF5 errors that aren't promoted to an exception by HighFive will simply print an error message. This means that tests can pass even if there are HDF5 errors. This will run the tests with `ctest --verbose` and check that the output doesn't contain `"HDF5-DIAG"`.


1uc:
Should reveal the bug fixed by #856.

matz-e (approved):

PR #869, opened 2023-11-27
Fix `create_datatype<bool>()`.

1uc:
Create the specialization from `create_datatype<bool>` and adds the tests to check that raw arrays of bool are bitwise compatible with the enum that's serialized.


alkino (approved):
Of course, I missed it when I added boolean

Issue #870, opened 2023-11-27
Unify `read(T*)` and `write_raw(T*)`.


1uc:
It's convenient to have the option of telling HighFive to just write from a pointer. Currently, one does this with:
```
dset.read(ptr);
dset.write_raw(ptr);
```
We should unify the two. Potentially, important twist to consider are things like `std::string*` which points to a pre-allocated array of strings. Which can't be used in `*_raw` because it'll segfault. However, it would make sense for `read(ptr)` and `write(ptr)` to use the pre-allocated array of `std::string`s and then copy the strings into/from the pre-allocated array from/to an internal buffer used to read/write the strings.

1uc:
Please consult the migration guide, it might help modernizing code.

Issue #865, opened 2023-11-24
Datatype of `char[][N]`.


1uc:
It's seems that a C array of C-string constants is serialized as an array of signed 8-bit ints. See,
```
$ h5dump -R build-2.5.1/tests/unit/h5_rw_vec_shortcut_test.h5
HDF5 "build-2.5.1/tests/unit/h5_rw_vec_shortcut_test.h5" {
GROUP "/" {
   DATASET "TmpCArray2dchar" {
      DATATYPE  H5T_STD_I8LE
      DATASPACE  SIMPLE { ( 4, 3 ) / ( 4, 3 ) }
      DATA {
      (0,0): 97, 97, 0,
      (1,0): 98, 98, 0,
      (2,0): 99, 99, 0,
      (3,0): 49, 50, 0
      }
   }
```
This behaviour might be older than `2.5.1`. We should review why this choice was made and if we can't find a good one make it serialize to something string-ish.

PR #871, opened 2023-11-27
Improve testing infrastructure.

1uc:
Adds/reimplements abstractions for the following:

  * Create multi-dimensional array filled with suitable values.
  * Traits for accessing values.
  * Traits for hiding the difference of DataSet and Attribute.
  * Useful utilities such as `ravel`, `unravel` and `flat_size`.

### DIFF
--- a/doc/developer_guide.md
+++ b/doc/developer_guide.md
@@ -91,3 +91,134 @@ release. Once this is done perform a final round of updates:
 * Update BlueBrain Spack recipe to use the archive and not the Git commit.
 * Update the upstream Spack recipe.
 
+## Writing Tests
+### Generate Multi-Dimensional Test Data
+Input array of any dimension and type can be generated using the template class
+`DataGenerator`. For example:
+```
+auto dims = std::vector<size_t>{4, 2};
+auto values = testing::DataGenerator<std::vector<std::array<double, 2>>::create(dims);
+```
+Generates an `std::vector<std::array<double, 2>>` initialized with suitable
+values.
+
+If "suitable" isn't specific enough, one can specify a callback:
+```
+auto callback = [](const std::vector<size_t>& indices) {
+    return 42.0;
+}
+
+auto values = testing::DataGenerator<std::vector<double>>::create(dims, callback);
+```
+
+The `dims` can be generated via `testing::DataGenerator::default_dims` or by
+using `testing::DataGenerator::sanitize_dims`. Remember, that certain
+containers are fixed size and that we often compute the number of elements by
+multiplying the dims.
+
+### Generate Scalar Test Data
+To generate a single "suitable" element use template class `DefaultValues`, e.g.
+```
+auto default_values = testing::DefaultValues<double>();
+auto x = testing::DefaultValues<double>(indices);
+```
+
+### Accessing Elements
+To access a particular element from an unknown container use the following trait:
+```
+using trait = testing::ContainerTraits<std::vector<std::array<int, 2>>;
+// auto x = values[1][0];
+auto x = trait::get(values, {1, 0});
+
+// values[1][0] = 42.0;
+trait::set(values, {1, 0}, 42.0);
+```
+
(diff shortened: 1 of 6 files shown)


doc/developer_guide.md
```diff
@@ -91,3 +91,134 @@ release. Once this is done perform a final round of updates:
 * Update BlueBrain Spack recipe to use the archive and not the Git commit.
 * Update the upstream Spack recipe.
 
+## Writing Tests
+### Generate Multi-Dimensional Test Data
+Input array of any dimension and type can be generated using the template class
+`DataGenerator`. For example:
+```
+auto dims = std::vector<size_t>{4, 2};
+auto values = testing::DataGenerator<std::vector<std::array<double, 2>>::create(dims);
```
tdegeus, line 100: I would use a library for this! For example, you can use the NumPy equivalent xtensor. Writing and maintaining some extra custom code is extra work and error-prone.

doc/developer_guide.md
```diff
@@ -91,3 +91,134 @@ release. Once this is done perform a final round of updates:
...
+auto default_values = testing::DefaultValues<double>();
+auto x = testing::DefaultValues<double>(indices);
+```
+
+### Accessing Elements
+To access a particular element from an unknown container use the following trait:
+```
+using trait = testing::ContainerTraits<std::vector<std::array<int, 2>>;
+// auto x = values[1][0];
+auto x = trait::get(values, {1, 0});
```
tdegeus, line 131: Following the command above:
```
auto x = a(1, 0);
```
there is also 
```
auto y = a.flat(10)
```
to read directly from the flat storage

tdegeus:
It's a good idea to streamline testing. However, as I stated in some comments, there are existing libraries for this, and I would strongly suggest not to seek a custom implementation. Using a library comes with the advantage that:
1. It does not imply extra code that has to be maintained.
2. It comes with a fixed API that is well-documented.
3. It comes with print functions. 
4. The libraries are usually themselves intensively tested.

doc/developer_guide.md
1uc: I'd like to test the following:
```
template<class Container>
void check_write_dataset(const Container& container) {
  file.createDataSet("foo", container);
}
```

Hence, I need to able to create arrays of all supported types
```
std::vector<T>
std::vector<std::array<T, 4>>
std::array<std::vector<T>, 4>
boost::
// xt:: (once supported in core)
```
and so forth, and be able to fill them with exactly the correct values. Being able to have any container filled with sensible values, requires being able to a) allocate containers of a given shape, b) assign values. Those two feature make up a good chunk of the PR. Then there very minor niceties like copying, etc.

The second issue is that I need a somewhat uniform API for creating the product of `template<class T> Container` with a bunch of different scalar types `T`. That's the other large chunk of the PR.

How does an `xt::array` help in this situation? Yes, I can test XTensor, but I can't test writing of any of the other containers. If I use XTensor arrays to write, I can read the other containers, but I still can't check that the values are correct, because I don't have access to some API unifying trait to hide the difference of `a[i][j]`, `a(i,j)` etc. (gets worse in 3D).

1uc:
In defense of the complexity: it found several bugs/quirks or things to follow up on:
* #875
* #874
* #869
* #879

It would trivially fix:
* #895

Any non-uniformity is very painful to us (not just our users), and therefore has a higher chance of getting noticed and fixed. The following come to mind:
* #865
* #870

doc/developer_guide.md
tdegeus: If there is a pressing reason to use a custom implementation, of course, it's fine. Anyway, this is not a public implementation, nor will change a lot. 

I just thought it would be easier to write some casters but avoid implementing (un)ravel(index) functions and equality operators.

ferdonline (approved):
I think it's a nice idea to unify Dataset and Property testing. You also added tests convering that, so I really don't have anything to nitpick about. If you want to wait for other reviews, great, but from my side good to go

PR #876, opened 2023-11-29
Move selection related tests.

1uc:
Part of reorganizing the tests requires grouping tests. One such group is anything related to selections, e.g. column, row or point selections, hyperslabs, etc.

This commit only moves code from one file to the next without any changes.


1uc:
See documentation in #871 for the general direction we're heading towards.

Issue #875, opened 2023-11-28
Segfault for `std::array<std::vector>`.


1uc:
Reading something like `std::array<std::vector<double>, 2>`, but presumably anything with a non-trivial ctor inside an `std::array` seems to cause a segfault because the size of the inner vector is zero.

```
#0  0x00007ffff6cac83c in ?? () from /usr/lib/libc.so.6
#1  0x00007ffff6c5c668 in raise () from /usr/lib/libc.so.6
#2  0x00007ffff6c444b8 in abort () from /usr/lib/libc.so.6
#3  0x00007ffff6edd3b2 in std::__glibcxx_assert_fail (file=<optimized out>, line=<optimized out>, function=<optimized out>, condition=<optimized out>)
    at /usr/src/debug/gcc/gcc/libstdc++-v3/src/c++11/debug.cc:61
#4  0x0000555555720b84 in std::vector<double, std::allocator<double> >::operator[] (this=0x7ffff4efe270, __n=0)
    at /usr/include/c++/13.2.1/bits/stl_vector.h:1125
#5  0x00005555557bcf8b in HighFive::details::inspector<std::vector<double, std::allocator<double> > >::unserialize<double const*> (
    vec_align=@0x7ffff4f55c60: 0x60c000015280, dims=std::vector of length 1, capacity 1 = {...}, val=std::vector of length 0, capacity 0)
    at HIGH_FIVE_DIR/HighFive/include/highfive/bits/H5Inspector_misc.hpp:425
#6  0x00005555557ab3cf in HighFive::details::inspector<std::array<std::vector<double, std::allocator<double> >, 5ul> >::unserialize<double const*> (
    vec_align=@0x7ffff4bf7de0: 0x60c000015280, dims=std::vector of length 2, capacity 2 = {...}, val=...)
    at HIGH_FIVE_DIR/HighFive/include/highfive/bits/H5Inspector_misc.hpp:553
#7  0x000055555571100e in HighFive::details::DeepCopyBuffer<std::array<std::vector<double, std::allocator<double> >, 5ul> >::unserialize (
    this=0x7ffff4f55550, val=...) at HIGH_FIVE_DIR/HighFive/include/highfive/bits/H5Converter_misc.hpp:94
#8  0x00005555556926c2 in HighFive::Attribute::read<std::array<std::vector<double, std::allocator<double> >, 5ul> > (this=0x7ffff4efe230, array=...)
[...]

PR #877, opened 2023-11-30
Ensure elements in `std::array` are allocated.

1uc:
A bug in `inspected::prepare` left the elements of an `std::array` as default allocated. This fails when putting an `std::vector` inside an `std::array`, because the `std::vector`s need to be allocated with the appropriate length.


1uc:
Fixes #875 more rigorous testing coming via #871.

PR #880, opened 2023-11-30
Improve `details::Boolean` compatibility checks.

1uc:
Instead of comparing values, use `std::memcmp` to compare bytes.


1uc:
MSVC complained (and I believe errored out) about the previous variation. This variation is more blunt about comparing bytes.

matz-e (approved):

PR #885, opened 2023-12-04
Wrap all used H5D functions.

1uc:


1uc:
This splits out a reviewable chunk from https://github.com/BlueBrain/HighFive/pull/863.

PR #897, opened 2023-12-18
Rework CMake code for v3.0.0.

1uc:
This PR develops the next version of HighFive's CMake code. The ideas are:
* Avoid one flag for each per optional dependency (other than testing and examples).
* Use targets only.


cmake/HighFiveWarnings.cmake
```diff
@@ -5,6 +5,8 @@ if(TARGET HighFiveWarnings)
     return()
 endif()
 
+message("HighFiveWanring")
```
ferdonline, line 8: Shouldnt the message be a bit more explanatory? Or dropped altogether?
```suggestion
message("HighFiveWarning")
```

tests/unit/tests_high_five_base.cpp
```diff
@@ -28,6 +28,15 @@
 #include <highfive/highfive.hpp>
 #include "tests_high_five.hpp"
 
+#ifdef H5_USE_BOOST
+#include <highfive/boost.hpp>
```
ferdonline, line 32: Just confirming the intent: these headers are not included by default, so it is users' responsibility to meet the dependencies if they want to use/include them? That is cool

CMakeLists.txt
```diff
@@ -66,76 +34,113 @@ option(HIGHFIVE_HAS_WERROR "Convert warnings to errors." OFF)
...
-    message(FATAL_ERROR "HighFive needs to be compiled with at least C++11")
-endif()
-
 if(NOT DEFINED CMAKE_CXX_STANDARD)
-    set(CMAKE_CXX_STANDARD 11)
+    if(HIGHFIVE_USE_XTENSOR)
+        set(CMAKE_CXX_STANDARD 14)
+    else()
+        set(CMAKE_CXX_STANDARD 11)
+    endif()
```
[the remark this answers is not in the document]
1uc: We gain very little from C++14. Therefore, I'm not sure it's worth the inconvenience to users.

cmake/HighFiveWarnings.cmake
1uc: Looks like debug output. This really is a sketch of WIP so that people can see what's meant by new CMake.

tests/unit/tests_high_five_base.cpp
1uc: Exactly, same mechanism that pybind11 and boost use to give you default specializations of traits for STL containers.

CMakeLists.txt
```diff
@@ -66,76 +34,113 @@ option(HIGHFIVE_HAS_WERROR "Convert warnings to errors." OFF)
...
+  set(HIGHFIVE_EXTRAS_DEFAULT ON)
+else()
+  set(HIGHFIVE_EXTRAS_DEFAULT OFF)
+endif()
+
+option(HIGHFIVE_UNIT_TESTS "Compile unit-tests" ${HIGHFIVE_EXTRAS_DEFAULT})
+option(HIGHFIVE_EXAMPLES "Compile examples" ${HIGHFIVE_EXTRAS_DEFAULT})
+option(HIGHFIVE_BUILD_DOCS "Build documentation" ${HIGHFIVE_EXTRAS_DEFAULT})
+
+option(HIGHFIVE_USE_BOOST "Enable Boost testing" OFF)
```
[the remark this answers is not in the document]
1uc: The fear is that CI will be green; but nothing ran because boost is spelled `BooST` or it's moved somewhere in the image and CMake can't find it anymore, or ...

Yes, name could be changed, again it's very much WIP.

1uc:
Currently, there's two interesting things. First:
```
find_package(HighFive)
target_link_libraries(foo PUBLIC HighFive)
```
still works, if one doesn't use any optional dependencies. If one does, it'll requires changes to the source code, but not the build-system.

Second: we don't provide code to link to optional dependencies. The argument is that if a library has dependency on, say, Boost, then they already have requirement to do `-I${BOOST_DIR}/include` and `-L${BOOST_DIR}/lib -lboost` (or whatever the commands are) even if they don't use HighFive. Hence, we could leave it to them to set it up correctly. Not including the dependencies has advantages, because users are free to pick how it work. For example when they vendor their other optional dependencies `find_dependency(Boost)` might do the wrong thing.

The question is should we help them? One could by doing the following:
```
find_package(HighFive COMPONENTS boost)
target_link_libraries(foo PUBLIC HighFive::HighFive HighFive::boost)
```
The `COMPONENTS boost` triggers the `find_dependency(boost)` and `HighFive::boost` triggers adding `-I` and `-L` as needed for the boost dependencies.

1uc:
AFAICT one one pitfall is that pHDF5 requires linking to MPI, i.e.
```
target_link_libraries(HighFive INTERFACE HDF5::HDF5)
if(HDF5_IS_PARALLEL)
  find_package(MPI)
  target_link_libraries(HighFive INTERFACE MPI::MPI_C MPI::MPI_CXX)
endif()
```
However, now the target `HighFive` uses the MPI target when on the build machine the HDF5 version happened to be pHDF5. This is why we're using `$<BUILD_INTERFACE:MPI::MPI_C>`.

CMakeLists.txt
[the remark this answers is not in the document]
ferdonline: I like the space savings with `enable_if_t`  😆

CMakeLists.txt
```diff
@@ -66,76 +34,113 @@ option(HIGHFIVE_HAS_WERROR "Convert warnings to errors." OFF)
...
 endif()
 
 add_compile_definitions(HIGHFIVE_CXX_STD=${CMAKE_CXX_STANDARD})
 
-# Search dependencies (hdf5, boost, eigen, xtensor, mpi) and build target highfive_deps
-include(${PROJECT_SOURCE_DIR}/CMake/HighFiveTargetDeps.cmake)
+# HighFive
+# --------
+add_library(HighFive INTERFACE)
+add_library(HighFive::HighFive ALIAS HighFive)
```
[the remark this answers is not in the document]
1uc: Yes, this makes everything perfectly uniform. `HighFive` (slightly backwards compatible) and `HighFive::HighFive` are the same thing and work regardless of how one vendors HighFive (submodules, fetch content, not vendored).

CMakeLists.txt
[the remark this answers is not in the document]
1uc: Renamed, but kept the variables for the reason given.

1uc:
> Second: we don't provide code to link to optional dependencies. 

I'm increasingly convinced that attempting to find and link optional dependencies will likely end up incorrect under too many circumstances. IIUC the Boost ecosystem seems to have not yet settled on a set of commands that works pretty much anywhere. OpenCV has targets, but doesn't document them AFAICT. Even if we did use its targets, they might change them in a manner that's not compatible with our CMake, which seems fair because the project is telling us to not use them. Moreover, we'd probably only include a very small subset of the library and users would almost certainly want more. Same argument for Boost.

Not using targets has issues as documented here:
https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#creating-relocatable-packages

CMakeLists.txt
```diff
@@ -66,76 +34,135 @@ option(HIGHFIVE_HAS_WERROR "Convert warnings to errors." OFF)
...
-if(CMAKE_CXX_STANDARD EQUAL 98)
-    message(FATAL_ERROR "HighFive needs to be compiled with at least C++11")
-endif()
-
 if(NOT DEFINED CMAKE_CXX_STANDARD)
-    set(CMAKE_CXX_STANDARD 11)
+    if(HIGHFIVE_TEST_XTENSOR)
+        set(CMAKE_CXX_STANDARD 14)
+    else()
+        set(CMAKE_CXX_STANDARD 11)
```
matz-e, line 76: Shouldn't this get updated now?

cmake/HighFiveConfig.cmake
```diff
@@ -0,0 +1,19 @@
...
+
+include("${CMAKE_CURRENT_LIST_DIR}/HighFiveTargets.cmake")
+
+if(HDF5_IS_PARALLEL)
+  find_dependency(MPI)
+  target_link_libraries(HighFive::HighFive INTERFACE MPI::MPI_C MPI::MPI_CXX)
+endif()
+
+add_library(HighFive ALIAS HighFive::HighFive)
+add_library(HighFiveInclude ALIAS HighFive::Include)
```
matz-e, line 19: Do we want this? Does this just encourage more confusion between targets and library names?

tests/cmake_integration/application/deps/HighFive
```diff
@@ -0,0 +1 @@
+/home/lucg/git/bbp/HighFive/tests/cmake_integration/../..
```
matz-e, line 1: errr.... Pretty sure I don't have that directory!

CMakeLists.txt
1uc: That's been prepped in the other PR: #957.

cmake/HighFiveConfig.cmake
1uc: In `CMakeLists.txt` we create a target called `HighFive`. If we export under a namespace it'll be called `HighFive::HighFive` when doing `find_package(HighFive ...)`. However, when we do `add_subdirectory(third_party/HighFive)` it'll still be called `HighFive`.

The idea is to have both mean the same thing with the hope that the difference then doesn't matter. We might be able to change it such that `HighFive::HighFive` is the only target everywhere.

Does the `::` convince CMake to always consider it a target rather than a string that could be converted to `libHighFive.so`; or do we just end up with `libHighFive::HighFive.so`?

tests/cmake_integration/application/deps/HighFive
1uc: Yes, it's automatically regenerated when the script that checks everything runs. Needs to be removed and added to `.gitignore`.

cmake/HighFiveConfig.cmake
matz-e: > Does the :: convince CMake to always consider it a target rather than a string that could be converted to libHighFive.so; or do we just end up with libHighFive::HighFive.so?

I _think_ that plain `HighFive` is ambiguous and may be library name or target, whereas `HighFive::HighFive` is only a target (valid or not).

Don't we alias target `HighFive` to `HighFive::HighFive` elsewhere in the CMake? We can leave this in, just seems like `HighFive::HighFive` would be more "canonical" by now (thanks CMake, for setting clear standards :) )

1uc: Easy enough to test. It fails with the following error:
```
CMake Error at CMakeLists.txt:5 (target_link_libraries):
  Target "foo" links to:

    HighFive::HighFive

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

> Don't we alias target HighFive to HighFive::HighFive elsewhere in the CMake?
Yes, in `CMakeLists.txt` we naturally have `HighFive` and create an alias `HighFive::HighFive`. While in `*Config.cmake` we naturally have `HighFive::HighFive` and create an alias `HighFive`. This way the two are synonyms and always available.

We could try if we can hide/eliminate the target `HighFive` (and `HighFiveInclude`) in favour of `HighFive::HighFive`.

matz-e: If CMake doesn't hide it, it's exposed :)

I suppose if we already have `HighFive` when used as a git submodule, we should keep it in the CMake install, too,

1uc: Yes, I came to the same conclusion. Renaming the target doesn't help, because even if there is no target called `HighFive`, the problematic line:
```
target_link_libraries(foo PUBLIC HighFive)
```
remains valid during configuration and only fails at compile time. Therefore, it's probably better to have the target so that if people correctly find or include HighFive they're not punished of guessing that the target is called `HighFive` like is was in `v2`.

README.md
```diff
@@ -168,86 +180,55 @@ project (typically as a git submodule), for example in `third_party/HighFive`.
...
+`CMakeLists.txt` should contain the following:
 
-```bash
-add_library(HighFive INTERFACE)
-target_include_directory(HighFive INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/third_party/HighFive/include)
+```cmake
+set(HIGHFIVE_FIND_HDF5 Off)
+find_package(HighFive REQUIRED)
+# Alternatively:
+# add_subdirectory(third_party/HighFive)
```
tdegeus, line 213: It might not be crystal clear that this is an alternative for `find_package`

tdegeus:
It seems fair to simply strip third-party dependencies from the target and leave it up to the user. To stay uniform, the `define`s should be removed everywhere, incl. form H5Easy.

1uc:
> To stay uniform, the defines should be removed everywhere, incl. from H5Easy.
I've spotted this. Solving this nicely kinda requires stripping out all the logic in `H5Easy` and move everything XTensor and Eigen related into core. After that H5Easy would do very little just a light API change.

The problem is that XTensor doesn't have a compile time constant rank. Which is a non-issue when implemented Easy-style. When we want to move it into the inspector it starts messing with deduction of rank, which becomes an issue with empty arrays and also messes with the "broad-casting" feature.

I've mildly come to peace with the idea that Easy is "easy" and opinionated. Therefore, it continues to guess the presence of libraries by checking the existence of dependency related macros.

OTOH, it would be nice to make Easy just a light API change.

README.md
```diff
@@ -168,86 +180,55 @@ project (typically as a git submodule), for example in `third_party/HighFive`.
...
-`third_party/HighFive`. Then one could create a target:
+To prevent HighFive from searching or "linking" to HDF5 the project's
+`CMakeLists.txt` should contain the following:
 
-```bash
-add_library(HighFive INTERFACE)
-target_include_directory(HighFive INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/third_party/HighFive/include)
+```cmake
+set(HIGHFIVE_FIND_HDF5 Off)
+find_package(HighFive REQUIRED)
```
1uc, line 211:
```suggestion
# Now include HighFive, either via `find_package` or by
# using `add_submodule`.
find_package(HighFive REQUIRED)
```

README.md
1uc: Thank you, we just spent the two immediately preceding subsections explaining the difference.

tdegeus:
> > To stay uniform, the defines should be removed everywhere, incl. from H5Easy.
> 
> I've spotted this. Solving this nicely kinda requires stripping out all the logic in `H5Easy` and move everything XTensor and Eigen related into core. After that H5Easy would do very little just a light API change.
> 
> The problem is that XTensor doesn't have a compile time constant rank. Which is a non-issue when implemented Easy-style. When we want to move it into the inspector it starts messing with deduction of rank, which becomes an issue with empty arrays and also messes with the "broad-casting" feature. Additionally, the not related issue of string vs arrays of character which cause stripping of dimensions doesn't make that part of the code any easier to understand.
> 
> I've mildly come to peace with the idea that Easy is "easy" and opinionated. Therefore, it continues to guess the presence of libraries by checking the existence of dependency related macros.
> 
> OTOH, it would be nice to make Easy just a light API change.

A simpler idea is to make the `define`s 'private' and have a uniform API:
```cpp
# include <highfive/eigen.hpp>
# include <highfive/xtensor.hpp>
# include <highfive/H5easy.hpp>
```
One could then statically assert on features not (yet) in the core. This would simplify the API, I think.

I am not quite up to speed with the templated rank issue. I think it is a separate issue, though. 
(That being said: highfive is header-only, it is fine to have templated features??)

1uc:
Let's track the H5Easy related questions in a different issue/PR. They don't prevent us from merging this as it is; and sort H5Easy separately.

matz-e (approved):

PR #899, opened 2023-12-18
Add named ctors for scalar and null dataspaces.

1uc:
The issue is that:
```
DataSpace{DataSpace::dataspace_scalar};
DataSpace{DataSpace::dataspace_null};
```
are a simple dataspace with 0 elements and a simple dataspace with one element, respectively. This provides a safer alternative.

PR #900, opened 2023-12-18
Use enum class for DataSpace::DataspaceType. 

1uc:
See #899 for what the issue is and a safe alternative. Unfortunately, changing an enum to an enum class is a breaking change and can't be done until `3.0.0`.


alkino (approved):

include/highfive/H5DataSpace.hpp
```diff
@@ -45,12 +45,16 @@ class DataSpace: public Object {
     /// This enum is needed otherwise we will not be able to distringuish between both with normal
     /// constructors. Both have a dimension of 0.
     /// \since 1.3
-    enum DataspaceType {
+    enum class DataspaceType {
```
1uc, line 48: Needs to be migrated to the other PR.

PR #905, opened 2023-12-21
Guard functionality introduced in 1.10.0.

1uc:


1uc:
Found while testing #897. Tests to prevent regressions will be added via that PR.

alkino (approved):

PR #912, opened 2024-01-12
Avoid non-collective behaviour.

1uc:
MPI collective IO, requires that certain HDF5 functions are called collectively. Here some short-circuiting logic for empty datasets interferes when some, but not all, selections from different MPI ranks are empty.


1uc:
Merge after #913.

pramodk (approved):
LGTM

PR #922, opened 2024-01-23
Refactor BufferInfo::Operation as enum class.

1uc:


1uc:
Since it's internal we should be free to change this.

PR #927, opened 2024-01-29
Remove unused specialization for createDataSet.

1uc:
This was likely needed to hide the bug that the specialization for `create_datatype<bool>` was missing. See, commit 9415a7c82.


matz-e (approved):

alkino (approved):
Better, thanks ;)

PR #928, opened 2024-01-29
Add `read_raw` and prepare to deprecate `read(T*, ...)`.

1uc:


1uc:
Unfortunately deprecating this is very noisy and misleading, since `int a[][] = {{1, 2}, {3, 4}}` uses the deprecated overload.

alkino (approved):

PR #938, opened 2024-02-03
No constexpr rank via constexpr min-max rank.

1uc:
The commit removes the requirement for a constexpr rank. Instead
containers need a minimum and maximum constexpr rank.

The difficulties are:

  * Empty arrays can't figure out their runtime rank. This is an issue
    when deducing the dimension when writing to disk. The "solution" is
    to deduce with lowest rank possible.


1uc:
This is a sketch and probably not the way to go.

1uc:
This allows us to do `xt::xarray`: #976

matz-e (approved):

PR #939, opened 2024-02-03
Pass dimensions to 'serialize'.

1uc:
By passing the dimensions to `serialize` allows serializing nested pointers. This was previously not possible because, we didn't know how many elements the pointer pointed to.

The logic for obtaining the dimensions of the array is:

  * When creating/opening the array the memspace is set to be the same as the filespace.

  * Performing a selection either leads to a simple (packed) multi-dimensional memspace; or to a one-dimensional memspace.

Therefore, if we perform a write without selection, we know the dimensions the input array must have. If we perform a selection we again know the dimensions the input buffer needs to have.


1uc:
Since it changes the inspector, we should probably not merge this into `v2`.

jorblancoa (approved):

matz-e (approved):

PR #941, opened 2024-02-06
Migration instructions for `read_raw`.

1uc:


matz-e (approved):
General comment: I personally would not use contractions in more "formal" documentation, and thus use `it is` over `it's`…

1uc:
I guess that's technically correct. I'll just consider our documentation not formal.

PR #947, opened 2024-02-08
Remove deprecated default ctor for DataSet

alkino:


1uc (dismissed):
To be discussed :)

1uc:
This demonstrates that we need to expand the test coverage to ensure it errors out correctly.

1uc (approved):

Issue #961, opened 2024-02-15
Quick-Fix CI for publishing docs.


1uc:
Currently docs are published for `master`. Which might be misleading. We should publish for the `v2` branch only.

PR #963, opened 2024-02-15
Fix when CI is run.

1uc:
The tests should run for every PR and when merged into the main or `v2.x` branch. Docs should be published only when merging into `v2.x` (for now).

We've removed `release/**` since we don't have those. It also removes `ci_test` because I haven't seen anyone test the CI that way, `git commit --allow-empty` can be used instead.


1uc:
See #961.

matz-e (approved):

PR #966, opened 2024-02-16
Fix when CI is run.

1uc:
The tests should run for every PR and when merged into the main or `v2.x` branch. Docs should be published only when merging into `v2.x` (for now).


1uc:
Backports #963.

matz-e (approved):

PR #976, opened 2024-04-05
Implement XTensor support in core.

1uc:
Adds support for `xt::xtensor`, both row and column major. We're missing `xt::xarray` and views. Views fail because they claim to have runtime variable rank.

### DIFF
--- a/include/highfive/xtensor.hpp
+++ b/include/highfive/xtensor.hpp
@@ -64,7 +64,7 @@ struct xtensor_inspector_base {
 
     static hdf5_type* data(type& val) {
         if (!is_trivially_copyable) {
-            throw DataSetException("Invalid used of `inspector<xarray>::data`.");
+            throw DataSetException("Invalid used of `inspector<XTensor>::data`.");
         }
 
         if (val.size() == 0) {
@@ -76,7 +76,7 @@ struct xtensor_inspector_base {
 
     static const hdf5_type* data(const type& val) {
         if (!is_trivially_copyable) {
-            throw DataSetException("Invalid used of `inspector<xarray>::data`.");
+            throw DataSetException("Invalid used of `inspector<XTensor>::data`.");
         }
 
         if (val.size() == 0) {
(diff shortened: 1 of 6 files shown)


include/highfive/xtensor.hpp
```diff
@@ -0,0 +1,106 @@
...
+        std::array<size_t, ndim> shape = val.shape();
+        std::vector<size_t> sizes(shape.begin(), shape.end());
+        return sizes;
+    }
+
+    static void prepare(type& val, const std::vector<size_t>& dims) {
+        auto shape = std::array<size_t, ndim>{};
+        std::copy(dims.begin(), dims.begin() + ndim, shape.begin());
+        val.resize(shape);
+    }
```
alkino, line 38: no recursivity here?

1uc: None of it works with non-trivial elements. I should check if they even allow anything non-trivial.

1uc: We're now asserting the fact. I propose we revisit, if we ever need to support them.

1uc:
@alkino The reason for not allowing non-trivial elements, is that this way we can use `xt::adapt` which also paves the way for implementing all other flavours of `xt` objects. With the same idea we will be able to take any `xt` object, and assign it to (read it from) memory we've wrapped with an `xt::adapt`. This idea works for row-/column-major and stridded/padded arrays.

It doesn't work for anything that contains non-trivial elements.

include/highfive/xtensor.hpp
```diff
@@ -0,0 +1,212 @@
...
+        return {shape.begin(), shape.end()};
+    }
+
+    static void prepare(type& val, const std::vector<size_t>& dims) {
+        val.resize(Derived::shapeFromDims(dims));
+    }
+
+    static hdf5_type* data(type& val) {
+        if (!is_trivially_copyable) {
+            throw DataSetException("Invalid used of `inspector<xarray>::data`.");
```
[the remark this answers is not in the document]
1uc: Thank you, there used to be separate inspectors for `xarray` and `xtensor`. Looks like I copied these lines from the `xarray` variation.

matz-e (approved):

Issue #978, opened 2024-04-14
Add a filter to HighFive


LauZanMo:
I would like to ask whether there is a way to change the default filter of highfive to blosc, because the default filter,which is set to szip, is slow and my system has real-time requirements. I notice that the highfive code does not seem to support the setting of filter. Can it be implemented using the native HDF5 function ?

LauZanMo:
I spent some time clarifying the structure of PropertyList in HighFive and found it very easy to add filter on this basis.
First, I need to declare and implement the Blosc class in H5PropertyList.hpp and H5PropertyList _ misc.hpp, respectively. Just like this:

```C++
# H5PropertyList.hpp

/// \implements PropertyInterface
class Blosc {
    public:
        explicit Blosc(unsigned level, unsigned shuffle, unsigned compressor);

    private:
        friend DataSetCreateProps;
        void apply(hid_t hid) const;
        unsigned _cd_values[7];
};



# H5PropertyList_misc.hpp

inline Blosc::Blosc(unsigned int level, unsigned int shuffle, unsigned int compressor) {
    _cd_values[4] = level;
    _cd_values[5] = shuffle;
    _cd_values[6] = compressor;
}

inline void Blosc::apply(const hid_t hid) const {
    detail::h5p_set_blosc(hid, _cd_values);
}
```

Then, as for the specific implementation function ‘detail::h5p_set_blosc’, I refer to the implementation of hdf5_blosc: #https://github.com/Blosc/hdf5-blosc/blob/master/src/example.c#L57 just like this:

```C++
# h5p_wrapper.hpp

#include "blosc_filter.h"
inline herr_t h5p_set_blosc(hid_t plist_id, const unsigned int *cd_values) {
    char *version, *date;
    if (register_blosc(&version, &date) < 0) {
        HDF5ErrMapper::ToException<PropertyException>("Blosc filter unavailable.");
    }
    
    herr_t err = H5Pset_filter(plist_id, FILTER_BLOSC, H5Z_FLAG_OPTIONAL, 7, cd_values);
    if (err < 0) {
        HDF5ErrMapper::ToException<PropertyException>("Error setting blosc property");
    }
    return err;
}

```

Of course I need to link hdf5-blosc dynamic library files:

```Cmake
# root CMakeLists.txt
# I install hdf5-blosc to the default path
set(blosc_filter_DIR /usr/local/lib/) 
target_link_libraries(HighFive blosc_filter)
[...]

PR #980, opened 2024-04-15
Add the option for using blosc filter

LauZanMo:
See #978 

This PR adds a custom blosc filter and has been tested in create_dataset_double.cpp in examples folder.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -101,6 +101,11 @@ target_link_libraries(HighFive INTERFACE HighFive::Include)
 if(HIGHFIVE_FIND_HDF5)
   find_package(HDF5 REQUIRED)
   target_link_libraries(HighFive INTERFACE HDF5::HDF5)
+  
+  # I install hdf5-blosc to the default path
+  set(blosc_filter_DIR /usr/local/lib/)
+  target_link_libraries(HighFive INTERFACE blosc_filter)
+  target_link_directories(HighFive INTERFACE /usr/local/include/)
 endif()
 
 if(HDF5_IS_PARALLEL)
--- a/include/highfive/bits/H5PropertyList_misc.hpp
+++ b/include/highfive/bits/H5PropertyList_misc.hpp
@@ -347,6 +347,16 @@ inline void Shuffle::apply(const hid_t hid) const {
     detail::h5p_set_shuffle(hid);
 }
 
+inline Blosc::Blosc(unsigned int level, unsigned int shuffle, unsigned int compressor) {
+    _cd_values[4] = level;
+    _cd_values[5] = shuffle;
+    _cd_values[6] = compressor;
+}
+
+inline void Blosc::apply(const hid_t hid) const {
+    detail::h5p_set_blosc(hid, _cd_values);
+}
+
 inline AllocationTime::AllocationTime(H5D_alloc_time_t alloc_time)
     : _alloc_time(alloc_time) {}
 
--- a/include/highfive/bits/h5p_wrapper.hpp
+++ b/include/highfive/bits/h5p_wrapper.hpp
@@ -239,6 +239,21 @@ inline herr_t h5p_set_szip(hid_t plist_id, unsigned options_mask, unsigned pixel
     return err;
 }
 
+#include "blosc_filter.h"
+
+inline herr_t h5p_set_blosc(hid_t plist_id, const unsigned int *cd_values) {
+    char *version, *date;
+    if (register_blosc(&version, &date) < 0) {
+        HDF5ErrMapper::ToException<PropertyException>("Blosc filter unavailable.");
+    }
+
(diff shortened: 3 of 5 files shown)


CMakeLists.txt
```diff
@@ -101,6 +101,11 @@ target_link_libraries(HighFive INTERFACE HighFive::Include)
 if(HIGHFIVE_FIND_HDF5)
   find_package(HDF5 REQUIRED)
   target_link_libraries(HighFive INTERFACE HDF5::HDF5)
+  
+  # I install hdf5-blosc to the default path
+  set(blosc_filter_DIR /usr/local/lib/)
+  target_link_libraries(HighFive INTERFACE blosc_filter)
+  target_link_directories(HighFive INTERFACE /usr/local/include/)
```
alkino, line 108: We will have to fix that.

include/highfive/bits/h5p_wrapper.hpp
```diff
@@ -239,6 +239,21 @@ inline herr_t h5p_set_szip(hid_t plist_id, unsigned options_mask, unsigned pixel
...
+    if (register_blosc(&version, &date) < 0) {
+        HDF5ErrMapper::ToException<PropertyException>("Blosc filter unavailable.");
+    }
+
+    herr_t err = H5Pset_filter(plist_id, FILTER_BLOSC, H5Z_FLAG_OPTIONAL, 7, cd_values);
+    if (err < 0) {
+        HDF5ErrMapper::ToException<PropertyException>("Error setting blosc property");
+    }
+    return err;
+}
```
alkino, line 255: @1uc I don't think this is how you think how your wrapper, right?
We will need to put this helper in an other place?

CMakeLists.txt
LauZanMo: Since blosc is an extra plugin of hdf5, it is necessary to compile and link this external libraries. hdf5-blosc can not be found using find_package macro since it is packaged too simple [here](https://github.com/Blosc/hdf5-blosc/blob/master/CMakeLists.txt#L75). I haven't figured out an elegant way to link this library.

include/highfive/bits/h5p_wrapper.hpp
LauZanMo: Unlike szip, a built-in filter in hdf5, blosc filter must be registered before use. Perhaps a register wrapper should be added here.

1uc:
To avoid the issue with forcing an optional dependency (blosc) on all users, we'll need to rework the PR. I think we can safely move all blocs related code to `highfive/blosc.hpp`. Ensure we include the header for blosc. The wrapper for `H5Pset_filter` stays in `h5p_wrapper.hpp`.

To fix the build-system maybe the following works:
1. Temporarily remove all CMake code.
2. For testing only add it back similar to how boost and Eigen are handled.

This has the advantage (for us) that we don't specify how to find/link to blosc, it's left up to the user. This is different everywhere and we're likely to get it wrong if blosc doesn't have `*-config.cmake` files. The solution @LauZanMo shows presumably works very well for their usecase, but unfortunately for a library that's not suitable. We can always add a target `highfive::blosc` if needed, but we can't take it away.

include/highfive/bits/h5p_wrapper.hpp
```diff
@@ -239,6 +239,21 @@ inline herr_t h5p_set_szip(hid_t plist_id, unsigned options_mask, unsigned pixel
     return err;
 }
 
+#include "blosc_filter.h"
+
+inline herr_t h5p_set_blosc(hid_t plist_id, const unsigned int *cd_values) {
+    char *version, *date;
```
1uc, line 245: These are both leaked, because `register_blosc` uses `strdup` which requires us to call `free`.

include/highfive/bits/H5PropertyList_misc.hpp
```diff
@@ -347,6 +347,16 @@ inline void Shuffle::apply(const hid_t hid) const {
...
 }
 
+inline Blosc::Blosc(unsigned int level, unsigned int shuffle, unsigned int compressor) {
+    _cd_values[4] = level;
+    _cd_values[5] = shuffle;
+    _cd_values[6] = compressor;
+}
+
+inline void Blosc::apply(const hid_t hid) const {
+    detail::h5p_set_blosc(hid, _cd_values);
```
1uc, line 357: This function likely can't be called twice. Therefore, if two property lists with Blosc are created I suspect a misleading error message will be printed. It's unclear if it'll work, because `register_blosc` seems to unconditionally `return 1;` (despite defining and setting a `retval`).

CMakeLists.txt
alkino: What I mean is that hard coded path to your system is not the valid way to give access to blosc.

1uc:
Thank you for the time and effort that went into the PR. Unfortunately, we'll need to close it, due to upcoming changes.

PR #986, opened 2024-04-19
Add `inspector::is_trivially_nestable`.

1uc:
We need this because `std::span` is considered `std::is_trivially_copyable`.

### DIFF
--- a/include/highfive/bits/H5Inspector_misc.hpp
+++ b/include/highfive/bits/H5Inspector_misc.hpp
@@ -132,6 +132,10 @@ inspector<T> {
     // If this value is false: serialize, unserialize are mandatory
     static constexpr bool is_trivially_copyable
 
+    // Is this type trivially nestable, i.e. is type[n] a contiguous
+    // array of `base_type[N]`?
+    static constexpr bool is_trivially_nestable
+
     // Reading:
     // Allocate the value following dims (should be recursive)
     static void prepare(type& val, const std::vector<std::size_t> dims)
@@ -162,6 +166,7 @@ struct type_helper {
     static constexpr size_t ndim = 0;
     static constexpr size_t recursive_ndim = ndim;
     static constexpr bool is_trivially_copyable = std::is_trivially_copyable<type>::value;
+    static constexpr bool is_trivially_nestable = is_trivially_copyable;
 
     static std::vector<size_t> getDimensions(const type& /* val */) {
         return {};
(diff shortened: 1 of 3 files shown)


matz-e (approved):

1uc:
Thanks for letting us know. I've checked the PR again, and nothing sticks out as wrong. Do you know what `T` is and if `T*` is inside some other container, the type of that container?

PhilipDeegan:
T is a `double` in this context, but we do some pointer casting to force a 1d vector into a 2 or 3d dataset

which is feasibly not how it should be done, but it's what has worked until now

```
template<std::size_t dim, typename Data>
NO_DISCARD auto pointer_dim_caster(Data* data)
{
    if constexpr (dim == 1)
        return data;
    if constexpr (dim == 2)
        return reinterpret_cast<Data const** const>(data);
    if constexpr (dim == 3)
        return reinterpret_cast<Data const*** const>(data);
}
```

1uc:
Yes! That was an interesting surprise. We still have a test for a variation using `write_raw` here:
https://github.com/BlueBrain/HighFive/blame/070badf6935e17bb74c5a9d0e08969f640b4a87b/tests/unit/test_legacy.cpp#L27-L29

I see three options:
* Use `write_raw` without (or with) the magic in `pointer_dim_caster`.
* Use the almost ready `reshapeMemSpace`, see #991. Comments are welcome. 
* Use `mdspan` which should be one of the next things I'd like to implement.

PhilipDeegan:
what we're doing is essentially the same as the legacy test, which I'm surprised if it's passing

I can test `write_raw` without the reinterpret_cast, as the shape given is the correct dimension
we have no plans for C++23 right now

PhilipDeegan:
without the reinterpret_cast I get 
image

1uc:
The `write_raw` doesn't care about dimensions and such, it's just a very shallow wrapper:
https://github.com/BlueBrain/HighFive/blob/070badf6935e17bb74c5a9d0e08969f640b4a87b/include/highfive/bits/H5Slice_traits_misc.hpp#L268-L279

The type of the pointer doesn't matter, they're all converted to `void *`  and sent to HDF5.

PhilipDeegan:
it seems ok with `write_raw` and no reintrepret_cast :+1:

1uc:
Glad to hear that!

1uc:
This change will be documented in the Migration Guide via https://github.com/BlueBrain/HighFive/pull/996.

Migration Guide: https://bluebrain.github.io/HighFive/md__2home_2runner_2work_2_high_five_2_high_five_2doc_2migration__guide.html

PR #991, opened 2024-05-06
Implement `squeeze` and `reshape`.

1uc:
These methods allow reshaping the memory space. This can be used as
an alternative to broadcasting.

Example:

Let `dset` be a dataset with shape `[3, 1, 1]`. Then,

    dset.reshapeMemSpace({3}).read<std::vector<double>>();
    dset.squeezeMemSpace({1, 2}).read<std::vector<double>>();

can be used to read into a one-dimensional dataset.

### DIFF
--- a/include/highfive/H5Attribute.hpp
+++ b/include/highfive/H5Attribute.hpp
@@ -13,6 +13,7 @@
 #include <H5Apublic.h>
 
 #include "H5DataType.hpp"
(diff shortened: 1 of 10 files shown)


1uc:
This is preparatory work to allow #992.

matz-e (approved):
LGTM

PR #1011, opened 2024-05-23
Remove duplicated tests.

1uc:
There's (at least) three attempts at testing read/write cycles for multi-dimensional arrays. The first is whatever is in `tests_*multi_dims.cpp`. Since it's partial, a second attempt was made: the first half of `test_all_types.cpp`. Since this was still partial and not quite DRY enough, a third attempt was made.

This PR removes the duplicated tests from the first and second attempt, one-by-one. Each commit removes one test. Or moves/reimplements a test if it's not yet covered by the third attempt.

There's a fourth attempt in `test_high_five_base.cpp`; but that's left for another PR.


1uc:
If it helps reviewing, we could split out the commits that move code, e.g. `compare_arrays` into a separate PR. Then this one will be removing code and reimplementing the odd test; but no refactoring.

alkino (approved):

alkino:
Nice to split `test_boost` and `test_stl` as `test_opencv` and `test_xtensor` already are.
For the next time, maybe a `test_eigen`?

1uc:
Interesting, I thought I'd split them out already, but yes, they can get split away too.

PR #1016, opened 2024-06-10
Refactor H5Easy to use core.

1uc:
Removes all custom implementations for serializing Eigen, XTensor, OpenCV, `std::vector` and uses HighFive core instead.

This allows us to remove most specialization, except the one for Eigen which serializes `Matrix<double, 1, -1>` as a one-dimensional array and reads it from one- or two-dimensional data.


1uc:
Note that it's current based on top of #1011.

alkino (approved):

PR #1031, opened 2024-07-25
Optimize chained hyperslab selection.

1uc:
A common pattern for creating semi-unstructured selection is to use many
(small) RegularHyperSlab and chain them:

```
HyperSlab hyperslab;
for(auto slab : regular_hyper_slabs) {
  hyperslab |= slab;
}
```

This eventually triggers calling:
```
for(auto slab : regular_hyper_slabs) {
  auto [offset, stride, counts, blocks] = slab;
  H5Sselect_hyperslab(space_id, offset, stride, counts, block);
}
```

Measurements show that this has runtime that's quadratic in the number
of regular hyper slabs. This starts becoming prohibitive at 10k - 40k
slabs.

We noticed that `H5Scombine_select` does not suffer from the same
performance issue. This allows us to optimize (long) chain of `Op::Or`
using divide and conquer.

The current implementation only optimizes streaks of `Op::Or`.

### DIFF
--- a/include/highfive/bits/H5Slice_traits.hpp
+++ b/include/highfive/bits/H5Slice_traits.hpp
@@ -162,20 +162,7 @@ class HyperSlab {
     }
 
     DataSpace apply(const DataSpace& space_) const {
-        auto space = space_.clone();
-        for (const auto& sel: selects) {
-            if (sel.op == Op::None) {
-                detail::h5s_select_none(space.getId());
-            } else {
-                detail::h5s_select_hyperslab(space.getId(),
-                                             convert(sel.op),
-                                             sel.offset.empty() ? nullptr : sel.offset.data(),
-                                             sel.stride.empty() ? nullptr : sel.stride.data(),
-                                             sel.count.empty() ? nullptr : sel.count.data(),
-                                             sel.block.empty() ? nullptr : sel.block.data());
-            }
-        }
-        return space;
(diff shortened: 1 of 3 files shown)


jorblancoa (approved):
LGTM!
Did you test it with libsonata to confirm that the regression is fixed?

1uc:
Yes, they provided the following reproducer:
```
import libsonata
import numpy as np
import time

np.random.seed(42)

sto = libsonata.NodeStorage('sscx-nodes-sonata.h5')
pop = sto.open_population('All')

#ids = np.arange(0, 100000, 2)
count = int(0.01*pop.size)
# count = 100
print(f'selecting {count} from {pop.size}')
ids = np.random.randint(0, pop.size, count)

t1 = time.perf_counter()
sel = libsonata.Selection(ids)
t2 = time.perf_counter()
print(f"elapsed = {t2 - t1}")

print(np.mean(pop.get_attribute('x', sel)))
```
The selection results in about 41k slabs, which takes 40s with the performance bug and 0.06 - 0.1s with `libsonata@master` and `highfive@1uc/backport-optimize-hyperslab-selection`. It takes 0.1 - 0.16s for `libsonata@0.1.24`.

We also ran their integration tests against the backport of this branch:
https://github.com/BlueBrain/HighFive-testing/actions/runs/10108668331

PR #1034, opened 2024-07-25
backport: Update Windows CI.

1uc:
Since a recent update it seems that Windows 2022 now uses Visual Studio 2022 17.9 and doesn't include toolset v141 anymore.


1uc:
Backports: #1000

jorblancoa (approved):

PR #1038, opened 2024-08-05
Fix corner cases of combine_selection.

1uc:


tests/unit/test_high_five_selection.cpp
```diff
@@ -591,3 +591,54 @@ TEST_CASE("select_multiple_ors", "[hyperslab]") {
         }
     }
 }
+
+TEST_CASE("select_multiple_ors_edge_cases", "[hyperslab]") {
```
alkino, line 595: what is the test only? expecting not throw?

1uc: Yes, these were to hunt down ways we can encounter a:
```
H5Scombine_select(lhs, Op::Or, rhs);
```
with invalid values for `lhs` and `rhs`. Note that only hyperslabs are valid. Critically, neither `NONE` nor `ALL` are considered a hyperslabs. There seems to be two ways we can end up with `NONE` selection by combining hyperslabs:
1. We set the counts to `{0, 0}`.
2. We use `AND` with two hyperslabs that have an empty intersection.

Case 1 can happen for `lhs` and `rhs`. Case 2 can only happen to `lhs` (but never `rhs`) when we combine a long chain with the existing slab.

I've been unable to create an `ALL` from `RegularHyperSlab`.

matz-e (approved):

